PR #227, opened 2020-11-04
Add Bitnami Redis

pascal-airy:


infrastructure/deployments/webhook-publisher.yaml
```diff
@@ -32,7 +32,7 @@ spec:
         - name: SERVICE_NAME
           value: webhook-publisher
         - name: REDIS_HOSTNAME
-          value: redis-cluster
+          value: redis-master
```
lucapette, line 35: didn't redis also move away from `master`? I think they did so we should also do that here

lucapette (changes requested):

PR #231, opened 2020-11-04
Map outbound content model in communication app

### LINKED ISSUE #216: Map message content
Replace API payloads that serve the message `content` as a source string with ones that return the output from the existing content mapper in `backend/lip/mapping`

- [x] `/conversations.info`
- [x] `/messages.list`
- [x] `/queue/message`

chrismatix:
fixes #216

### DIFF
--- a/backend/api/communication/src/test/java/co/airy/core/api/communication/WebSocketControllerTest.java
+++ b/backend/api/communication/src/test/java/co/airy/core/api/communication/WebSocketControllerTest.java
@@ -41,6 +41,7 @@ import org.springframework.web.socket.messaging.WebSocketStompClient;
 
 import java.lang.reflect.Type;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
@@ -96,9 +97,6 @@ public class WebSocketControllerTest {
 
     @Autowired
     private MockMvc mvc;
-
-    @Autowired
-    private Stores store;
 
     @Autowired
     private Jwt jwt;
--- a/backend/api/communication/BUILD
+++ b/backend/api/communication/BUILD
@@ -9,6 +9,7 @@ app_deps = [
     "//backend/avro/communication:channel",
     "//backend/avro/communication:metadata-action",
     "//backend/avro/communication:read-receipt",
+    "//backend/lib/mapping",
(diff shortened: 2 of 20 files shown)


backend/api/communication/src/test/java/co/airy/core/api/communication/WebSocketControllerTest.java
```diff
@@ -135,17 +133,17 @@ void init() throws Exception {
 
     @Test
     void sendsToWebsocket() throws Exception {
-        final CompletableFuture<MessageUpsertPayload> messageFuture = subscribe(port, MessageUpsertPayload.class, QUEUE_MESSAGE, jwt);
+        final CompletableFuture<Map> messageFuture = subscribe(port, Map.class, QUEUE_MESSAGE, jwt);
```
chrismatix, line 136: cc @paulodiniz and @lucapette The tests were failing, because we cannot parse the message content back with Jackson. If we want to be able to do that we would have to add [Jackson polymorphic type annotations](https://github.com/FasterXML/jackson-docs/wiki/JacksonPolymorphicDeserialization) to the content model, which would mean that message content would by default be returned with a `type` field.

Perphaps we have to this anyway, because otherwise, clients will also not know how to deserialize these fields. What do you think?

chrismatix: Going the Jackson polymorphism + generating Typescript definitions way would yield something like this: https://jechlin.github.io/ts-gen-aws/

paulodiniz: @chrismatix @lucapette I like and agree with the idea of adding a `type` to the payload. It *immediately* brought the `json-api` spec to mind.

https://jsonapi.org/

What do you think about trying to go full `json-api` spec by making the response in this format?

```json5
{
  "data": [{
     "type": ":conversation"
     ...
  }],
  
}
```

paulodiniz: As @chrismatix pointed, changing to `json-api` would demand changing the whole api.

On the `type` attribute, I like it 👍

lucapette: same, a `type` field can come handy in a many situations

paulodiniz (approved):

PR #239, opened 2020-11-06
Fix empty request payload NPEs

### LINKED ISSUE #235: Check endpoints for request payload NPEs
**Current state**
Some endpoints (conversation info, mark read, delete tag etc.) throw an NPE for empty request payloads.

**Expected state**
Request payloads are properly annotated so that missing properties result in a `400 Bad Request` response.

chrismatix:
fixes #235


backend/api/admin/src/main/java/co/airy/core/api/admin/TagsController.java
```diff
@@ -91,12 +100,18 @@
...
-        tag.setColor(tagColors.get(payload.getColor()));
-        tag.setName(payload.getName());
+        if (tagColors.containsKey(payload.getColor())) {
+            tag.setColor(tagColors.get(payload.getColor()));
+        } else if (payload.getColor() != null) {
+            return ResponseEntity.status(HttpStatus.BAD_REQUEST)
+                    .body(new RequestErrorResponsePayload(wrongTagColorErrorMessage));
+        }
+
+        tag.setName(Optional.ofNullable(payload.getName()).orElse(tag.getName()));
```
lucapette, line 114: I think of controller actions always the same way and here we'd benefit from the approach:

- verify input (if not, return error kinf of lines like in line 103)
- do the action
- return stuff

So what I'd do is test for `payload.getColor() == null || !tagColors.containsKey(payload.getColor())` and return bad request before doing the update action

backend/lib/payload/src/main/java/co/airy/payload/response/RequestErrorResponsePayload.java
```diff
@@ -7,6 +7,6 @@
 
 @Data
 @AllArgsConstructor
-public class RequestError implements Serializable {
+public class RequestErrorResponsePayload implements Serializable {
```
lucapette, line 10: good gardening 👏

lucapette (changes requested):
I'd change the tag update action to avoid the if then else chain (I'm also not sure the current logic is correct. We should probably cover tags with more tests to be fair)

lucapette (dismissed):
:shipit:

PR #241, opened 2020-11-06
Generate typescript definitions from content model

### LINKED ISSUE #217: Generate Typescript definitions from message content
The typed message content is modeled as Java classes in `co.airy.mapping.model` (`backend/lib/mapping/src/main/java/co/airy/mapping/model/`). In order to make them usable in the clients we want to be able to automatically generate Typescript definitions.

```
backend.
├── backend/
│   └── lib/
│       └── mapping/
│           └── src/main/java/co/airy/mapping/
│               └── model/
│                   ├── Content.java (content dto model)
│                   └── Text.java
└── frontend/
    └── types/
        ├── Content.d.ts (generate by running a Bazel command)
        └── Text.d.ts
```

- [x] transpile using a bazel command
- [x] update docs

chrismatix:
fixes #217


backend/lib/mapping/BUILD
```diff
@@ -19,6 +19,17 @@ java_library(
     deps = lib_deps,
 )
 
+java_binary(
```
chrismatix, line 22: @lucapette and @paulodiniz as long as we don't want to create ts definitions for anything else (which I think makes sense, because this problem is so unique), I would keep the generator as a binary that is local to this package.

frontend/components/src/content.ts
```diff
@@ -0,0 +1,14 @@
...
+export interface Content {
+    type: "text";
+}
+
+export interface Text extends Content {
+    type: "text";
+    text: string;
+}
+
+export type ContentUnion = Text;
```
chrismatix, line 14: With this, we can build a `switch { case }` function in the components lib that takes a `Content` payload and depending on the type field returns a react component.

paulodiniz:
Nice work!

backend/lib/mapping/src/ts-generator/java/co/airy/ts_generator/Main.java
```diff
@@ -0,0 +1,35 @@
...
+        settings.setExcludeFilter(List.of("java.io.Serializable"), List.of("**.**Builder"));
+
+        final TypeScriptGenerator generator = new TypeScriptGenerator(settings);
+
+        final Input.Parameters parameters = new Input.Parameters();
+
+        parameters.debug = false;
+        parameters.classNamePatterns = List.of("co.airy.mapping.model.**");
+
+        final File output = new File(System.getenv().get("BUILD_WORKSPACE_DIRECTORY") + "/frontend/components/src/content.ts");
```
chrismatix, line 30: In the same spirit, everything here is hardcoded on purpose.

paulodiniz (approved):

PR #258, opened 2020-11-09
Do not use javatuples

lucapette:


backend/api/communication/src/main/java/co/airy/core/api/communication/dto/CountAction.java
```diff
@@ -1,6 +1,31 @@
...
+public class CountAction implements Serializable {
+    public enum ActionType {
+        INCREMENT,
+        RESET
+    }
+
+    private long readDate;
+    private ActionType actionType;
+
+    CountAction(ActionType actionType, long readDate) {
```
chrismatix, line 19: `@AllArgsConstructor` does not give us this? 🤔

chrismatix (dismissed):
LGTM with a question

backend/api/communication/src/main/java/co/airy/core/api/communication/dto/CountAction.java
lucapette: yeah good point! I'll update

chrismatix (approved):

PR #275, opened 2020-11-11
Add Google message sender app

### LINKED ISSUE #266: Google message sender

chrismatix:
fixes #266


backend/sources/google/sender/src/main/java/co/airy/core/sources/google/GoogleConfig.java
```diff
@@ -0,0 +1,29 @@
...
+@Configuration
+public class GoogleConfig {
+    private static final Logger log = AiryLoggerFactory.getLogger(GoogleConfig.class);
+    final ObjectMapper objectMapper;
+    public GoogleConfig(ObjectMapper objectMapper) {
+        this.objectMapper = objectMapper;
+    }
+
+    @Bean
+    public GoogleServiceAccount googleServiceAccount(@Value("${google.auth.sa}") String googleServiceAccountFile) {
```
chrismatix, line 21: For the google sender I would take the approach we took in the tedi automation app. Instead of accessing the google's `channel→token` users provide a single line version of the Service Account file. This means:

- `token` becomes optional for Google channels
- We don't need a channel's table in the sender app

As long as we do not have product (user) input for this, I believe that this approach fits the Google authentication flow better.

paulodiniz (approved):

backend/sources/facebook/sender/src/main/java/co/airy/core/sources/facebook/services/Api.java
```diff
@@ -48,10 +48,10 @@ public Api(ObjectMapper objectMapper, RestTemplateBuilder restTemplateBuilder) {
     }
 
 
-    public void sendMessage(final String pageToken, SendMessagePayload fbSendMessagePayload) {
+    public void sendMessage(final String pageToken, SendMessagePayload sendMessagePayload) {
         String fbReqUrl = String.format(requestTemplate, pageToken);
 
-        restTemplate.postForEntity(fbReqUrl, new HttpEntity<>(fbSendMessagePayload, httpHeaders), FbSendMessageResponse.class);
+        restTemplate.postForEntity(fbReqUrl, new HttpEntity<>(sendMessagePayload, httpHeaders), FbSendMessageResponse.class);
```
lucapette, line 54: ❤️ 🍃

backend/sources/google/sender/src/main/java/co/airy/core/sources/google/Sender.java
```diff
@@ -0,0 +1,109 @@
...
+            message.setDeliveryState(DeliveryState.DELIVERED);
+
+            return message;
+        } catch (ApiException e) {
+            log.error(String.format("Failed to send a message to Facebook \n SendMessageRequest: %s \n FB Error Message: %s \n", sendMessageRequest, e.getMessage()), e);
+        } catch (Exception e) {
+            log.error(String.format("Failed to send a message to Facebook \n SendMessageRequest: %s", sendMessageRequest), e);
+        }
+
+        // TODO move the change state logic to backend/avro/message
```
lucapette, line 86: I think I'd do already? (but I'm find we want to follow up with it. If that, I think we should have an issue though)

lucapette (approved):

backend/sources/google/sender/src/main/java/co/airy/core/sources/google/Sender.java
chrismatix: True, did not have the time yesterday but will do it right after the conversation test fix

PR #276, opened 2020-11-11
Add test deploy step

pascal-airy:


.github/workflows/main.yml
```diff
@@ -57,6 +57,26 @@ jobs:
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
 
+      - name: Push beta app images
+        if: ${{github.ref == 'refs/heads/develop'}}
+        run: |
+          echo ${{ secrets.PAT }} | docker login ghcr.io -u pascal-airy --password-stdin
```
lucapette, line 63: we have to use a machine user here so that the credentials don't depend on any team member

infrastructure/helm-chart/charts/api-admin/templates/deployment.yaml
```diff
@@ -22,21 +22,30 @@ spec:
     spec:
       containers:
       - name: app
-        image: ghcr.io/airyhq/api/admin:latest
+        image: "{{ .Values.image }}:{{ .Values.global.appImageTag }}"
         imagePullPolicy: Always
         env:
         - name: KAFKA_BROKERS
-          value: airy-cp-kafka-headless:9092
```
lucapette, line 29: since we're changing the names of this (which is great), I think it makes sense to drop the `airy-` from everywhere. It would be cleaner and I don't think the prefix is a requirement for anything (correct me if I'm wrong though)

infrastructure/helm-chart/charts/api-auth/templates/deployment.yaml
```diff
@@ -22,29 +22,28 @@ spec:
...
-          value: <jwt_secret>
-        - name: MAIL_URL
-          value: <mail_url>
-        - name: MAIL_PORT
-          value: "<mail_port>"
-        - name: MAIL_USERNAME
-          value: <mail_username>
-        - name: MAIL_PASSWORD
-          value: <mail_password>
-        - name: MAIL_FROM
```
lucapette, line 46: those are requirements for the api auth application so, unless I'm not getting how this works now, the pod will never reach running state with this deployment descriptor?

infrastructure/helm-chart/charts/postgres/values.yaml
```diff
@@ -1 +1,3 @@
+pgUser: postgresadmin
 pgPassword: <pg_password>
+dbName: airy_core
```
lucapette, line 3: probably the best name here would be `admin` (as this is the db of the api admin service)

infrastructure/scripts/airy-conf.sh
```diff
@@ -2,75 +2,18 @@
...
-kubectl apply -f ~/airy-core/sources-facebook-sender.yaml
-kubectl apply -f ~/airy-core/sources-facebook-webhook.yaml
-kubectl apply -f ~/airy-core/webhook-consumer.yaml
-kubectl apply -f ~/airy-core/webhook-publisher.yaml
-kubectl apply -f ~/airy-core/sources-chatplugin.yaml
-
-kubectl scale deployment airy-cp-schema-registry --replicas=1
-kubectl exec kafka-client -- /root/wait-for-service.sh airy-cp-schema-registry 8081 15 Schema-registry
+kubectl scale deployment airy-schema-registry --replicas=1
+kubectl exec kafka-client -- /root/wait-for-service.sh airy-schema-registry 8081 15 Schema-registry
```
lucapette, line 16: this looks so much better now 👏

lucapette (changes requested):
First pass, wery happy about this!

Requesting changes because of the user but I would indeed address all the naming changes as well

infrastructure/helm-chart/charts/api-auth/templates/deployment.yaml
pascal-airy: If you look up a couple of lines the whole user-config cm is used as an env input. Which works but I get that explicit is better than implicit since maybe in the future there will be vars in the user-config that should not be in this pod

PR #277, opened 2020-11-11
Fix chat plugin messaging behavior

### LINKED ISSUE #273: Chatplugin smoke test fixes
The smoke test revealed the following bugs in the chat plugin source:

- [x] Websocket messages are not filtered per source and conversation
- [x] The chat plugin does not mark messages sent from the inbox as delivered after sending it to the queue
- [x] We currently broadcast all messages to all users

chrismatix:
fixes #273

### DIFF
--- a/backend/sources/chat-plugin/src/main/java/co/airy/core/chat_plugin/Principal.java
+++ b/backend/sources/chat-plugin/src/main/java/co/airy/core/chat_plugin/Principal.java
@@ -1,18 +1,22 @@
 package co.airy.core.chat_plugin;
 
-import lombok.AllArgsConstructor;
-import lombok.Builder;
+import co.airy.uuid.UUIDV5;
 import lombok.Data;
 
 @Data
-@Builder
-@AllArgsConstructor
 public class Principal implements java.security.Principal {
     private String channelId;
     private String sessionId;
+    private String conversationId;
+
+    public Principal(String channelId, String sessionId) {
+        this.channelId = channelId;
+        this.sessionId = sessionId;
+        this.conversationId = UUIDV5.fromNamespaceAndName(channelId, sessionId).toString();
+    }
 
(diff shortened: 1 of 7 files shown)


backend/sources/chat-plugin/src/main/java/co/airy/core/chat_plugin/Principal.java
```diff
@@ -13,6 +14,7 @@
 
     @Override
     public String getName() {
-        return sessionId;
+        // == conversation id
+        return UUIDV5.fromNamespaceAndName(channelId, sessionId).toString();
```
lucapette, line 18: probably over zealous _but_ we could generate in the constructor and return the cached value here

backend/sources/chat-plugin/src/main/java/co/airy/core/chat_plugin/Stores.java
```diff
@@ -42,9 +45,23 @@
...
                 .peek((messageId, message) -> webSocketController.onNewMessage(message));
 
+        messageStream.filter((messageId, message) -> !message.getSenderType().equals(SenderType.SOURCE_CONTACT)
+                && message.getDeliveryState().equals(DeliveryState.PENDING)
+        )
+                .mapValues((messageId, message) -> {
+                    message.setDeliveryState(DeliveryState.DELIVERED);
+                    webSocketController.onNewMessage(message);
+                    return message;
+                })
```
lucapette, line 62: if something goes wrong with the websocket, we end up not changing the state of the message so I'm thinking we should do the websocket part in a terminal node

lucapette (changes requested):
Looking good! I think we have to address the topology as it's a simple change and will give the whole source better stability

lucapette (approved):
:shipit:

PR #280, opened 2020-11-12
Fix reverse conversation list order

lucapette:
In the context of this fix, I refactored the conversation generator. While I like the approach much more, I don't really like the `TestConversation.from` method so we should talk about that before we merge


chrismatix (dismissed):
approvato

paulodiniz (approved):

Issue #269, opened 2020-11-11
Requests are not logged


lucapette:
The apps at the moment are a little too silent and we should provide basic logging for requests (and possibly all the exceptions that may occur at runtime)

PR #291, opened 2020-11-13
Log web requests

lucapette:
The journey of this pull request is interesting. #269 is happening because we miss `lib/spring/spring-web` dep from the web projects. When I started looking into how to fix it, I found myself lost in th way our deps are organised. I did a first step: make everything explicit, but I would argue we can create some more umbrella dep (in `backend/BUILD`) to DRY them a little


BUILD
```diff
@@ -95,6 +95,88 @@ java_plugin(
...
+    ],
+)
+
+java_library(
+    name = "springboot_test",
+    visibility = [
+        "//visibility:public",
+    ],
+    exports = [
+        ":junit",
```
lucapette, line 171: not a big fan of this, but works well in practise... I'm thinking if we introduce `backend/BUILD` we can have an umbrella dep there

BUILD
```diff
@@ -95,6 +95,88 @@ java_plugin(
...
+        "@maven//:org_springframework_boot_spring_boot_starter",
+        "@maven//:org_springframework_boot_spring_boot_starter_actuator",
+        "@maven//:org_springframework_boot_spring_boot_starter_jetty",
+        "@maven//:org_springframework_boot_spring_boot_starter_web",
+        "@maven//:org_springframework_spring_beans",
+        "@maven//:org_springframework_spring_context",
+        "@maven//:org_springframework_spring_core",
+        "@maven//:org_springframework_spring_web",
+        "@maven//:org_springframework_spring_webmvc",
+    ],
```
lucapette, line 120: we could extract more libraries from here. like core/boot/actuator/web

backend/api/auth/BUILD
```diff
@@ -1,8 +1,12 @@
 app_deps = [
+    "//:jackson",
     "//:lombok",
+    "//:springboot",
```
lucapette, line 4: I believe all the apps have jackson lombok springboot and log  so that would probably be our starting point

paulodiniz (approved):

backend/lib/spring/web/src/main/java/co/airy/spring/web/filters/RequestLoggingFilter.java
```diff
@@ -70,8 +70,6 @@ private String createMessage(HttpServletRequest request) {
             if (sanitizedHeaders.get("content-type").startsWith("application/json")) {
                 String payload = getMessagePayload((ContentCachingRequestWrapper) request);
                 ThreadContext.put("payload", payload);
-                ThreadContext.put("deviceType", sanitizedHeaders.get("device_type"));
```
chrismatix, line 73: 🌱

backend/sources/facebook/sender/BUILD
```diff
@@ -3,12 +3,10 @@ load("//tools/build:junit5.bzl", "junit5")
...
 
 app_deps = [
-    "//backend/avro/communication:channel",
-    "//backend/avro/communication:message",
-    "//backend/lib/payload",
-    "//backend/lib/kafka/schema:kafka-schema",
-    "//backend/lib/kafka/schema:application-communication-channels",
-    "//backend/lib/kafka/schema:application-communication-messages",
+    "//backend:base_app",
+    "//backend:channel",
```
chrismatix, line 7: Reads much nicer. The domain concern is composed of both so instead of bundling one into the other we create this shorthand 👍

chrismatix (approved):
LGTM, still needs squash

PR #294, opened 2020-11-16
[215] Infra/publish images

pascal-airy:


scripts/bootstrap.sh
```diff
@@ -123,4 +123,4 @@ fi
 
 cd $infra_path
 vagrant destroy -f
-vagrant up
+APP_IMAGE_TAG=$APP_IMAGE_TAG vagrant up
```
lucapette, line 126: this must default to `release` (guessing that's the right name)

infrastructure/scripts/provision/airy-core.sh
```diff
@@ -2,27 +2,27 @@
 set -euo pipefail
 IFS=$'\n\t'
 
+app_image_tag="${APP_IMAGE_TAG:-latest}"
```
lucapette, line 5: the default should come from the outside I think (as in the bootstrap...?) otherwise it may be error prone

.github/workflows/main.yml
```diff
@@ -57,6 +57,26 @@ jobs:
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
 
+      - name: Push beta app images
+        if: ${{github.ref == 'refs/heads/develop'}}
+        run: |
+          echo ${{ secrets.PAT }} | docker login ghcr.io -u pascal-airy --password-stdin
```
lucapette, line 63: we need the machine user here

.github/workflows/main.yml
```diff
@@ -57,6 +57,26 @@ jobs:
...
+          release_targets=$(bazel query "filter("beta$", //...)" --output label)
+          for target in $release_targets; do
+            echo "Deploying $target"
+            bazel run $target
+          done
+
+      - name: Push release app images
+        if: ${{github.ref == 'refs/heads/main'}}
+        run: |
+          echo ${{ secrets.PAT }} | docker login ghcr.io -u pascal-airy --password-stdin
```
lucapette, line 73: and here

I'm thinking we deleted the release script (on my suggestion I know) but maybe we should put it back and reuse it in both actions (which effectively are the same thing and the only difference is the tag)

lucapette (changes requested):

backend/api/auth/BUILD
```diff
@@ -45,10 +45,19 @@ load("//tools/build:junit5.bzl", "junit5")
 load("@io_bazel_rules_docker//container:container.bzl", "container_push")
 
 container_push(
```
chrismatix, line 47: To dry up the build files you could also write: 
```python
[
container_push(
    name = tag,
    format = "Docker",
    image = ":image",
    registry = "ghcr.io/airyhq/api",
    repository = "auth",
    tag = tag,
) for tag in ["latest, beta"]
]
```

Or even better: Add a macro to the `tools/build` folder :)

chrismatix (commented):
Good stuff, looking forward to the charts 👍

backend/api/auth/BUILD
pascal-airy: but when I did this I got no such target

lucapette (approved):
:shipit:

PR #304, opened 2020-11-16
Add APP_IMAGE_TAG variable

ljupcovangelski:


lucapette (changes requested):
`CORE_VERSION=beta scripts/bootstrap.sh` would read much nicer. `APP_IMAGE_TAG` leaks internals to the public API

pascal-airy (approved):

PR #375, opened 2020-11-23
[324] Reflect in the docs our understanding of the release process

### LINKED ISSUE #324: Hotfixes that don't have an issue break our convention of putting the issue in the merge message
We should discuss if we loosen our convention or make it part of a hotfix to always create an issue for it

lucapette:
Fixes #324

Furthermore, this commit prepares the code base so that we can automate
release notes (by introducing more explicit conventions for branches).


chrismatix (dismissed):
LGTM

docs/docs/guidelines/release-process.md
```diff
@@ -12,11 +12,11 @@ Here's an outline of the process:
...
 - Once release days comes, we execute the following steps:
-  - We create a release branch from `develop`
-  - The branch must follow the convention `release/x.y.z`
+  - We create an issue "Release x.y.z"
+  - We create a release branch `release/x.y.z` from `develop`
   - We test our release and any additional hot-fix is committed directly to the release branch
-  - Once we're satisfied with the release, we update the `VERSION` file with the current release number
-  - We merge the release branch into `main`, tag `main` following the same naming convention and push to `main`
+  - Once we're satisfied with the release, we update the `VERSION` file with the current release number. The commit message must be `Fixes #issue-number` where `issue-number` is the number of the current release issue.
+  - We merge the release branch into `main`, tag `main` with `x.y.z`and push to `main`
```
ljupcovangelski, line 19: The images pushed from the `release/x.y.z` branch already have the tag `release`. So we can test the whole process by running `AIRY_VERSION=release ./scripts/bootstrap.sh`. 

However, if we merge to main and wait for the images to be pushed with the `latest` tag, this can put us in a period of 15-20 minutes of inconsistency (while the images with latest are built and pushed). As the images are already pushed, maybe would be better if we have a script which labels all of the `release` images with `latest` and then right after we push to `main`. It just shortens the period I think.

lucapette: I will note down how to start the bootstrap process in release mode. 

As for the inconsistency, short-term I would probably not do anything. We're going to "test" the release process many times in the coming weeks, so I would wait that it's stable enough and then automate it all (including the tagging)

chrismatix (approved):

docs/docs/guidelines/contributing.md
```diff
@@ -0,0 +1,161 @@
...
+To keep a clean track of what is being released, every feature must contain only
+one commit when merged. The commit message of the squashed commit is very
+important, since it will be used to keep track of the features in a release.
+
+The conventional format is: `[#issue] - description`. For the example, if your
+pull requests refers to the issue "Introduce meaning of life" with number 42,
+the squashed commit message must be:
+
+```
+[#42] Introduce meaning of life
```
paulodiniz, line 158: [#42] - Introduce meaning of life

chrismatix: Let's hope that's not all there is to life.

paulodiniz (approved):

AitorAlgorta (approved):
👏

Issue #408, opened 2020-11-30
Speed-up the bootstrap process


ljupcovangelski:
As all of the apps start now at the same time, we need to optimize the sequence in which the `Airy apps` are being started, so that we prioritize the apps which must there in the first run.

lucapette:
this was done in #410

PR #410, opened 2020-11-30
[#408] Prioritize first apps

ljupcovangelski:


lucapette (approved):
CPU/RAM usage of the vagrant box feels much better now 👏

pascal-airy (approved):

Issue #411, opened 2020-11-30
Add /metadata.set Endpoint


paulodiniz:
The endpoint `/metadata.set` must be able to set a specific value to a key 

```
POST /metadata.set

{
  "conversation_id": <conversation-id>,
  "key": "foo.bar",
  "value": "baz"
}
```

Issue #413, opened 2020-11-30
Add /metadata.remove endpoint


paulodiniz:
The endpoint `/metadata.remove` must be able to remove a specific value from a key

```
POST /metadata.remove

{
  "conversationId": <conversation-id>,
  "key": "foo.bar"
}
```

PR #414, opened 2020-11-30
[#413] Add /metadata.set endpoint

paulodiniz:
Fixes https://github.com/airyhq/airy/issues/411

### DIFF
--- a/backend/api/communication/src/test/java/co/airy/core/api/communication/MetadataControllerTest.java
+++ b/backend/api/communication/src/test/java/co/airy/core/api/communication/MetadataControllerTest.java
@@ -90,13 +90,5 @@ public class MetadataControllerTest {
                         .andExpect(status().isOk()),
                 "Error setting metadata"
         );
-
-        retryOnException(
-                () -> webTestHelper.post("/metadata.set",
-                        "{\"conversation_id\":\"" + conversationId + "\", \"key\": \"__key__\", \"value\": \"awesome-value\"}",
-                        "user-id")
-                        .andExpect(status().isOk()),
-                "Error setting metadata"
-        );
     }
 }
--- a/backend/api/communication/src/test/java/co/airy/core/api/communication/MetadataControllerTest.java
+++ b/backend/api/communication/src/test/java/co/airy/core/api/communication/MetadataControllerTest.java
@@ -90,5 +90,13 @@ public class MetadataControllerTest {
                         .andExpect(status().isOk()),
                 "Error setting metadata"
         );
+
+        retryOnException(
+                () -> webTestHelper.post("/metadata.set",
+                        "{\"conversation_id\":\"" + conversationId + "\", \"key\": \"__private.key__\", \"value\": \"awesome-value\"}",
+                        "user-id")
+                        .andExpect(status().isBadRequest()),
+                "Error setting metadata"
+        );
     }
 }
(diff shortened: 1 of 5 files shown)


lucapette (changes requested):
The code looks great but we need docs 👼

docs/docs/api/http.md
```diff
@@ -643,3 +643,22 @@ The response comes in two parts:
   - `total`
 
     The total number of elements across all pages.
+
+### Metadata
+
+Airy provides a mechanism to add metadata to a conversation. A metadata entry consists of a key and a value.
+A key can use the dot notation to represent namespaces.
```
lucapette, line 650: I would have the defintions in the glossary and then refer to them from here (sorry I didn't realise I could have suggested this in the previous pass).

docs/docs/glossary.md
```diff
@@ -121,7 +121,16 @@ Header data contains information that is important for downstream processing. It
...
 
 A user represents one authorized agent in the Airy Core Platform.
+
+# Metadata
+
+Metadata is data attached to a conversation consisting of a set of Key/Value pairs. A key can use the dot notation to represent namespaces.
+
+e.g.
+  - Key: "sender.id" Value: "123"
+  - Key: "sender.contact.first_name" Value: "Grace"
```
paulodiniz, line 134: Urgh no. Will think of a better presentation

backend/api/communication/src/test/java/co/airy/core/api/communication/MetadataControllerTest.java
```diff
@@ -0,0 +1,102 @@
...
+                        .andExpect(status().isOk()),
+                "Error setting metadata"
+        );
+
+        retryOnException(
+                () -> webTestHelper.post("/metadata.set",
+                        "{\"conversation_id\":\"" + conversationId + "\", \"key\": \"__key__\", \"value\": \"awesome-value\"}",
+                        "user-id")
+                        .andExpect(status().isOk()),
+                "Error setting metadata"
```
lucapette, line 99: not needed anymore as far as I can say

docs/docs/api/http.md
```diff
@@ -643,3 +643,24 @@ The response comes in two parts:
...
     The total number of elements across all pages.
+
+### Metadata
+
+Please refer to our [metadata](glossary.md#metadata) definition for more
+information.
+
+### Setting metadata
+
+Airy provides a mechanism to add metadata to a conversation.
```
lucapette, line 654:
```suggestion
```

Yes I'm suggesting to remove... but mostly because I cant come up with anything constructive and probably we can get away without it anyway (if we want to keep a description... we'll have to change the subject because it reads as we as a company provide that but it's the platform)

docs/docs/api/http.md
```diff
@@ -643,3 +643,24 @@ The response comes in two parts:
...
+
+`POST /metadata.set`
+
+```json
+{
+  "conversation_id": "conversation-id",
+  "key": "source.contact.first_name",
```
lucapette, line 661:
```suggestion
  "key": "ad.id",
```

Proposing a different example so we don't confuse readers with actual keys we use internally

lucapette (changes requested):
Minor comments but looking great!

chrismatix (approved):

lucapette (approved):
:shipit:

PR #420, opened 2020-12-01
[#290] Channels page

### LINKED ISSUE #290: Channels page
Create a page to manage all kind of channels that the system have.

bitboxer:
closes #290


frontend/demo/src/actions/user/index.ts
```diff
@@ -30,12 +30,13 @@ export function loginViaEmail(requestPayload: LoginViaEmailRequestPayload) {
   return async (dispatch: Dispatch<any>) => {
     return doFetchFromBackend('users.login', requestPayload)
       .then((response: UserPayload) => {
-        dispatch(setCurrentUserAction(userMapper(response)));
-        return true;
+        const user = userMapper(response);
+        dispatch(setCurrentUserAction(user));
+        return Promise.resolve(user);
```
bitboxer, line 35: I am not sure what the generated code will return... @chrismatix maybe you can write here what is wrong?

frontend/demo/src/api/airyConfig.ts
```diff
@@ -1,15 +1,16 @@
 import {getAuthToken} from './webStore';
 
 export class AiryConfig {
-  static API_URL = 'airy.api';
+  static API_URL = 'http://api.airy';
   static NODE_ENV = process.env.NODE_ENV;
+  static FACEBOOK_APP_ID = 'CHANGE_ME';
```
bitboxer, line 6: We need to move this to a config file.

frontend/demo/src/actions/channel/index.ts
```diff
@@ -0,0 +1,78 @@
...
+export const setCurrentChannelsAction = createAction(SET_CURRENT_CHANNELS, resolve => (channels: Channel[]) =>
+  resolve(channels)
+);
+
+export const addChannelsAction = createAction(ADD_CHANNELS, resolve => (channels: Channel[]) => resolve(channels));
+
+export function getChannels() {
+  return async (dispatch: Dispatch<any>) => {
+    return doFetchFromBackend('channels.list')
+      .then(response => {
```
AitorAlgorta, line 27: We need to type the responses also. If we type them and the backend changes something it will automatically scream and won't lead to odd issues.

bitboxer:
Screen Shot 2020-12-01 at 14 00 44

frontend/demo/src/model/Channel.ts
```diff
@@ -0,0 +1,56 @@
+export interface Channel {
+  name: string;
+  source: string;
+  source_channel_id: string;
```
AitorAlgorta, line 4: we want to use camelCase in out interfaces, that's why we have the mapper, to map from backend code (snake_case) to frontend code (camelCase)

frontend/demo/src/model/Channel.ts
```diff
@@ -0,0 +1,56 @@
...
+    }
+  );
+};
+
+export const channelMapper = (payload: ChannelPayload): Channel => {
+  return {
+    name: payload.name,
+    source: payload.source,
+    source_channel_id: payload.source_channel_id,
+    image_url: payload.image_url,
```
AitorAlgorta, line 53: it should be camelCase

frontend/demo/src/pages/Channels/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+type ChannelsConnectProps = {} & ConnectedProps<typeof connector> & RouteComponentProps;
+
+const Channels = (props: ChannelsConnectProps) => {
+  const [facebookToken, setFacebookToken] = useState('');
+  useEffect(() => {
+    props.getChannels();
+  }, []);
+
+  const connect = token => {
```
AitorAlgorta, line 38: let's try to strong type everything. Pretty sure we will miss some, but let's try

AitorAlgorta (changes requested):
We need to try to strong type everything we can, responses in api requests included. We also need to avoid storing snake_case keys in redux, let's use the mapper for this.

frontend/demo/src/actions/user/index.ts
chrismatix: Sorry, I don't understand the question. For now we only generate typings for the message content model.

frontend/demo/src/api/airyConfig.ts
chrismatix: Late to the discussion, but because this is served by an nginx pod, which has access to the env, we could also inject these values using lua scripts https://github.com/bungle/lua-resty-template cc @lucapette and @AitorAlgorta

package.json
```diff
@@ -5,22 +5,25 @@
...
     "@types/react-router-dom": "^5.1.0",
+    "@types/react": "16.9.34",
     "core-js": "3",
     "emoji-mart": "^3.0.0",
     "linkifyjs": "^2.1.9",
-    "preact": "^10.5.7",
+    "lodash-es": "4.17.15",
     "preact-router": "^3.2.1",
-    "react": "16.12.0",
+    "preact": "^10.5.7",
```
chrismatix, line 19: Would be great (but optional) if you could fix this version

chrismatix (approved):

frontend/demo/src/model/Channel.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+export interface ExploreChannelRequestPayload {
+  source: string;
+  token: string;
+}
+
+export interface ConnectChannelRequestPayload {
+  source: string;
+  source_channel_id: string;
```
AitorAlgorta, line 24: shouldn't be all of this camelCase now?

frontend/demo/src/actions/user/index.ts
bitboxer: so its up to us to generate the code that talks to the api? then i will keep it this way.

frontend/demo/src/model/Channel.ts
bitboxer: if we want to do it that way, I need another mapper into this. Right now I am confused what the generator will generate, what we have to do and who will write what.

frontend/demo/src/api/airyConfig.ts
bitboxer: Yeah, could be one possibility. Let's discuss this in [this](https://github.com/airyhq/airy/issues/421) ticket.

package.json
bitboxer: We talked about it. This is just here because it was resorted. Leaving it as is.

frontend/demo/src/model/Channel.ts
bitboxer: talked with chris. not longer confused :D

frontend/demo/src/reducers/data/channels/index.ts
```diff
@@ -0,0 +1,26 @@
...
+import * as actions from '../../../actions/channel';
+import {unionWith} from 'lodash-es';
+
+type Action = ActionType<typeof actions>;
+
+export const initialState = [];
+
+const mergeChannels = (channels: Channel[], newChannels: Channel[]) =>
+  unionWith(newChannels, channels, (channelA: Channel, channelB: Channel) => {
+    return channelA.sourceChannelId === channelB.sourceChannelId;
```
chrismatix, line 12: Just out of curiosity: Why the `sourceChannelId` and not the channel id? 🤔 We now ensure on the backend that source channel ids are unique

chrismatix (dismissed):
LGTM with minor question

AitorAlgorta (approved):

frontend/demo/src/reducers/data/channels/index.ts
bitboxer: Because the api returns the source channel id and not always the `id` and the `source_channel_id` is always there. See docs: https://docs.airy.co/api/http#channels .

If this is wrong I am happy to change it.

chrismatix: The only one that does not return the id, is `channels.explore` as far as I can tell from the docs. So I would suggest we change this.

bitboxer: Yes, should I open a PR for that?

chrismatix (approved):

PR #426, opened 2020-12-01
[#425] Migrate Code formating to Bazel tools

### LINKED ISSUE #425: Migrate code formating rules to Bazel tools
We are moving our Bazel tooling to https://github.com/airyhq/bazel-tools 

We want our code formating to use said rules.

chrismatix:
fixes #425


chrismatix:
Waits for https://github.com/airyhq/bazel-tools/pull/6

chrismatix:
@lucapette A problem I noticed when doing this locally: IDE plugins can no longer pickup the lint configuration files. I think the implication is that we have to duplicate them for every code base (n=2). What do you think?

lucapette:
@chrismatix yeah at first it doesn't sound so good you have to duplicate the files... but in a way you can consider that a feature like "hey we bring you the tooling but maybe you want to choose your own config". The "best" would be you can do both (as in bringing the default) but current proposal is good enough!

bitboxer:
For me not having format on save would be a bit troublesome. Is there no way to get it back? Can't you point bazel to the config in the root?

chrismatix:
@bitboxer So here's is how I am doing it in this PR: We are keeping the `.prettierrc.json` (it's barely any config). But @lucapette and @paulodiniz  for CheckStyle I think we should use the HTTP URL feature of the IntelliJ plugin. What do you think?

CONTRIBUTING.md
```diff
@@ -1,9 +1,9 @@
 # Contributing
 
-We love every form of contribution. By participating to this project, you
+We love every form of contribution. By participating in this project, you
 agree to abide to the `Airy` [code of conduct](/code_of_conduct.md).
 
-Please read to our [contribuing guide](/docs/docs/contributing.md) to
+Please read to our [contributing guide](/docs/docs/guidelines/contributing.md) to
```
pascal-airy, line 6:
```suggestion
Please read our [contributing guide](/docs/docs/guidelines/contributing.md) to
```

chrismatix: Wow, so I found two issues in that sentence, but missed the third :D Thank you!

pascal-airy (approved):

PR #430, opened 2020-12-03
[#413] Add /metadata.remove endpoint

paulodiniz:


docs/docs/api/http.md
```diff
@@ -663,3 +662,16 @@ information.
...
+`POST /metadata.remove`
+
+```json
+{
+  "conversation_id": "conversation-id",
+  "key": "ad.id"
+}
+```
+
+This endpoint returns `200` if the operation was successful and `400` otherwise.
```
chrismatix, line 677: I don't think this can return 400, right? 🤔

paulodiniz: Typo, will change it

chrismatix (approved):

PR #447, opened 2020-12-07
Add go lib

pascal-airy:


lib/go/main_test.go
```diff
@@ -0,0 +1,24 @@
+package main
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGetFaces(t *testing.T) {
```
lucapette, line 10:
```suggestion
func TestSignupt *testing.T) {
```

lib/go/payloads.go
```diff
@@ -0,0 +1,27 @@
+package main
+
+type SignupRequestPayload struct {
```
lucapette, line 3: I know it's more tedious to write it manually _but_ we'll probably end up generating this code so it's better to move the payloads to their own package, one payload per file

lib/go/go.mod
```diff
@@ -0,0 +1,5 @@
+module cli
```
lucapette, line 1: this should probably be something like `api-client`

the path must also reflect that `lib/go/api-client` and we need to build the lib with bazel (so we need a BUILD file)

lucapette (commented):
First pass!

Issue #452, opened 2020-12-08
Split auth dependency


chrismatix:
**Current state:**
The `spring-test` library depends on `spring-auth` in order to use the `Jwt` class. This means that tests that use this library will also inject the authentication filter into the app they are testing.

The result is that apps could look like they require authentication in tests even though the `spring-auth` dependency is missing from the prod app.

**Proposed solution**
Split the auth bundle, so that `spring-test` can import the `Jwt` code only.

PR #456, opened 2020-12-08
[#452] Split auth dependency

paulodiniz:


lib/java/jwt/BUILD
```diff
@@ -0,0 +1,22 @@
...
+    "//:spring",
+    "//:springboot_security",
+    "//lib/java/log",
+    "//lib/java/spring/core:spring-core",
+    "@maven//:javax_servlet_javax_servlet_api",
+    "@maven//:javax_xml_bind_jaxb_api",
+]
+
+# Due to a bug in the abbreviation checkstyle rule
+# we need to skip style checking on this lib
```
chrismatix, line 16: Here too? :) Which abbreviation breaks?

lib/java/jwt/src/main/java/co/airy/jwt/Jwt.java
```diff
@@ -20,7 +20,7 @@
 
 @Component
 public class Jwt {
-    private static final Logger log = AiryLoggerFactory.getLogger(Jwt.class);
+    private static final Logger log = AiryLoggerFactory.getLogger(co.airy.jwt.Jwt.class);
```
chrismatix, line 23: This can stay `Jwt.class` I think

backend/api/communication/BUILD
```diff
@@ -11,6 +11,7 @@ app_deps = [
     "//backend:channel",
     "//backend:metadata",
     "//backend:read-receipt",
+    "//lib/java/jwt:airy-jwt",
```
chrismatix, line 14: If you make `spring-auth` export `//lib/java/jwt:airy-jwt` (usually we avoid this, but here it seems ok) you only have to change the dependencies of the test library 🙏

paulodiniz: NICE 👍  TIL

lib/java/spring/jwt/BUILD
```diff
@@ -0,0 +1,20 @@
...
+    "//:lombok",
+    "//:spring",
+    "//:springboot_security",
+    "//lib/java/log",
+    "//lib/java/spring/core:spring-core",
+    "@maven//:javax_servlet_javax_servlet_api",
+    "@maven//:javax_xml_bind_jaxb_api",
+]
+
+java_library(
```
chrismatix, line 15: This should use our own `custom_java_library`:
```
load("//tools/build:java_library.bzl", "custom_java_library")
```

chrismatix (approved):

PR #477, opened 2020-12-11
Several initial updates

olia-la:
- Corrected a few typos and spelling issues.
- Brought headers to the sentence case (as opposed to title case). It's a better practice to use sentence case. 
- Some endpoints miss a definition or at least some explanation (especially towards the end of the document). 
- `Reset password` endpoint mentions a reset password token - not clear what it is and how to get it. Add this info or call the token in a different way.
- `Tag a conversation` endpoint mentions an _existing tag_. Not clear where it should exist, how it's related etc. - add this info to avoid confusion. Especially because there is a whole section `Tags` further in the document, are those the same tags? Or different tags?

### DIFF
--- a/docs/docs/api/http.md
+++ b/docs/docs/api/http.md
@@ -3,34 +3,31 @@ title: API
 sidebar_label: HTTP
 ---
 
-This documents aims to offer an high-level overview of the different parts that
-compose our API.
+This documents offers a high-level overview of the different parts that
+compose the Airy API.
 
 ## Introduction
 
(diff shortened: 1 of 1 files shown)


lucapette:
@olia-la thank you very much for these improvements. I will need some time to go over your questions so I will leave the pull request open as a to do item for myself. Looking forward to merging it!

lucapette (approved):
I created follow up tickets for your suggestions. Thank you very much for your contribution!

PR #514, opened 2020-12-15
[#499] Future-proof metadata model

### LINKED ISSUE #499: Refactor metadata model
**Problem** Metadata is too conversation specific right now, even though it could work great in other places
**Problem** Metadata can not be log compacted, because it is keyed by conversation id. This means that if a noisy source keeps sending us conversation context (e.g. google) the topic will grow without bounds

**Current data model**
Key: conversation id
value:
```
{"name": "conversationId", "type": "string"},
{"name": "key", "type": "string"},
{"name": "value", "type": "string"},
{"name": "actionType", "type": {"name": "MetadataActionType", "type": "enum", "symbols": ["SET", "REMOVE"]}},
{"name": "timestamp", "type": "long", "logicalType": "timestamp-millis"}
```

**Proposed data model**
Key: UUIDv5 of (subject, metadata key)
Value:
```
{"name": "subject", "type": "string"}, // e.g. "conversation:conversation-uuid", or "channel:channel-uuid"
{"name": "key", "type": "string"},
{"name": "value", "type": "string"},
{"name": "createdAt", "type": "long", "logicalType": "timestamp-millis"},
{"name": "updatedAt", "type": "long", "logicalType": "timestamp-millis"}
```

The subject consists of a namespace (e.g. conversation or channel) that tells an app what this metadata is about. So you would build a conversation metadata table like so:

```
.filter((k,v) → v.getSubject().startsWith("conversation"))
.selectKey((k,v) → split on the ":" and get the id)
.aggregate(...)
```

It also consists of an id (e.g. channel or conversation id) so that you can aggregate on it.

**Implications:**
- No more `actionType` since we can simply send tombstone records instead of `REMOVE`
- We can log compact metadata down to the key level (storage becomes O(# of metadata values) vs O(# of metadata updates))
- We can use metadata for any subject

chrismatix:
fixes #499


paulodiniz (commented):
Nice! 👍 

The `toString()` naming is the only thing that catches my eyes, because I'm never in favor of overloading an operator that Java would use.

What do you think?

backend/avro/communication/metadata/src/main/java/co/airy/avro/communication/MetadataRepository.java
```diff
@@ -0,0 +1,29 @@
...
+        return new Subject(namespace, identifier);
+    }
+
+    public static UUID getId(Metadata metadataAction) {
+        return UUIDv5.fromNamespaceAndName(metadataAction.getSubject(), metadataAction.getKey());
+    }
+    public static UUID getId(Subject subject, String key) {
+        return UUIDv5.fromNamespaceAndName(subject.toString(), key);
+    }
+}
```
lucapette, line 29: As we're introducing this class (big 👍 for it), I'm thinking it may be nicer long term if MetadataAction creation details are all here, meaning we don't construct the object manually but have "shortcuts" like `MetadataRepository.newConversationMetadata(key, value)` (naming can be improved, I'm sure) `MetadataRepository.newChannelMetadata` and so on.

That way we could "hide" the `type:id` convention here and make the caller code a little more compact/readable

chrismatix: Will add these helpers.

paulodiniz (approved):

Issue #440, opened 2020-12-04
Add demo to the docs


lucapette:

lucapette:
We should extract from the demo readme (added in #518) and move to the contributing document in the context of this issue

lucapette:
closing in favor of #605

PR #518, opened 2020-12-15
Doc/489 how to run the frontend

### LINKED ISSUE #489: Documentation: How to run the frontend?
Improve https://github.com/airyhq/airy/tree/develop/frontend and https://github.com/airyhq/airy/tree/develop/frontend/demo

ogbeche77:
Fixes #489

The idea was to have a structure first and then we continue improving the frontend the documentation. e.g with gifs (To demonstrate the endpoint section). Please give feedbacks and open issues concerning the documentation so we can continually improve it.


frontend/README.md
```diff
@@ -1,32 +1,33 @@
+<p align="center">
+    <a href="https://airy.co/">
+    <img width="350" src="./assets/airy_primary_rgb.svg" alt="Airy Logo" />
+    </a>
+</p>
+
+# [Airy](https://airy.co/)- A Conversational Data Platform to drive sales, improve marketing and provide better service
```
lucapette, line 7: I wouldn't want to go for "the airy marketing pitch" in the readme of a subfolder of an open-source project. It reads really strange.

frontend/README.md
```diff
@@ -1,32 +1,33 @@
...
+
 # Frontend
 
-The `frontend` top-level folder of the Airy Core Platform
-[monorepo](https://en.wikipedia.org/wiki/Monorepo) contains the UI related code
-of the platform.
+The Frontend section of the Airy core platform contains the user interaction(UI) related code.
 
-Here is a quick introduction to the frontend projects:
+They are grouped into four frontend sub-projects:
```
lucapette, line 14: The previous sentence reads better

frontend/README.md
```diff
@@ -1,32 +1,33 @@
...
 
-- `showcase`
+- `Showcase`
 
   This directory contains a small React application that showcases the
-  components of the [Airy Components
-  Library](https://www.npmjs.com/package/@airyhq/components), we host the `main`
-  branch version at [https://components.airy.co](https://components.airy.co).
+  components of the [Airy Components Library](https://www.npmjs.com/package/@airyhq/components), we host the `main`
+  branch version [here](https://components.airy.co).
```
lucapette, line 24: Since we're touching this.... this is actually out of date as the components and the show case library are not there anymore so we should just remove all together.

frontend/README.md
```diff
@@ -1,32 +1,33 @@
...
 
-- `demo`
+- `Demo`
 
-  This project is a minimum UI implementation of the provided [Airy Core Platform API](https://docs.airy.co/api/http).
-  Unlike `showcase` it does not use the `npm` version of the `components` library, but instead uses the
-  local repository version.
+  The [Demo project](https://github.com/airyhq/airy/tree/develop/frontend/demo) is a minimum UI implementation of the provided [Airy Core Platform API](https://docs.airy.co/api/http). Unlike `showcase` it does not use the `npm` version of the `components` library, but instead uses the local repository version.
```
lucapette, line 28: this is not true anymore so if we're touching it we should update

frontend/demo/README.md
```diff
@@ -1,3 +1,75 @@
...
+If you're still not sure where to start, open a [new
+issue](https://github.com/airyhq/airy/issues/new) and we'll gladly help you get
+started.
+
+### Code of Conduct
+
+To ensure a safe experience and a welcoming community, the Airy Core Platform
+project adheres to the [contributor
+convenant](https://www.contributor-covenant.org/) [code of
+conduct](/code_of_conduct.md).
```
lucapette, line 75:
```suggestion
```

Yes suggesting to remove all together because it's redundant. If at all, we can refer back to the readme of the mono repo (but to be fair I wouldn't do that either)

lucapette (changes requested):
I left a round of comments about things I'm sure we have to change.

to be fair, I'm a little unsure about the readme updates as well. I understand the need: we want to document how to run the demo application but here's the problem:

- I don't think the readme is the right place for this information. It will run out of sync with the rest of the docs (and we'll heavily change how the bootstrap/interacting with the local cluster works in the coming weeks) because people will not look here
- The "right" place is the contributing guide of the docs (the idea is to have a sub section about the frontend apps there). 

the reason why we did not mention anything about the frontend apps yet in docs.airy.co is that we're waiting for the UI to have at least one messaging feature (the conversation list) and then we'd work on this issue https://github.com/airyhq/airy/issues/440 (I understand the issue says nothing... but truth to be told I'm taking care of the docs myself most of the time and I don't personally need more info as I know what's the plan. Obviously it's suboptimal and it's much nicer to get more contributions to the docs!)

So what I think we can do now is:

- Address the feedback I left already
- Merge the readme as is
- Update the above mentioned issue so that we say "extract content from the demo readme" and link the contributing section back there when we're ready

lucapette (approved):
:shipit:

PR #521, opened 2020-12-16
[#493] Route Google metadata to get displayname

### LINKED ISSUE #493: Ingest Google conversation metadata
Google sends conversation "context" along with messages. In the `EventsRouter` we should map this incoming context to metadata.

chrismatix:
fixes #493


backend/sources/google/events-router/src/main/java/co/airy/core/sources/google/EventsRouter.java
```diff
@@ -75,51 +85,78 @@ private void startStream() {
...
-                    GoogleEventInfo googleEventInfo = GoogleInfoExtractor.extract(webhookEvent);
-                    googleEventInfo.setEventPayload(sourceEvent);
-                    googleEventInfo.setTimestamp(Instant.parse(webhookEvent.getSendTime()).toEpochMilli());
+                    final EventInfo eventInfo = InfoExtractor.extract(webhookEvent);
+                    eventInfo.setEvent(webhookEvent);
+                    eventInfo.setTimestamp(Instant.parse(webhookEvent.getSendTime()).toEpochMilli());
 
-                    if (!webhookEvent.isMessage()) {
-                        return KeyValue.pair(googleEventInfo.getAgentId(), null);
+                    if (!webhookEvent.isMessage() && !webhookEvent.hasContext()) {
```
chrismatix, line 92: To discuss: Functionally we could let all records pass to the `flatMap`, but by filtering out anything that is either not a message or has no context we reduce the repartition store size. Is this worth it? cc @lucapette and @paulodiniz

lucapette (approved):
:shipit:

PR #528, opened 2020-12-16
[#494] Fetch Facebook metadata

### LINKED ISSUE #494: Fetch Facebook Metadata
Also addresses #482 

Facebook offers two APIs that can be used to retrieve metadata of a new conversation now:

https://developers.facebook.com/docs/messenger-platform/identity/user-profile/
https://developers.facebook.com/docs/graph-api/reference/conversation/#edges

Instead of determining if this conversation is new, we can appropriate the Facebook `sender` app to check for missing metadata.

- [x] Add aggregated metadata to the `context` table
- [x] Fetch profile information if certain metadata (e.g. first and last name) is missing
- [x] Give the sender a more generic name: `TODO` find a name ("worker"?)

chrismatix:
fixes #494


backend/sources/facebook/sender/src/main/java/co/airy/core/sources/facebook/Sender.java
```diff
@@ -64,27 +97,41 @@ public void startStream() {
...
-                })
+                .join(conversationTable, (message, conversation) -> new SendMessageRequest(conversation, message))
                 .mapValues(this::sendMessage)
                 .to(new ApplicationCommunicationMessages().name());
 
+        // Fetch missing metadata
+        conversationTable
+                // To avoid any redundant fetch contact operations the suppression interval should
+                // be higher than the timeout of the Facebook API
+                .suppress(Suppressed.untilTimeLimit(Duration.ofMillis(streams.getSuppressIntervalInMs()), Suppressed.BufferConfig.unbounded()))
```
chrismatix, line 115: This suppression is optional, but we can use it to minimize the risk of a new conversation requesting the contact data too many times. If we stream the table without suppression this app could be badly behaved and call the Facebook API multiple times when a stream of messages for a new conversation comes in.

backend/sources/facebook/sender/src/main/java/co/airy/core/sources/facebook/Sender.java
```diff
@@ -101,9 +148,49 @@ private Message sendMessage(SendMessageRequest sendMessageRequest) {
...
+            final Metadata lastName = newConversationMetadata(conversationId, MetadataKeys.Source.Contact.LAST_NAME, profile.getLastName());
+            recordList.add(KeyValue.pair(getId(lastName).toString(), lastName));
+        }
+
+        if (profile.getProfilePic() != null) {
+            final Metadata avatarUrl = newConversationMetadata(conversationId, MetadataKeys.Source.Contact.AVATAR_URL, profile.getProfilePic());
+            recordList.add(KeyValue.pair(getId(avatarUrl).toString(), avatarUrl));
+        }
+
+        final String fetchSuccessState = recordList.size() > 0 ? "ok" : "failed";
```
chrismatix, line 177: Optionally we could add retry counts here

backend/sources/facebook/sender/src/main/java/co/airy/core/sources/facebook/services/Api.java
```diff
@@ -54,6 +61,55 @@ public void sendMessage(final String pageToken, SendMessagePayload sendMessagePa
         restTemplate.postForEntity(fbReqUrl, new HttpEntity<>(sendMessagePayload, httpHeaders), FbSendMessageResponse.class);
     }
 
+    // See "Retrieving a Person's Profile" in https://developers.facebook.com/docs/messenger-platform/identity/user-profile
+    public UserProfile getProfileFromContact(String sourceConversationId, String token) {
```
chrismatix, line 65: This code works on platform ™️

backend/sources/facebook/connector/src/main/java/co/airy/core/sources/facebook/Connector.java
```diff
@@ -0,0 +1,218 @@
...
+            log.error(String.format("Failed to send a message to Facebook \n SendMessageRequest: %s", sendMessageRequest), e);
+        }
+
+        updateDeliveryState(message, DeliveryState.FAILED);
+        return message;
+    }
+
+    private boolean needsMetadataFetched(String conversationId, Conversation conversation) {
+        final Map<String, String> metadata = conversation.getMetadata();
+        final String fetchState = metadata.get(MetadataKeys.Source.CONTACT_FETCH_STATE);
```
lucapette, line 153: at glance, I feel like `MetadataKeys.Source.Contact.FETCH_STATE` would be a better name/place for this?

backend/sources/facebook/connector/src/main/java/co/airy/core/sources/facebook/Connector.java
```diff
@@ -0,0 +1,218 @@
...
+
+        return !"ok".equals(fetchState) && !"failed".equals(fetchState);
```
lucapette, line 155: I want to propose we have constants for this but can't think of a place apart  from `MetadataKeys.Source.Contact` (which I don't think it's a great place for it as we'd expect keys there...)

backend/sources/facebook/connector/src/main/resources/application.properties
```diff
@@ -2,3 +2,5 @@ kafka.brokers=${KAFKA_BROKERS}
 kafka.schema-registry-url=${KAFKA_SCHEMA_REGISTRY_URL}
 kafka.cleanup=${KAFKA_CLEANUP:false}
 kafka.commit-interval-ms=${KAFKA_COMMIT_INTERVAL_MS:30000}
+
+kafka.suppress-interval-ms=3000
```
lucapette, line 6:
```suggestion
kafka.suppress-interval-ms={KAFKA_SUPPRESS_INTERVAL_MS:3000}
```

may be a bit more future proof

backend/sources/facebook/connector/src/test/java/co/airy/core/sources/facebook/SendMessageTest.java
```diff
@@ -81,12 +82,7 @@ static void afterAll() throws Exception {
     @BeforeEach
     void beforeEach() throws InterruptedException {
         MockitoAnnotations.initMocks(this);
-
-        if (!streamInitialized) {
-            retryOnException(() -> assertEquals(worker.getStreamState(), RUNNING), "Failed to reach RUNNING state.");
-
-            streamInitialized = true;
-        }
+        retryOnException(() -> assertEquals(worker.getStreamState(), RUNNING), "Failed to reach RUNNING state.");
```
lucapette, line 85: 🍃

infrastructure/helm-chart/charts/apps/charts/sources-facebook-sender/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
 appVersion: "1.0"
-description: A Helm chart for the Sources Facebook Sender app
-name: sources-facebook-sender
-version: 0.1.0
\ No newline at end of file
+description: A Helm chart for the Sources Facebook Connector app
```
lucapette, line 3:
```suggestion
description: A Helm chart for the Facebook Connector app
```

maybe?

lucapette (commented):
First pass. I like where this is going!

backend/sources/facebook/connector/src/main/java/co/airy/core/sources/facebook/Connector.java
chrismatix: This was my initial idea, too, but the issue is that for the `conversations.list` endpoint we take everything in `source.contact.*` and map it to a response. Since I wanted to hide this value, I moved it up in the hierarchy.

backend/sources/facebook/connector/src/main/java/co/airy/core/sources/facebook/Connector.java
chrismatix: Yes, this is a good point. I think I will add them there as an enum. Because you could argue that it can only be a key, if you actually get a `String`.

Then I'm wondering though if we should rename the whole class to something along the line of `MetadataStructure`. What do you think? 🤔

lucapette (approved):
:shipit:

PR #554, opened 2020-12-28
[#515] introduce eslint

### LINKED ISSUE #515: Introduce eslint
As discussed offline, I played around with `eslint` as we're all happy to introduce it here.

The following configuration: 

```
module.exports = {
    "env": {
        "browser": true,
        "es2021": true,
        "node": true
    },
    "extends": [
        "eslint:recommended",
        "plugin:react/recommended",
        "plugin:@typescript-eslint/recommended"
    ],
    "parser": "@typescript-eslint/parser",
    "parserOptions": {
        "ecmaFeatures": {
            "jsx": true
        },
        "ecmaVersion": 12,
        "sourceType": "module"
    },
    "plugins": [
        "react",
        "@typescript-eslint"
    ],
    "rules": {
       "@typescript-eslint/no-unused-vars": ["error", {"varsIgnorePattern": "_" }]
    },
      "settings": {
        "react": {
          "version": "16.12"
        }
    },
  "overrides": [
    {
      "files": [ "frontend/chat-plugin/**/*.tsx" ],
      "settings": {
        "react": {
          "pragma": "h"
        }
    }
    }
  ]
};
```

Solves the two most noisy problems: we need `_` for bazel/typescript to play nice and the `overrides` to avoid confusing eslint about the chat plugin (it uses preact).

You can see the remaining errors [here](https://gist.github.com/lucapette/ed83bfe2a4b1f20fd1bc6995569f3de5)

The most present error `warning  Missing return type on function           @typescript-eslint/explicit-module-boundary-types` seems to be React specific and maybe we can get away with configuration only. And I'm a bit unsure what's the best way to fix `warning  Unexpected any. Specify a different type          @typescript-eslint/no-explicit-any` because we use `Dispatch<any>` in quite a few places.

ogbeche77:
Fixes #515

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,45 @@
+{
+  "extends": ["eslint:recommended", "plugin:react/recommended", "plugin:@typescript-eslint/recommended"],
+  "plugins": ["react", "@typescript-eslint"],
+  "parser": "@typescript-eslint/parser",
+  "parserOptions": {
+    "ecmaFeatures": {
+      "jsx": true
+    }
+  },
+
+  "env": {
+    "browser": true,
+    "es2021": true,
+    "node": true
+  },
+
+  "rules": {
+    "react/prop-types": 0,
+    "@typescript-eslint/no-unused-vars": ["error", {"varsIgnorePattern": "_"}],
+    "@typescript-eslint/explicit-module-boundary-types": "off",
+    "@typescript-eslint/no-explicit-any": "off",
+    "@typescript-eslint/no-empty-interface": 0,
+    "@typescript-eslint/ban-types": ["error", {"extendDefaults": true, "types": {"{}": false, "Object": false}}]
+  },
+  "settings": {
+    "react": {
+      "version": "16.12"
+    }
+  },
+  "overrides": [
+    {
+      "files": ["frontend/chat-plugin/**/*.tsx"],
+      "settings": {
+        "react": {
+          "pragma": "h"
+        },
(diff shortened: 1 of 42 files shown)


lucapette (approved):
:shipit:

ogbeche77:
While most of the errors detected by eslint has been resolved. The two persistent warnings which eslint continues to return are `@typescript-eslint/explicit-module-boundary-types` and `@typescript-eslint/no-explicit-any`. 

Some of the `@typescript-eslint/explicit-module-boundary-types` warnings were solved in files where functional components(stateless components) had been implemented because eslint insisted on using explicit function return types. By just including the React Element (In this case JSX.Element) fixed this issue. This however, did not work for files in chatplugin as they all use preact?( Researching on this to find the equivalent).

In other files where the warnings `@typescript-eslint/explicit-module-boundary-types` still persists and class based components had been used, eslint warnings could be resolved by adding React Node to render.

 For `@typescript-eslint/no-explicit-any`, the warnings could be resolved by actually dispatching the action itself `dispatch: Dispatch<Action Type>` instead of using `dispatch: Dispatch<any>`. As this appears to be a coding style and not an error, I have decided to wait and get the feedbacks from other code owners to decide if we stick to this coding style or implement the change.

Another error `Property 'airy' does not exist on type 'Window & typeof globalThis'.` is still yet to be resolved. The temporary fix `declare const window: any;`doesn't solve the problem. I still till have to find a much better solution

Issue #558, opened 2020-12-28
Import core go libraries from anywhere


ljupcovangelski:
We need to make our go libraries importable from anywhere.

ljupcovangelski:
This is already part of #466

PR #559, opened 2020-12-28
[#558] Change public name of go modules

ljupcovangelski:


ljupcovangelski:
Another issue needs to be referenced.

Issue #400, opened 2020-11-27
Load messages of conversation


AitorAlgorta:
Load text messages of conversation and display them correctly

PR #567, opened 2020-12-29
[#400] Load messages of conversations

thorstenairy:
Solves #400


frontend/demo/src/pages/Inbox/Messenger/MessageList/Avatar.scss
```diff
@@ -0,0 +1,49 @@
+// .DisplayName {
```
bitboxer, line 1: This could be deleted

frontend/demo/src/pages/Inbox/Messenger/MessengerListItem/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+const connector = connect(mapStateToProps, null);
+
+const MessengerListItem = (props: MessengerListItemProps) => {
+  const {messageText, messageSenderType, message, lastMessages} = props;
+  const isUser = messageSenderType !== MessageSenderType.appUser;
+
+  const messageAvatar = (messageId: string) => {
+    Object.values(lastMessages).forEach(lastMessage => {
+      if (messageId === lastMessage.lastMessage.id) {
+        return <Avatar isLastMessage={true} avatarUrl={lastMessage.contact.avatarUrl} />;
```
bitboxer, line 32: Out of curiosity couldn't you inline the if statement to the `isLastMessage` and remove the need for two returns and the else?

frontend/demo/src/pages/Inbox/Messenger/MessengerListItem/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+        return <Avatar isLastMessage={false} />;
+      }
+    });
+  };
+
+  return (
+    <div className={styles.messageListItemContainer}>
+      <div className={styles.messageListItem}>
+        {!isUser ? (
+          <div className={styles.messageListItemMember}>
```
bitboxer, line 43: can this be deleted?

frontend/demo/src/pages/Inbox/Messenger/MessengerListItem/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+            {/* {messageText} */}
+            {/* {messageAvatar(message.id)} */}
+          </div>
+        ) : (
+          <div className={styles.messageListUserContainer}>
+            {console.log(messageAvatar(message.id))}
```
bitboxer, line 49: 💣

bitboxer (approved):
Really like this already. Some super minor comments

VERSION
```diff
@@ -1 +1 @@
-0.3.0
+0.4.0
```
lucapette, line 1: something wrong with this branch I think

lucapette (changes requested):

frontend/demo/src/pages/Inbox/ConversationListItem/index.tsx
```diff
@@ -31,8 +31,8 @@ const mapStateToProps = (state: StateModel) => {
 const connector = connect(mapStateToProps, null);
 
 const FormattedMessage = ({message}: FormattedMessageProps) => {
-  if (message && message.content) {
-    return <>{message.content.text}</>;
+  if (message && message.content[0]) {
```
AitorAlgorta, line 34: This needs improvement but should be enough to unblock the ticket

lucapette: true! We should probably already add a follow up ticket (that discussion is relevant for the chat plugin as well)

frontend/demo/src/actions/messages/index.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export const loadingMessagesAction = createAction(
+  MESSAGES_LOADING,
+  resolve => (messagesInfo: {conversationId: string; messages: Message[]}) => resolve(messagesInfo)
+);
+
+export function listMessages(conversationId: string) {
+  return async (dispatch: Dispatch<any>) => {
+    return HttpClient.listMessages({
+      conversation_id: conversationId,
```
lucapette, line 15: the listMessages input type could respect JS conventions right? We could do the mapping inside the call to the backend? I'm wondering if the rest of the library looks like that and I didn't notice

frontend/demo/src/pages/Inbox/Messenger/MessageList/Avatar.tsx
```diff
@@ -0,0 +1,25 @@
...
+  return avatarUrl && isLastMessage ? (
+    <div className={'avatarImage'}>
+      <img src={airyAvatarImage} />
+    </div>
+  ) : (
+    <div className={'noAvatar'}>
+      <img />
+    </div>
+  );
+};
```
lucapette, line 23: this would probably be a little more readable with an early return for the last message avatar and an explicit return for the no avatar (which I'd probably extract into a local `const noAvatar = () => ...`)

frontend/demo/src/pages/Inbox/Messenger/MessageList/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+    messageListRef.current.scrollTop = messageListRef.current.scrollHeight;
+  };
+
+  return (
+    <div className={styles.messageList} ref={messageListRef}>
+      {messages.map((message: Message) => {
+        return (
+          <MessageListItem
+            key={message.id}
+            messageText={message.content[0].text}
```
lucapette, line 64: it's probably the same here right? (I mean the comment about the content part)

frontend/demo/src/pages/Inbox/Messenger/MessengerContainer/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+const MessengerContainer = (props: MessengerContainerProps) => {
+  const {conversations, match} = props;
+
+  return (
+    <div className={styles.messengerContainer}>
+      {!conversations ? (
+        <div className={styles.emptyState}>
+          <h1>Your conversations will appear here as soon as a contact messages you.</h1>
+          <p>Airy Messenger only shows new conversations from the moment you connect your Facebook Pages.</p>
```
lucapette, line 26:
```suggestion
          <p>Airy Messenger only shows new conversations from the moment you connect at least one channel.</p>
```

frontend/demo/src/reducers/data/messages/index.ts
```diff
@@ -0,0 +1,46 @@
...
+import {Message} from 'httpclient';
+import {DataState} from '..';
+
+type Action = ActionType<typeof actions>;
+
+export type MessagesState = {
+  data: DataState;
+};
+
+export type MessageMap = {
```
lucapette, line 13: if I understand this correctly we could call the type `MessageById`

frontend/demo/src/reducers/data/messages/index.ts
```diff
@@ -0,0 +1,46 @@
...
+const initialState = {
+  all: {},
+};
+
+function organiseMessages(messages: Message[]): MessageMap {
+  const newMessages: MessageMap = {};
+  messages.forEach((message: Message) => {
+    newMessages[message.id] = message;
+  });
+  return newMessages;
```
lucapette, line 30: `_.keyBy(messages)` should do the same (we already have lodash right?)

lib/typescript/httpclient/endpoints/listConversations.ts
```diff
@@ -5,12 +5,12 @@ import {ConversationPayload} from '../payload/ConversationPayload';
 import {MessagePayload} from '../payload/MessagePayload';
 import {PaginatedPayload} from '../payload/PaginatedPayload';
 
-const messageMapper = (payload: MessagePayload): Message => {
+export const messageMapper = (payload: MessagePayload): Message => {
```
lucapette, line 8: the mappers are private "by choice" as there's no need to export this (from the perspective of the api lib I mean). I also can't find where we're using this. But I imagine we may need to extract it from this endpoint and use it in one than one place. When I discussed this with @AudreyKj we thought to wait until it was really needed (now it seems so) before introducing one more folder (`httpclient/mappers` ?). But the point of not making those importable from the lib stays

lib/typescript/httpclient/model/Message.ts
```diff
@@ -28,13 +30,36 @@ export enum MessageState {
   delivered = 'DELIVERED',
 }
 
+export enum MessageSenderType {
```
lucapette, line 33: as we call this `SenderType` it may be worth reflecting the naming in the frontend. Especially because we intend to simplify in #572 soon enough and it would be easier to catch this

lib/typescript/httpclient/model/Message.ts
```diff
@@ -28,13 +30,36 @@ export enum MessageState {
...
+  senderType: MessageSenderType;
+  sentAt: Date;
 }
+export interface MessagePayloadData {
+  data: MessagePayload[];
+}
+
+export const messageMapperData = (payload: MessagePayloadData): Message[] => {
+  const messages = [];
+  payload.data.forEach((messagePayload: MessagePayload) => {
```
lucapette, line 54: `return payload.data.map` so we can save the var?

lucapette (commented):
Looking very good! 

First pass,

lib/typescript/httpclient/payload/PaginatedPayload.ts
```diff
@@ -1,4 +1,5 @@
+import {ResponseMetadataPayload} from '../payload/ResponseMetadataPayload';
 export interface PaginatedPayload<T> {
   data: T[];
-  responseMetadata: {previousCursor: string; nextCursor: string; total: number};
+  responseMetadata: ResponseMetadataPayload;
```
lucapette, line 4: this type was also not introduced on purpose... it's not really needed right?

AitorAlgorta: why is not needed?

lucapette: because it's not really its own thing... it's always embedded in the paginatedpayload that is already a type...

frontend/demo/src/pages/Inbox/ConversationListItem/index.tsx
chrismatix: Since we will always have to deal with content lists, I think we need to reorganize the code so that the message renders as a wrapper that contains a mapped view of the content. Every single content block contains enough information to render it on its own.

Pseudo code:
```jsx
<Message>
{message.content.map((content) => renderMessageContent(content))}
</Message>
```

chrismatix: The challenge for the data model is that as a default we have to accommodate sources that send multiple content pieces per message. This is why we had to make content a list, which is easy to handle with maps :)

frontend/demo/src/reducers/data/messages/index.ts
AitorAlgorta: `return _.keyBy(messages, 'id');`

lucapette (approved):
:shipit:

Issue #522, opened 2020-12-16
Introduce HttpClient lib


lucapette:
The idea is to extract the code we use to communicate with the api into its own library so that, soon enough, we can ship a NPM package in sync with the version of core (and add a clients section to the api reference of docs.airy.co).

Important bits of information:

- `HttpClient` may not be the best name _but_ it's practical and keeps the door open for `WebSocketClient`. We will rename the go library as well (and call the java one the same when we introduce it)
- We place it in `lib/typescript/httpclient` or `lib/typescript/clients/http` (but I think we have to keep the name to `HttpClient` either way). The placement under `lib/typescript` is part of the larger effort of introducing clients in different languages. A bit more about that in the next point.
- We put _all_ the payloads in the name directory (say `lib/typescript/httpclient/payload`, one payload per file. I understand it's a little tedious to do this manually, but the idea is exactly we won't do this over time. We plan to generate the payloads with bazel (haven't gone over the details yet so there's no issue in github. But there's multiple ways to achieve this)

lucapette:
this was done in #571  we will open a new issue to discuss further steps

PR #571, opened 2020-12-29
Feature/522 introduce httpclient lib

AudreyKj:
httpclient lib: ref issue [522](https://github.com/airyhq/airy/issues/522)


frontend/demo/src/actions/channel/index.ts
```diff
@@ -1,20 +1,13 @@
...
-  ChannelsPayload,
-  channelsMapper,
-  channelMapper,
-  connectChannelApiMapper,
-  disconnectChannelApiMapper,
   ConnectChannelRequestPayload,
   ExploreChannelRequestPayload,
   DisconnectChannelRequestPayload,
-} from '../../model/Channel';
+  AiryHttpClient,
```
lucapette, line 9: I would drop the `Airy` prefix (It's a comment I leave in most pull requests now so I'm considering adding some naming convention guide)

frontend/demo/src/actions/conversations/index.ts
```diff
@@ -1,9 +1,12 @@
 import {Dispatch} from 'redux';
 import {createAction} from 'typesafe-actions';
-import {doFetchFromBackend} from '../../api/airyConfig';
-
-import {Conversation, ConversationPayload, conversationsMapper} from '../../model/Conversation';
-import {ResponseMetadata} from '../../model/ResponseMetadata';
+import {
+  AiryHttpClient,
+  Conversation,
+  conversationsMapper,
```
lucapette, line 6: Gut feel tells me the mapper is a concern of the demo app more than the library. But, as I said, it's not structured feedback as I'm not sure I understand the code well enough so the comment is more to initiate a conversation around it

frontend/demo/src/actions/conversations/index.ts
```diff
@@ -34,17 +37,10 @@ export const removeErrorFromConversationAction = createAction(
...
-  metadata: ResponseMetadata;
-}
-
 export function fetchConversations() {
   return async (dispatch: Dispatch<any>) => {
     dispatch(loadingConversationsAction());
-    return doFetchFromBackend('conversations.list', {
-      page_size: 10,
-    })
+    return AiryHttpClient.fetchConversations()
```
lucapette, line 43: we should expose a payload that reflects https://docs.airy.co/api/http#list-conversations . To be fair the only way to know the exact payload at the moment is probably looking at the backend code. The point stays though: it's a library so we should expose, for example, the page size (maybe have a default sure) .

lucapette (changes requested):
OK did a first pass, Loving where we're going with this!

Happy to see we also moved the models. There's a few general things:

- I'm unsure about the mappers being either here or exposed (It feels like they should be private code... not sure 100%). But I'd delay the decision there till we address the existing comments and the next point (that way the code will look close to its final stage and will give us some clarity
- the payload dir should contain one file per payload (and nothing else) if we want that to be future proof (in terms of auto generation). So what I think would work is a folder structure like:
  - `payload`
  - `model`
  - `endpoints`

lib/typescript/httpclient/BUILD
```diff
@@ -0,0 +1,16 @@
...
+    name = "httpclient",
+    deps = [
+        "//lib/typescript/types",
+        "@npm//@types/node",
+        "@npm//lodash-es",
+        "@npm//react",
+        "@npm//redux",
+        "@npm//redux-starter-kit",
+        "@npm//typesafe-actions",
+    ],
```
lucapette, line 15: unless I'm misunderstanding the code, we shouldn't need more than the types as deps here?

lib/typescript/httpclient/README.md
```diff
@@ -0,0 +1,40 @@
...
+- fetchNextConversations
+
+TAGS 
+- getTags
+- createTag
+- updateTag
+- deleteTag
+
+USER 
+- loginViaEmail
```
lucapette, line 37: <3

lib/typescript/httpclient/BUILD
AudreyKj: yes totally, forgot to delete those

lib/typescript/httpclient/endpoints/getTags.ts
```diff
@@ -0,0 +1,43 @@
...
+const tagsMapper = (serverTags: Tag[]): Tag[] => {
+  const tags: Tag[] = [];
+  const _ = serverTags.map((tag: Tag) => {
+    tag.color = colorMapper(tag.color);
+    tags.push(tag);
+  });
+  return tags;
+};
+
+export function getTags() {
```
lucapette, line 34: i'd rename in style of the conversations changes. `listTags` (same for the payloads)

lib/typescript/httpclient/payload/ListConversationsResponse.ts
```diff
@@ -0,0 +1,7 @@
+import {ConversationPayload} from './ConversationPayload';
+import {ResponseMetadata} from './ResponseMetadata';
+
+export interface ListConversationsResponse {
```
lucapette, line 4:
```suggestion
export interface ListConversationsResponsePayload {
```

lucapette (changes requested):
Second pass!

This is looking better and better

lib/typescript/httpclient/model/Tag.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export interface ModalType {
+  modal: {
+    type: string;
+    tagId: string;
+    tagName: string;
+    delete: string;
+    error: string;
+  };
+}
```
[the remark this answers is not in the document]
AudreyKj: for the types on Tags: yes they are def part of the demo app but they are reused across several pages / components so maybe we could create a Tag types file in demo for this local use

lib/typescript/httpclient/payload/index.ts
```diff
@@ -0,0 +1,17 @@
...
+export * from './ExploreChannelRequestPayload';
+export * from './ListConversationsResponsePayload';
+export * from './MessagePayload';
+export * from './ResponseMetadata';
+export * from './LoginViaEmailRequestPayload';
+export * from './UserPayload';
+export * from './ListTagsResponsePayload';
+export * from './TagPayload';
+export * from './CreateTagRequestPayload';
+export * from './ConversationListRequestPayload';
```
lucapette, line 17: `ListConversationsRequestPayload` would follow conventions

lib/typescript/httpclient/endpoints/listTags.ts
```diff
@@ -0,0 +1,28 @@
+import {doFetchFromBackend} from '../api';
+import {Tag, ColorTag} from '../model';
+import {ListTagsResponsePayload} from '../payload';
+
+const colorMapper = (color: string): ColorTag => {
+  const colors = ['BLUE', 'RED', 'GREEN', 'PURPLE'];
```
lucapette, line 6: I'd move out of the method definition. In order to simplify the code further we can also do something like:

```json5
  const tagMapper = {
  'BLUE': 'tag-blue',
  'RED': 'tag-red'
  // etc...
  }
```

that way we could remove the function all together as you could `tagMapper[color.toLowerCase()] || 'tag-blue'`

lib/typescript/httpclient/endpoints/listTags.ts
```diff
@@ -0,0 +1,28 @@
...
+  } else {
+    return 'tag-blue';
+  }
+};
+
+const tagsMapper = (serverTags: Tag[]): Tag[] => {
+  return serverTags.map(t => {
+    return {id: t.id, name: t.name, color: colorMapper(t.color)};
+  });
+};
```
lucapette, line 18: the map function here can be inlined (we'd save the return and the the single line would be good enough, not too long)

lib/typescript/httpclient/endpoints/loginViaEmail.ts
```diff
@@ -0,0 +1,24 @@
...
+const userMapper = (payload: UserPayload): User => {
+  const user: User = {
+    id: payload.id,
+    firstName: payload.first_name,
+    lastName: payload.last_name,
+    displayName: payload.first_name + ' ' + payload.last_name,
+    token: payload.token,
+  };
+  return user;
+};
```
lucapette, line 14: Not that I would necessarily change the code here, but I'm wondering if `return {...` would also work (given we're specifying the return type of the function anyway)

lib/typescript/httpclient/payload/index.ts
```diff
@@ -0,0 +1,17 @@
+export * from './ChannelApiPayload';
+export * from './ChannelsPayload';
+export * from './ConnectChannelRequestApiPayload';
+export * from './ConnectChannelRequestPayload';
+export * from './ConversationPayload';
```
lucapette, line 5: If I understand how the code is organised now, we could export only Request/Response payloads here and we probably should?

lib/typescript/httpclient/model/Tag.ts
```diff
@@ -0,0 +1,7 @@
+export type ColorTag = 'tag-blue' | 'tag-red' | 'tag-green' | 'tag-purple';
```
lucapette, line 1: I'd call the type `TagColor` I think, it seems more fitting? (naming is hard though and I may be getting this wrong)

lib/typescript/httpclient/endpoints/updateTag.ts
```diff
@@ -0,0 +1,12 @@
+import {doFetchFromBackend} from '../api';
+import {Tag} from '../model';
+
+export function updateTag(tag: Tag) {
+  return doFetchFromBackend('tags.update', {
+    id: tag.id,
+    name: tag.name,
+    color: tag.color,
+  })
```
lucapette, line 9: (honest question) would `{...tag}` would here?

lucapette (commented):
Last pass!

lib/typescript/httpclient/endpoints/loginViaEmail.ts
AudreyKj: it does work!

lib/typescript/httpclient/endpoints/updateTag.ts
AudreyKj: yesss the tag object has exactly the same format so it works!

lucapette (approved):
:shipit:

lucapette (approved):

Issue #327, opened 2020-11-18
Introduce a release script that automates the process


pascal-airy:
The script has to be interactive so we can test things during the process.
We will consider using the git-flow tool which wraps some steps for us.
Need a way to push the release tag. "Git push origin —-tags" pushed all local tags

PR #586, opened 2021-01-06
[#327] Introduce a release script that automates the process

paulodiniz:
Fixes https://github.com/airyhq/airy/issues/327

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -1,4 +1,6 @@
 #!/bin/bash
+set -eo pipefail
+IFS=$'\n\t'
 
 start() {
     release_number=$1
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -4,18 +4,19 @@ IFS=$'\n\t'
 
 start() {
     release_number=$1
-    echo -e "Starting release ${release_number}"
+    echo -e "Starting release ${release_number}\n"
     create_issue
     create_release_branch
 }
 
 create_issue() {
-    command curl -s \
+    issue_number=$(command curl -s \
         -X POST \
         -H "Accept: application/json" \
         -H "Authorization: token ${GITHUB_TOKEN}" \
         https://api.github.com/repos/airyhq/airy/issues \
-        -d "{\"title\":\"Release ${release_number}\", \"labels\": [\"release\"]}" | jq '.number'
+        -d "{\"title\":\"Release ${release_number}\", \"labels\": [\"release\"]}" | jq '.number')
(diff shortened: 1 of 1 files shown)


scripts/release.sh
```diff
@@ -0,0 +1,72 @@
...
+    command git push origin release/${release_number}
+}
+
+finish() {
+    release_number=$1
+    echo -e "Finishing release ${release_number}"
+    increase_version
+    commit_version
+    merge_main
+    merge_develop
```
paulodiniz, line 32: When do we close the issue?

ljupcovangelski (commented):
I think we need couple of things more at this stage:
- check for the variables if they are there ($1, $2) and what are their values, so that we don't get any unexpected behavior
- short help/usage message, so that we see how to invoke and run the release

We can give it a proper test as soon as next Tuesday I guess. The others things we can add along the way.

scripts/release.sh
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
```
lucapette, line 1: we have this "unofficial bash script mode" in other scripts and we should add it here as well

scripts/release.sh
lucapette: it should close automatically based on the commit message?

lucapette (commented):
First quick pass! Looking great already :)

scripts/release.sh
```diff
@@ -0,0 +1,88 @@
...
+}
+
+merge_main() {
+    command git checkout main
+    command git pull origin main
+    command git merge --no-ff release/${release_number}
+    command git tag ${release_number}
+    command git push origin main
+    command git push origin ${release_number}
+    echo -e "Sucesfully merged into main branch\n"
```
lucapette, line 62:
```suggestion
    echo -e "Sucessfully merged into main branch\n"
```

lucapette, line 62:
```suggestion
    echo -e "Successfully merged into main branch\n"
```

scripts/release.sh
```diff
@@ -0,0 +1,88 @@
...
+}
+
+merge_develop() {
+    command git checkout develop
+    command git pull origin develop
+    command git merge --no-ff release/${release_number}
+    command git push origin develop
+    echo -e "Sucesfully merged into develop branch\n"
```
lucapette, line 70:
```suggestion
    echo -e "Successfully merged into develop branch\n"
```

pascal-airy (approved):

Issue #497, opened 2020-12-14
Media resolver app


chrismatix:
Requires #495 and #496 

Source media URLs can appear in two places: As metadata on the conversation (e.g. contact avatar) or as a payload of a message (e.g. an image message). Because those URLs tend to expire we need to store their content in the media storage provided by #495

Implementation notes:

## Stream app
Required env var: `storage.host`, which points to the host and directory the media should be uploaded to (e.g. `https://mybucket.s3.amazonaws.com/core-runtime/media/`)

**Metadata**
- Stream metadata
- Check pre-defined metadata URLs for external URLs: If they exist, upload their content to the storage service (see URL folder structure below)
- Produce the uploaded metadata KV to `{metadata key}.resolved`, i.e. `source.contact.avatar_url.resolved`

**Messages**
We add a new type of metadata for messages where the `subject` is `message:{message-id}`

- Stream from messages
- Use the new media content model (#496) to extract media URLs
- Use the same check for the metadata to see if we need to upload the URL
- Upload the URL (see URL folder structure below)
- Produce a message metadata record with metadata key = `data_{source url}` and the value  = `storage.host` + "{message source}/data_{sha1 of source url}"

**URL folder structure**
```
.
├── {conversation-id}/
│   └── {metadata-key}.resolved
└── data_{sha256 of data url}
```

**Rendering media data**
- The mapping library render method takes a message metadata map as a second parameter
- When a data url is encountered: **If** a key (`data.{source url}`) in the metadata is present, replace it with the metadata value

- [ ] media resolver app
- [x] content mapper render metadata

chrismatix:
We noticed that making there are two ways to go with replacing the URLs: Either we make the storage host an environment variable that gets replaced on the fly or we write the URLs to the metadata and don't have a config.

We have opted to go with the latter for the following reasons:
- While migrating data is easy with the host variable (you just switch it after migration), you could also migrate record by record using the metadata
- Mapping library does not depend on infrastructure configuration
- Strong startup check for the media resolver: If the storage host is not accessible, data URLs cannot be written

PR #594, opened 2021-01-07
[#497] Dynamically map source data urls in content mapper

chrismatix:
required for #497


lib/java/mapping/src/main/java/co/airy/mapping/ContentMapper.java
```diff
@@ -35,12 +46,54 @@ public ContentMapper(List<SourceMapper> sourceMappers, OutboundMapper outboundMa
...
         }
 
-        return sourceMapper.render(message.getContent());
+        final List<Content> contentList = sourceMapper.render(message.getContent());
+
+        if (storageHost == null) {
+            return contentList;
+        }
+
+        /*
```
chrismatix, line 55: Usually not a fan of long methods, but this loop cannot be moved without forwarding the metadata and message, which is not much cleaner.

backend/media/src/main/java/co/airy/core/media/Resolver.java
```diff
@@ -0,0 +1,46 @@
...
+
+    public Resolver(KafkaStreamsWrapper streams) {
+        this.streams = streams;
+    }
+
+    private void startStream() {
+        final StreamsBuilder builder = new StreamsBuilder();
+
+        streams.start(builder.build(), appId);
+    }
```
lucapette, line 28: We've been skipping this method in other streaming apps (and directly using the onApplicationEvent), I would probably do this year as well

backend/media/BUILD
```diff
@@ -0,0 +1,41 @@
+load("//tools/build:springboot.bzl", "springboot")
```
lucapette, line 1: `backend/media/resolver`? I understand why we didn't do yet (after all there's nothing else in the directory and probably nothing coming for a while). The proposed folder structure would create a certain symmetry. No strong opinion though, just though it was worth sharing

lib/java/mapping/src/main/java/co/airy/mapping/ContentMapper.java
lucapette: It's not so bad :)

lucapette (commented):
First pass! Looking pretty good!

backend/media/BUILD
chrismatix: It's a good point, but for now, I would keep it like so if only to keep it simple. I think it is harder to maintain that symmetry in core, e.g. look at the chat plugin which does not match the other sources. Or sources in general vs `api`. 

Overall I appreciate that we are a bit more flexible on core because it encourages us to think harder about directory structure.

lucapette (approved):
:shipit:

Issue #526, opened 2020-12-16
Global resource prefix


chrismatix:
## Userstory

As a user of the airy core platform
I want to be able to add a global resource prefix
so that I can run the Core platform along with other services

## Acceptance Criteria

- **GIVEN** A bootstrap configuration with a prefix exists
- **WHEN** the cluster is bootstrapped
- **THEN** all Kafka topics, the Postgres database, Kubernetes resources and redis installation all receive that prefix

- [x] Prefix Kafka topics
- [x] Prefix Postgres resources
- [ ] Prefix Kubernetes resources

PR #595, opened 2021-01-08
[#526] Introduce namespace var

pascal-airy:
This PR adds a new Helm variable that defaults to "default" but can be set in the airy.conf in case all resources should be deployed to a different namespace.


infrastructure/scripts/provision/core.sh
```diff
@@ -15,7 +15,12 @@ echo "Deploying the Airy Core Platform with the ${APP_IMAGE_TAG} image tag"
 cd /vagrant/scripts/
 wait-for-service-account
 
-helm install core ~/airy-core/helm-chart/ --set global.appImageTag=${APP_IMAGE_TAG} --version 0.5.0 --timeout 1000s > /dev/null 2>&1
+if [ -f "/vagrant/airy.conf" ]; then
+    cp /vagrant/airy.conf ~/airy-core/helm-chart/values.yaml
+    
+fi
+
+helm install core ~/airy-core/helm-chart/ -f ~/airy-core/helm-chart/values.yaml --version 0.5.0 --timeout 1000s
```
pascal-airy, line 23: I think it is better to pass all values instead of just the individual APP_IMAGE_TAG

lucapette (approved):

ljupcovangelski (approved):
Looks good.

PR #596, opened 2021-01-08
Use karapace.io schema registry

lucapette:


infrastructure/images/schema-registry/Makefile
```diff
@@ -0,0 +1,10 @@
+build:
+	docker build -t schema-registry . --build-arg SCHEMA_REGISTRY_VERSION=${SCHEMA_REGISTRY_VERSION}
```
lucapette, line 2: so the current build system assumes you have the corresponding egg file in the current directory. I did this for simplicity short-term but we should probably improve over time

ljupcovangelski (approved):
The image is much smaller, starts much faster and the helm chart looks much simple.
Awesome :tada:

PR #603, opened 2021-01-11
[#501] Resume conversation in chat plugin

### LINKED ISSUE #501: Chat plugin: Identity Capture
## Userstory
As an Airy developer displaying Airy Chat for all potential contacts, I want to continue an open conversation in the future so that I would like to capture their identification during the initial conversation

## Clients
* [x] API

## Acceptance Criteria
- *GIVEN* contact is identified with a unique code after initiation of first conversation
- *WHEN* the contact sends their first message to Airy
- *THEN* Airy responds with an automated message sharing a unique code as a way of differentiating between contacts

### LINKED ISSUE #607: Resume conversation: page reloads
### Summary
In order to get a better conversation experience, we should offer the ability to resume a previous conversation after page reloads.

### Requirements
On Airy Core, the ability to resume conversation should be extended to any instance where the page has been reloaded. The results of this will be:
- the ability to maintain a previous conversation
- the creation of less unwanted "new conversations" with the same contact

Storing the resume token inside localStorage
- Offer the option so that the plugin saves the current conversation in localStorage

ogbeche77:
Fixes #501 
Fixes #607

### DIFF
--- a/frontend/chat-plugin/src/components/chat/index.tsx
+++ b/frontend/chat-plugin/src/components/chat/index.tsx
@@ -46,7 +46,7 @@ const Chat = (props: Props) => {
       localStorage.setItem('resume_token', queryParams.get('resume_token'));
     }
 
-    return queryParams.get('resume_token') ? queryParams.get('resume_token') : localStorage.getItem('resume_token');
+    return queryParams.get('resume_token') || localStorage.getItem('resume_token');
   };
 
   useEffect(() => {
--- a/frontend/chat-plugin/src/components/chat/index.tsx
+++ b/frontend/chat-plugin/src/components/chat/index.tsx
@@ -42,6 +42,9 @@ const Chat = (props: Props) => {
 
   const getResumeToken = () => {
     const queryParams = new URLSearchParams(window.location.search);
+    if (queryParams.has('resume_token')) {
+      localStorage.setItem('resume_token', 'resume_token');
+    }
     return queryParams.get('resume_token') ? localStorage.getItem('resume_token') : null;
   };
 
--- a/frontend/chat-plugin/src/components/chat/index.tsx
+++ b/frontend/chat-plugin/src/components/chat/index.tsx
@@ -40,8 +40,17 @@ const Chat = (props: Props) => {
   const [isChatHidden, setIsChatHidden] = useState(true);
   const [messages, setMessages] = useState([welcomeMessage]);
 
+  const getResumeToken = () => {
+    const queryParams = new URLSearchParams(window.location.search);
+    if (queryParams.has('resume_token')) {
+      queryParams.get('resume_token');
+      return 'resume_token';
+    }
+    return null;
+  };
+
   useEffect(() => {
-    ws = new Websocket(props.channel_id, onReceive);
+    ws = new Websocket(props.channel_id, onReceive, getResumeToken());
     ws.start().catch(error => {
       console.error(error);
       setInstallError(error.message);
--- a/frontend/chat-plugin/src/components/chat/index.tsx
+++ b/frontend/chat-plugin/src/components/chat/index.tsx
@@ -43,9 +43,10 @@ const Chat = (props: Props) => {
   const getResumeToken = () => {
     const queryParams = new URLSearchParams(window.location.search);
     if (queryParams.has('resume_token')) {
-      localStorage.setItem('resume_token', 'resume_token');
+      localStorage.setItem('resume_token', queryParams.get('resume_token'));
     }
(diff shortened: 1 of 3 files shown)


bitboxer (approved):
LGTM

frontend/chat-plugin/src/components/websocket/index.ts
```diff
@@ -65,12 +67,20 @@ class Websocket {
   };
 
   async start() {
+    let authenticateChannelAndResumeToken: any = {
```
AitorAlgorta, line 70: can we have a better typing here?

frontend/chat-plugin/src/components/chat/index.tsx
```diff
@@ -40,8 +40,17 @@ const Chat = (props: Props) => {
   const [isChatHidden, setIsChatHidden] = useState(true);
   const [messages, setMessages] = useState([welcomeMessage]);
 
+  const getResumeToken = () => {
+    const queryParams = new URLSearchParams(window.location.search);
+    if (queryParams.has('resume_token')) {
+      queryParams.get('resume_token');
```
chrismatix, line 46: Shouldn't this `return queryParams.get('resume_token');`? 🤔

AitorAlgorta (approved):
AuthProps might be even better name, but LGTM

frontend/chat-plugin/src/components/chat/index.tsx
```diff
@@ -40,8 +40,16 @@ const Chat = (props: Props) => {
   const [isChatHidden, setIsChatHidden] = useState(true);
   const [messages, setMessages] = useState([welcomeMessage]);
 
+  const getResumeToken = () => {
+    const queryParams = new URLSearchParams(window.location.search);
+    if (queryParams.has('resume_token')) {
+      return queryParams.get('resume_token');
+    }
+    return null;
```
lucapette, line 48: `return queryParams.has('resume_token') ? queryParams.get('resume_token') : null` maybe the only very good use case for ternary operators :)

lucapette (commented):
Minor comments but looking good!

frontend/chat-plugin/src/components/chat/index.tsx
lucapette: also I just read the docs and `return queryParams.get('resume_token')` would return null if the key isn't there so it can be simplifed further (unless there's any other reason to check the key is there and I'm not aware of)

frontend/chat-plugin/src/components/chat/index.tsx
```diff
@@ -40,8 +40,13 @@ const Chat = (props: Props) => {
   const [isChatHidden, setIsChatHidden] = useState(true);
   const [messages, setMessages] = useState([welcomeMessage]);
 
+  const getResumeToken = () => {
+    const queryParams = new URLSearchParams(window.location.search);
+    return queryParams.has('resume_token') ? queryParams.get('resume_token') : null;
```
bitboxer, line 45: Hm, doesn't it automatically return null if the key does not exist?

bitboxer (commented):
I have one little comment

frontend/chat-plugin/src/components/chat/index.tsx
lucapette:
```suggestion
    return queryParams.get('resume_token');
```

should do it (as @bitboxer is also saying :))

lucapette (changes requested):
Looking better, I"d still incorporate the feedback left.

Also we're saying this solves #607 but it does not so maybe it should be a draft?

AitorAlgorta (approved):
LGTM the only thing I would do is moving the API calls to a different folder, maybe `/api` to have to code a little bit more clean

frontend/chat-plugin/src/components/chat/index.tsx
```diff
@@ -40,8 +40,17 @@ const Chat = (props: Props) => {
   const [isChatHidden, setIsChatHidden] = useState(true);
   const [messages, setMessages] = useState([welcomeMessage]);
 
+  const getResumeToken = () => {
+    const queryParams = new URLSearchParams(window.location.search);
+    if (queryParams.has('resume_token')) {
+      localStorage.setItem('resume_token', queryParams.get('resume_token'));
+    }
+
+    return queryParams.get('resume_token') ? queryParams.get('resume_token') : localStorage.getItem('resume_token');
```
AitorAlgorta, line 49: small improvement => `return queryParams.get('resume_token') || localStorage.getItem('resume_token');
`

AitorAlgorta (approved):
LGTM

lucapette (approved):
:shipit:

Issue #546, opened 2020-12-23
Image tag doesn't work from airy.conf`


ljupcovangelski:
Specifying 
```
global:
 appImageTag: beta
```
in `airy.conf` doesn't work and the apps are not deployed with the `beta` image. We need to read this from the conf file, so that provisioning works properly.

ljupcovangelski:
This issue was partially solved with https://github.com/airyhq/airy/pull/610. However, when we run the `conf` provisionner, the version defaults to the one with which `core` was deployed initially. So the issue stays open.

ljupcovangelski:
Was fixed with the `0.5.0` release. Currently the priority of the `appImageTag` is:
- `airy.conf` file
- `AIRY_VERSION` explicitly set during the bootstrap 
- the default `latest` value, specified in the helm chart

PR #606, opened 2021-01-11
[#546] Fix/546 revise image tags

pascal-airy:
This will remove the necessity to specify a AIRY_VERSION when running the bootstrap but the env var will still override the setting derived from the branch.
There will only be image tags develop for the current development version of an app or X.Y.Z for a released version of an app or service


chrismatix (approved):
Some of these changes I don't understand well enough to comment on, the rest looks good

infrastructure/scripts/provision/core.sh
```diff
@@ -6,14 +6,29 @@ SCRIPT_PATH=$(cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P)
...
+    branch_name="$(git symbolic-ref HEAD 2>/dev/null)" ||
+    branch_name="(unnamed branch)"     # detached HEAD
+
+    branch_name=${branch_name##refs/heads/}
+    case "$branch_name" in
+        main|release* )
+            AIRY_VERSION=(`cat ../../VERSION`)
+            ;;
+        * )
+            AIRY_VERSION=develop
```
ljupcovangelski, line 24: Should this be `beta` or `develop` ?

ljupcovangelski: It's ok, I see we will also push the images with the `develop` tag, instead of `beta`
.

infrastructure/scripts/conf.sh
```diff
@@ -5,30 +5,35 @@ IFS=$'\n\t'
...
+source ${INFRASTRUCTURE_PATH}/scripts/lib/k8s.sh
+
 
 kubectl delete pod startup-helper --force 2>/dev/null || true
 kubectl run startup-helper --image busybox --command -- /bin/sh -c "tail -f /dev/null"
 
-helm upgrade core ${INFRASTRUCTURE_PATH}/helm-chart/ --values ${INFRASTRUCTURE_PATH}/airy.conf --set global.appImageTag=${AIRY_VERSION} --timeout 1000s > /dev/null 2>&1
+helm upgrade core ${INFRASTRUCTURE_PATH}/helm-chart/ --values ${INFRASTRUCTURE_PATH}/airy.yaml --set global.appImageTag=${AIRY_VERSION} --timeout 1000s > /dev/null 2>&1
 
 kubectl scale deployment schema-registry --replicas=1
```
ljupcovangelski, line 39: This way, we don't consider the version with which the core was deployed, as well as if we are on `main` and we want to `override` the version.
Or if we are on another branch and want to test a specific version (example 0.5.0).

pascal-airy: you can always export AIRY_VERSION=beta before starting the bootstrap

ljupcovangelski (approved):
Looks good :+1:

PR #617, opened 2021-01-12
Feature/598 read unread state of conversations

### LINKED ISSUE #598: Read/Unread state of conversations

AudreyKj:
resolves #598 

changes: 
- [readConversations](https://develop.docs.airy.co/api/http#conversations) endpoint in http lib 
- readConversations action + reducer
- added check to api config because the readConversations endpoint returns a 202 with an empty response


lib/typescript/httpclient/api/airyConfig.ts
```diff
@@ -34,10 +34,16 @@ export const doFetchFromBackend = async (url: string, body?: Object): Promise<an
 
 async function parseBody(response: Response): Promise<any> {
   if (response.ok) {
-    return response.json();
+    try {
+      let jsonResponse = await response.json();
```
chrismatix, line 38: Could you explain a bit the reasoning behind this change? Now it seems like clients of this function (so everywhere we call `doFetchFromBackend`) get `{}` in the unlikely case that the API returns non-json data for a successful call.

I get the idea of making it more robust, but I would suggest that we can just remove the `return` so that the error code further down in the function throws a proper error. What do you think?

lucapette, line 38: const should do it?

lucapette (approved):
Minor comment. :shipit:

lib/typescript/httpclient/api/airyConfig.ts
bitboxer, line 38: i think this could just be a `return await response.json()`

bitboxer (approved):
A small nitpick, looking good.

lib/typescript/httpclient/endpoints/readConversations.ts
```diff
@@ -0,0 +1,7 @@
+import {doFetchFromBackend} from '../api';
+
+export function readConversations(conversationId: string) {
+  return doFetchFromBackend('conversations.read', {conversation_id: conversationId})
+    .then(() => Promise.resolve(true))
+    .catch((error: Error) => Promise.reject(error));
```
chrismatix, line 6: The `catch` line is redundant because it catches and immediately throws the error again

lib/typescript/httpclient/api/airyConfig.ts
AudreyKj: will just do return await response.json()

AudreyKj: The idea behind this was that for successful calls that return non-json data, we wouldn't want to return an error and we wouldn't be able to consume the response again, so this why I returned {} (more like a temporary solution). 

I see your point and the problem of just returning {} in all cases, and I like the idea of removing the return in catch to let the function throw a proper error, as it would give us more flexibility. I guess we would automatically get an error for trying to call .text() on the response.

lib/typescript/httpclient/api/airyConfig.ts
```diff
@@ -34,10 +34,16 @@ export const doFetchFromBackend = async (url: string, body?: Object): Promise<an
 
 async function parseBody(response: Response): Promise<any> {
   if (response.ok) {
-    return response.json();
+    try {
+      return await response.json();
+    } catch {
+      {
+      }
```
chrismatix, line 41: The empty `{}` block can also go 🙏

lib/typescript/httpclient/endpoints/readConversations.ts
```diff
@@ -0,0 +1,7 @@
+import {doFetchFromBackend} from '../api';
+
+export function readConversations(conversationId: string) {
+  return doFetchFromBackend('conversations.read', {conversation_id: conversationId})
+    .then(() => Promise.resolve(true))
+    .catch((error: Error) => error);
```
chrismatix, line 6: Same as with the other catch block here (It's enough to just remove L6)

chrismatix (commented):
Second pass

bitboxer (approved):
LGTM!

bitboxer:
Sadly there is a merge conflict 😢 . Sorry that you need to change something again.

AitorAlgorta (approved):

PR #647, opened 2021-01-13
[#621] Controller starts/stops apps based on config map changes

### LINKED ISSUE #621: Controller starts/stops apps based on configMap
When a configMap is created, the controller should scale all the deployments which have replicas `0` and are using the configMap (both for ENV variables and volumes).

Similarly, when a configMap is deleted, the controller should scale down the deployment which is using the configMap.

These processes shouldn't disrupt the reload of the deployments, when a configmap is changed.

chrismatix:
fixes #621 

- [x] Ensure all config maps are present before deployment is scaled up
- [x] Applying helm chart when controller is running does not work


infrastructure/controller/pkg/configmap-controller/create.go
```diff
@@ -0,0 +1,40 @@
+package configmapController
```
lucapette, line 1: I think I missed this when we introduced the package. That's not very golang in terms of conventions. 

I'd suggest something like `cmcontroller` or `cmctrl` (the guideline is short, abbreviated names https://blog.golang.org/package-names). I have no strong opinion about the name itself though

infrastructure/controller/pkg/configmap-controller/create.go
```diff
@@ -0,0 +1,40 @@
...
+		return errGetDeployments
+	}
+
+	for _, affectedDeployment := range affectedDeployments {
+		klog.Infof("Scheduling start for deployment: %s", affectedDeployment)
+		if err := handler.ScaleDeployment(handler.ScaleCommand{
+			ClientSet: clientSet,
+			Namespace: "default",
+			DeploymentName: affectedDeployment,
+			DesiredReplicas: 1,
```
lucapette, line 32:
```suggestion
			DesiredReplicas: 1, //TODO extract from annotation
```

so that we don't forget about how we'd want to solve this :)

infrastructure/controller/pkg/configmap-controller/create.go
```diff
@@ -0,0 +1,40 @@
...
+	if errGetDeployments != nil {
+		klog.Errorf("Error retrieving affected deployments %v", errGetDeployments)
+		return errGetDeployments
+	}
+
+	for _, affectedDeployment := range affectedDeployments {
+		klog.Infof("Scheduling start for deployment: %s", affectedDeployment)
+		if err := handler.ScaleDeployment(handler.ScaleCommand{
+			ClientSet: clientSet,
+			Namespace: "default",
```
lucapette, line 30: I'm not sure what's the best solution here... but as I understand it this wouldn't work if one would namespace their core installation?

infrastructure/controller/pkg/configmap-controller/create.go
chrismatix: Good point!

infrastructure/controller/pkg/configmap-controller/create.go
chrismatix: True, I will make it so that it can read this from the configuration. Or do you think it should pick up the default from the config file?

lucapette: I like the ENV idea as discussed offline 👍

infrastructure/helm-chart/charts/apps/charts/frontend-chat-plugin/templates/deployment.yaml
```diff
@@ -25,6 +25,9 @@ spec:
       - name: app
         image: "{{ .Values.global.containerRegistry}}/{{ .Values.image }}:{{ .Values.global.appImageTag }}"
         imagePullPolicy: Always
+        envFrom:
+        - configMapRef:
+            name: frontend-config
```
chrismatix, line 30: Dummy config, so that frontends can be started by the controller

pascal-airy (approved):
Tested starting stopping and updating from config maps 👍

lucapette (approved):

PR #652, opened 2021-01-14
[#651] Use Google Cloud Storage for the bazel remote cache

### LINKED ISSUE #651: CI cache misses
The CI is no longer able to create the `bazel` cache key and it is currently missing. We need a more resilient setup that makes use of [restore keys](https://docs.github.com/en/free-pro-team@latest/actions/guides/caching-dependencies-to-speed-up-workflows#using-the-cache-action)

image

chrismatix:
fixes #651


lucapette (approved):
👏 👏 👏

PR #658, opened 2021-01-14
[#546] Fix/546 revise image tags

pascal-airy:
This will remove the necessity to specify a AIRY_VERSION when running the bootstrap but the env var will still override the setting derived from the branch.
There will only be image tags develop for the current development version of an app or X.Y.Z for a released version of an app or service


ljupcovangelski (dismissed):
Looks good :+1:

lucapette (dismissed):
:shipit:

lucapette (dismissed):

Issue #624, opened 2021-01-13
Enable users to use the API with their own instance


AudreyKj:
**current state:** 
the api config only uses airy url for the api url (http://api.airy) 

**to do:**
tweak api config initialisation to enable users to call the endpoint with their own instance (use airy as default)

**goal:**
enable users of the library to call the endpoints with their own instance 

---> overlapping with / related to #626

RESOLVED: pull request https://github.com/airyhq/airy/pull/664

PR #664, opened 2021-01-15
refactored lib with class that can be instantiated

### LINKED ISSUE #626: refactor library's module with a class
**Current state:** 
the library's module is a iife function, no instantiation possible 

**To do:**
refactor library's module with a class that can be instantiated with the user's own instance

--> overlapping with + will also resolve #624

AudreyKj:
resolves #626 and #624

changes: 
- refactored library with a class that can be instantiated with the auth token + api url. I made both the token and api url optional because not all endpoints require a token and the default for the api url is airy.co.
- moved the endpoints in the class 
- moved cookies to demo 
- moved AiryConfig object to demo 
- instantiated the library's class, using the auth token in cookies, in demo 
- fixed the PaginatedPayload interface (the backend's response gives a response_metadata field and not responseMetadata)
- updated the readme (def not the final docs, but it's just to have it in line with the code)


AitorAlgorta (approved):
LGTM there are a couple thing we can improve but we will do it in future PRs

PR #668, opened 2021-01-18
Introduce /client.config

lucapette:


paulodiniz (dismissed):

chrismatix (approved):
Good stuff 👍

backend/sources/facebook/connector/src/main/java/co/airy/core/sources/facebook/ChannelsController.java
```diff
@@ -104,4 +106,10 @@ public ChannelsController(Api api, Stores stores) {
             return ResponseEntity.status(HttpStatus.SERVICE_UNAVAILABLE).build();
         }
     }
+
+    @Override
+    public Health health() {
```
chrismatix, line 111: This indicator should already be on the `Stores` class

chrismatix (changes requested):

chrismatix (dismissed):

backend/sources/facebook/connector/src/main/java/co/airy/core/sources/facebook/ChannelsController.java
lucapette: will move all of them there! It's a better place :)

chrismatix (approved):

Issue #620, opened 2021-01-13
Reload configuration based on config file


pascal-airy:
`airy config reload`

This command updates config maps based on `airy.yml`

ljupcovangelski:
This is done. Works with
```
airy config apply
```

PR #673, opened 2021-01-18
[#620] Download kubeconf file

ljupcovangelski:
- Move the cli configuration to the directory `~/.airy`
- Download kubeconf file to the directory

### DIFF
--- a/infrastructure/cli/cmd/root.go
+++ b/infrastructure/cli/cmd/root.go
@@ -14,7 +14,8 @@ import (
 	"github.com/spf13/viper"
 )
 
-const configFileName = ".airycli.yaml"
+const configFileName = "cli.yaml"
+const configDirName = ".airy"
 
 var configFile string
 var Version string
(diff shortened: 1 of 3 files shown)


infrastructure/cli/cmd/root.go
```diff
@@ -55,7 +57,16 @@ var initCmd = &cobra.Command{
...
 
-		err = viper.WriteConfigAs(path.Join(home, configFileName))
+		configDirPath := path.Join(home, configDirName)
+
+		if _, err := os.Stat(configDirPath); os.IsNotExist(err) {
+			errDir := os.MkdirAll(configDirPath, 0700)
+			if errDir != nil {
+				log.Fatal(err)
+			}
+		}
```
ljupcovangelski, line 67: Check if the directory `~/.airy` exists, if not - create it.

scripts/bootstrap.sh
```diff
@@ -140,3 +140,7 @@ if [ -z ${AIRY_VERSION+x} ]; then
 fi
 
 AIRY_VERSION=${AIRY_VERSION} vagrant up
+
+mkdir -p ~/.airy
+cd $infra_path
+vagrant ssh -c "cat /etc/rancher/k3s/k3s.yaml" 2>/dev/null | sed "s/127.0.0.1/192.168.50.5/g" > ~/.airy/kube.conf
```
ljupcovangelski, line 146: We can also do this with a provisionner, this is more lightweight I think, because otherwise we need to install another plugin, to be able to run commands on the host (the shell provisionner runs commands only on the guest vm).

lucapette (dismissed):

lucapette (dismissed):

paulodiniz (approved):

Issue #644, opened 2021-01-13
Facebook connector send out template content


chrismatix:
Depends on #641

When clients send a message with content-type `SourceTemplate` (added in issue mentioned above), take the stringified content and pass it as as a JSON directly in the `message` [field](https://developers.facebook.com/docs/messenger-platform/send-messages/templates) to the Facebook API.

This may take a bit of coordination with the frontend depending on how they will construct template payload

Issue #643, opened 2021-01-13
Render images and a template


chrismatix:
Depends on #642 

To prove that we can render our typed outbound model and the source templates, render 

- [ ] images
- [ ] a [facebook template](https://developers.facebook.com/docs/messenger-platform/send-messages/templates) of your choice

For the latter we currently only care about the outbound structure of the rendered template

chrismatix:
Requires docs produced in https://github.com/airyhq/airy/issues/640 and https://github.com/airyhq/airy/issues/644

PR #680, opened 2021-01-19
[#640] Enrich send message API with source template messages

### LINKED ISSUE #640: Send Message API include content type
Currently `/messages.send` only takes in `text` messages.

To support the existing outbound content model and source templates, we need to update the send message API so that the request payload captures the right type of the content. I.e.

```json
{
  "message": {
    "url": "url to an image",
    "type": "image"
  }
}
```

This outbound content should be deserialized and then serialized using `objectMapper.readValue(content, Content.class)` (to check that the format is correct) to make use of [Jackson polymorphic serialization](https://github.com/FasterXML/jackson-docs/wiki/JacksonPolymorphicDeserialization) and the existing content model in `lib/java/mapping/src/main/java/co/airy/mapping/model`.

- [ ] update docs
- [ ] update send message api

chrismatix:
fixes #640

### DIFF
--- a/docs/docs/api/http.md
+++ b/docs/docs/api/http.md
@@ -352,13 +352,45 @@ This is a [paginated](#pagination) endpoint. Messages are sorted from oldest to 
 
 Sends a message to a conversation and returns a payload.
 
-**Sample request**
+**Sending a text message**
 
 ```json5
 {
   "conversation_id": "a688d36c-a85e-44af-bc02-4248c2c97622",
   "message": {
-    "text": "{String}"
+    "text": "Hello World",
+    "type": "text"
+  }
+}
+```
+
+**Sending an attachment message**
+
+```json5
+{
+  "conversation_id": "a688d36c-a85e-44af-bc02-4248c2c97622",
+  "message": {
+    "url": "http://example.org/myfile",
+    "type": "image|video|audio|file"
+  }
+}
+```
+
+**Sending source templates**
+
+Some sources support sending templates, which can be used to display rich content such as buttons or cards. You
+can send source templates by setting the type to `source.template`. Please refer to the source documentation
+to see the expected values for the `payload` field.
+
+```json5
+{
+  "conversation_id": "a688d36c-a85e-44af-bc02-4248c2c97622",
+  "message": {
+    "payload": {
+      "template_type": "buttons",
+      "buttons": ["Welcome to our shop"]
+    },
+    "type": "source.template"
   }
 }
 ```
@@ -372,12 +404,9 @@ Sends a message to a conversation and returns a payload.
     {
       "text": "{String}",
       "type": "text"
-      // Determines the schema of the content
     }
   ],
-  // typed source message model
-  "state": "{String}",
-  // delivery state of message, one of PENDING, FAILED, DELIVERED
+  "state": "pending|failed|delivered",
   "sender_type": "{string/enum}",
   // See glossary
   "sent_at": "{string}"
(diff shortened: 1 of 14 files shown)


docs/docs/api/http.md
```diff
@@ -352,13 +352,42 @@ This is a [paginated](#pagination) endpoint. Messages are sorted from oldest to
...
+can send source templates by setting the type to `source.template`. Please refer to the source documentation
+to see the expected values for the `payload` field.
 
 ```json5
 {
   "conversation_id": "a688d36c-a85e-44af-bc02-4248c2c97622",
   "message": {
-    "text": "{String}"
+    "payload": "{\"template_type\":\"buttons\",\"buttons\":[\"Welcome to our shop\"]}",
+    "type": "source.template"
```
chrismatix, line 390: @airyhq/codeowners-frontend The idea is that the backend does not maintain the data model of the sources, because it changes frequently and requires a lot of code that would in part be duplicated by the rendering on the frontend.

**Example**
To make this more concrete here is how this would work for Facebook (TODO #644):
```json
{
  "conversation_id": "a688d36c-a85e-44af-bc02-4248c2c97622",
  "message": {
    "type": "source.template",
    "payload": {
      "template_type":"generic",
      "elements":[
        {
          "title":"<TITLE_TEXT>",
          "image_url":"<IMAGE_URL_TO_DISPLAY>",
          "subtitle":"<SUBTITLE_TEXT>",
          "default_action": {
            "type": "web_url",
            "url": "<DEFAULT_URL_TO_OPEN>",
            "webview_height_ratio": "<COMPACT | TALL | FULL>"
          },
          "buttons": []
        }
      ]
    }
  }
}
```
Here the `payload` field is the same as the `payload` [you provide](https://developers.facebook.com/docs/messenger-platform/send-messages/template/generic/) when sending a template to Facebook directly.

As you can already imagine, the exact use of the `payload` field is dependent on each source and will be documented in the source docs. The implementation and documentation of the above payload will happen in #644

**Getting templates from the API**
The backend will store this as a stringified version and return it on messaging apis (`messsage.list`, websocket etc.) like so:
```
{
  ...
  "content": [
    {
      "type": "source.template",
      "payload": "{\"template_type\":\"generic\",\"elements\":[{\"title\":\"<TITLE_TEXT>\",\"image_url\":\"<IMAGE_URL_TO_DISPLAY>\",\"subtitle\":\"<SUBTITLE_TEXT>\",\"default_action\":{\"type\":\"web_url\",\"url\":\"<DEFAULT_URL_TO_OPEN>\",\"webview_height_ratio\":\"<COMPACT | TALL | FULL>\"},\"buttons\":[]}]}"
    }
  ]
}
```

Rendering one of those templates will be implemented in #643

**Google example**

```json
{
  "conversation_id": "...",
  "message": {
    "type": "source.template",
    "payload": {
      "suggestions": [
        {
          "reply": {
            "text": "Main Menu",
            "postbackData": "main-menu"
          }
        },
        {
          "action": {
            "text": "Some Link",
            "postbackData": "tapped-link1",
            "openUrlAction": {
              "url": "https://www.google.com"
            }
          }
        }
      ]
    }
  }
}
```

AitorAlgorta: My question is: If the structure of the sources changes a lot, wouldn't this also fail when they change it? We will have to change the code every time they change something right?

chrismatix: @AitorAlgorta that's true and the reason why we would like to only handle this code on the frontend, where it needs to be rendered.

chrismatix: By the way, this is what the `SourceTemplate` typescript interface would look like:
```

export interface SourceTemplate extends Content {
    type: "source.template";
    payload: any;
}
```

bitboxer: The problem for me: if we don't clean it up in the backend ,we need to change all frontends for every change. Is it possible to have an api where the changes are abstracted fully in the backend? That way we can react to changes quicker and don't have to invest in coding 3 solutions and deploying all of them.

lucapette (approved):

lucapette (approved):

PR #683, opened 2021-01-19
[#644] Facebook connector sends out templates

### LINKED ISSUE #644: Facebook connector send out template content

chrismatix:
fixes #644

### DIFF
--- a/backend/sources/facebook/connector/src/main/java/co/airy/core/sources/facebook/api/Mapper.java
+++ b/backend/sources/facebook/connector/src/main/java/co/airy/core/sources/facebook/api/Mapper.java
@@ -3,18 +3,23 @@ package co.airy.core.sources.facebook.api;
 import co.airy.avro.communication.Message;
 import co.airy.core.sources.facebook.api.model.SendMessagePayload;
 import co.airy.core.sources.facebook.dto.SendMessageRequest;
+import co.airy.mapping.ContentMapper;
+import co.airy.mapping.model.Content;
+import co.airy.mapping.model.SourceTemplate;
+import co.airy.mapping.model.Text;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.springframework.stereotype.Service;
 
+import java.util.List;
+
 
 @Service
 public class Mapper {
+    private final ContentMapper mapper;
 
-    private final ObjectMapper objectMapper;
-
-    Mapper(ObjectMapper objectMapper) {
-        this.objectMapper = objectMapper;
+    Mapper(ContentMapper mapper) {
+        this.mapper = mapper;
     }
 
     public SendMessagePayload fromSendMessageRequest(SendMessageRequest sendMessageRequest) throws Exception {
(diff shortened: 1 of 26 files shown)


backend/sources/facebook/connector/src/main/java/co/airy/core/sources/facebook/api/model/SendMessagePayload.java
```diff
@@ -15,36 +16,33 @@
 @AllArgsConstructor
 @JsonInclude(JsonInclude.Include.NON_EMPTY)
 public class SendMessagePayload {
-
-    @JsonProperty("messaging_type")
```
chrismatix, line 19: Lots of platform copy pasta, which was not needed

docs/docs/sources/facebook.md
```diff
@@ -170,3 +170,43 @@ the nature of the request, the response time may vary.
   ]
 }
 ```
+
+## Send a template message
```
chrismatix, line 174: Frontend relevant changes here ✨

frontend/chat-plugin/src/components/api/index.tsx
```diff
@@ -9,10 +11,12 @@ declare const window: {
 const API_HOST = window.airy ? window.airy.h : 'chatplugin.airy';
 const TLS_PREFIX = window.airy ? (window.airy.no_tls === true ? '' : 's') : '';
 
-export const sendMessage = (message: string, token: string) => {
+export const sendMessage = (message: Text, token: string) => {
```
chrismatix, line 14: And here 🌈

backend/sources/facebook/connector/src/main/java/co/airy/core/sources/facebook/api/Mapper.java
```diff
@@ -3,28 +3,43 @@
...
+                .orElseThrow(() -> new Exception(""));
+
+        if (content instanceof Text) {
+            messagePayload.setText(((Text) content).getText());
+        } else if (content instanceof SourceTemplate) {
+            messagePayload.setAttachment(SendMessagePayload.AttachmentPayload.builder()
+                    .type("template")
+                    .payload(((SourceTemplate) content).getPayload())
+                    .build());
+        }
```
paulodiniz, line 42: We can use a `Map<String,Function>` here
It's a pattern we used on the source mappers and I think they could be useful here

paulodiniz (changes requested):

backend/sources/facebook/connector/src/main/java/co/airy/core/sources/facebook/api/Mapper.java
chrismatix: How would that work in this case; i.e. what would be the key?

paulodiniz: Hm, good point

Do we have a `type` that we can use? If not maybe a `String` key that is the Class name
So a map could be

`"Text" -> renderingTextFunction`

`renderingTextFunction` is a Function that takes a MessagePayload and returns a MessagePayload

```
    private static final Function<MessagePayload, MessagePayload> renderingTextFunction = (messagePayload) -> <Return new modified message payload>;
```

What do you think?

chrismatix: Unfortunately, we have neither, because we have to check against the instance. The string type key only exists after serialization. In any case for an _n=2_ I think the code would end up looking more complex than it is now.

paulodiniz (approved):

PR #686, opened 2021-01-20
Introduce status command

lucapette:
Notable changes:

- `airy auth` -> `airy api login` & `airy api signup`
- the `airy api login` command stores the JWT token in the config so that subsequent calls work. The error handling and the error messages are subpar/close to non existent. Best approach there is creating a new milestone for this.
- added `airy status`. Doesn't do much yet but will scale nicely with `/client.config`
- make `/client.config` refresh state information in the background so the endpoint is fast


pascal-airy (approved):
It's beautiful

PR #694, opened 2021-01-20
Reorganize integration tests 

lucapette:


infrastructure/cli/integration/api_login_test.go
```diff
@@ -0,0 +1,47 @@
+package integration
+
+import (
+	"os/exec"
+	"testing"
+
+	"reflect"
+)
+
+func TestApiLogin(t *testing.T) {
```
lucapette, line 10: the idea here is to have one file per command (so I split the existing tests).

There's still some code that can be extracted in functions (especially the way we run the binaries... but I prefer to do that in a follow up pr)

infrastructure/cli/integration/golden/cli.yaml
```diff
@@ -1,2 +1,2 @@
-apihost: http://localhost:3001
+apihost: http://localhost:64789
```
lucapette, line 1: not really a relevant change... the way it works now this is bound to happen. I think the best shot for us here is to use the init command and generate a file for the current test (but we need to improve init so that it can take a dir/file argument)

infrastructure/cli/integration/mockserver.go
```diff
@@ -0,0 +1,54 @@
...
+	"goji.io/pat"
+)
+
+type MockServer struct {
+	l    net.Listener
+	Host string
+	mux  *goji.Mux
+}
+
+func NewMockServer(t *testing.T) *MockServer {
```
lucapette, line 22: somewhat "common" testing pattern in go. You pass down `t *testing.T)` so you can do things like `t.Log` or `t.Fail` from your testing helpers

infrastructure/cli/integration/mockserver.go
```diff
@@ -0,0 +1,54 @@
...
+		mux:  mux,
+		Host: fmt.Sprintf("http://localhost:%d", listener.Addr().(*net.TCPAddr).Port),
+	}
+}
+
+func (ms *MockServer) Serve() {
+	http.Serve(ms.l, ms.mux)
+}
+
+func mockEndpoint(endpoint string) func(w http.ResponseWriter, r *http.Request) {
```
lucapette, line 43: another somewhat common pattern. A function that returns a function given a specific param... it dries up the code pretty nicely

pascal-airy (approved):
I super approve

PR #702, opened 2021-01-21
Remove suppression

chrismatix:
We investigated a crash on a core instance with lots of production data. Suppression is not needed so far


lucapette (approved):
:shipit:

PR #712, opened 2021-01-22
[#698] Provide endpoint configuration to the frontend

### LINKED ISSUE #698: Inject endpoint configuration
**Current state**
Variables like the `API_HOST` are hardcoded. However in practice, the frontend singe page applications can be on any host, so they cannot infer where the API hosts are supposed to be.

**Desired state**
The `API_HOST` should be available to the frontend on the environment.

**Proposed solution**
Use the Nginx [lua resty](https://github.com/openresty/lua-nginx-module) module to template Kubernetes config variables into the `index.html`

- Install lua resty in build processs
- By default envs are not passed in nginx, so this needs to be [enabled](http://nginx.org/en/docs/ngx_core_module.html#env)
- Add host config to deployments
- Document and implement `window.airy` host configs for frontend

chrismatix:
fixes #698


AitorAlgorta (approved):

ogbeche77 (approved):

infrastructure/helm-chart/charts/apps/charts/sources-chatplugin/templates/deployment.yaml
```diff
@@ -35,7 +35,7 @@ spec:
         - name: JWT_SECRET
           valueFrom:
             configMapKeyRef:
-              name: api-config
+              name: sources-chatplugin
```
lucapette, line 38: external-content duckduckgo com

lucapette (approved):

Issue #403, opened 2020-11-27
Filter conversations


AitorAlgorta:
Create the UI to filter conversations

chrismatix:
This library can be used to construct lucene queries: https://www.npmjs.com/package/lucene-query-string-builder

This file contains the fields that a conversation can be filtered on: https://github.com/airyhq/airy/blob/2b0dda91b411913457cfc3b3b68bc133f4903bb5/backend/api/communication/src/main/java/co/airy/core/api/communication/dto/ConversationIndex.java

PR #744, opened 2021-01-26
[#403] Filter for conversations

AitorAlgorta:
#403


frontend/demo/src/actions/conversationsFilter/index.ts
```diff
@@ -0,0 +1,107 @@
...
+    if (filterQuery === '') {
+      filterQuery = 'source:(' + filter.bySources.join(' ') + ')';
+    } else {
+      filterQuery += ' AND source:(' + filter.bySources.join(' ') + ')';
+    }
+  }
+
+  console.log(filterQuery);
+
+  return filterQuery === '' ? null : filterQuery;
```
lucapette, line 106: if we introduce a `let filters = []`  we can do `filters.join("AND ")` to convert it back into a string right at the end. It should simplify the code quite a bit as we'd be able to save half of the ifs

Also console.log 💣

frontend/demo/src/actions/conversationsFilter/index.ts
```diff
@@ -0,0 +1,88 @@
...
+  if (filter.unreadOnly) {
+    filterQuery.push('unread_count:[1 TO *]');
+  } else if (filter.readOnly) {
+    filterQuery.push('unread_count:0');
+  }
+  if (filter.displayName) {
+    filterQuery.push('display_name:*' + filter.displayName + '*');
+  }
+  if (filter.byTags && filter.byTags.length > 0) {
+    filterQuery.push('tag_id:(' + filter.byTags.join(' ') + ')');
```
chrismatix, line 79: We will have to check if this works because we noted that UUIDs need to be escaped for the analyzer. On the backend tests we escape them like so: `.replace("-", "\\-")`

frontend/chat-plugin/src/airyRenderProps/AiryInputBar/index.tsx
```diff
@@ -13,7 +13,7 @@ const AiryInputBar = (props: Props) => {
     setMessageString('');
   }, []);
 
-  const resizeTextarea = e => {
+  const resizeTextarea = (e: any) => {
```
chrismatix, line 16: Is `any` not the default? 🤔

frontend/demo/src/actions/conversationsFilter/index.ts
```diff
@@ -0,0 +1,88 @@
...
+    cursor,
+    filters: filterToLuceneSyntax(filter),
+  })
+    .then((response: {data: Conversation[]; metadata: ResponseMetadataPayload}) => {
+      dispatch(mergeFilteredConversationsAction(response.data, filter, response.metadata));
+      return Promise.resolve(true);
+    })
+    .catch((error: Error) => {
+      return Promise.reject(error);
+    });
```
chrismatix, line 65:
```
.catch((error: Error) => {
      return Promise.reject(error);
    });
```
Doesn't really do anything so we can delete it

frontend/demo/src/components/SimpleIconChannel/index.tsx
```diff
@@ -0,0 +1,52 @@
+import * as React from 'react';
```
chrismatix, line 1:
```suggestion
import React from 'react';
```

frontend/demo/src/components/SimpleIconChannel/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+import {Channel} from 'httpclient';
+import {ReactComponent as GoogleIcon} from '../../assets/images/icons/google_avatar.svg';
+import {ReactComponent as WhatsappIcon} from '../../assets/images/icons/whatsapp_avatar.svg';
+import {ReactComponent as SmsIcon} from '../../assets/images/icons/sms_avatar.svg';
+import {ReactComponent as FacebookIcon} from '../../assets/images/icons/messenger_avatar.svg';
+import {useState} from 'react';
+
+import styles from './index.module.scss';
+
+const FB_GRAPH_API_URL = 'AiryConfig.FB_GRAPH_API_URL';
```
chrismatix, line 11: I think we should get rid of the Facebook code in this file, what do you think? :)

frontend/demo/src/pages/Inbox/ConversationList/index.tsx
```diff
@@ -104,11 +119,37 @@ const ConversationList = (props: ConversationListProps) => {
...
+    /*
+     This makes sure the div is resized and fit to the height of the
+     final size of the conversation list. Otherwise it would have two
+     scrollbars. One for the content of the conversation list and one for
+     the outer div that contains the conversation list. Sadly on some
+     render engine 20ms is not enough, but 200 works fine. BUT we want
+     to remove the scrollbar as soon as possible to remove flicker. Because
+     of that we calculate it twice, the 2nd one just to be sure it is really
+     gone even on slower machines.
+    */
```
chrismatix, line 140: I had hoped to never see this code again 😄 What do you think about a follow-up ticket just to refactor the conversation list code?

frontend/demo/src/reducers/data/conversations/index.ts
```diff
@@ -185,6 +194,63 @@ function allReducer(state: AllConversationsState = initialState, action: Action)
...
+function filteredReducer(
+  state: FilteredState = {
+    items: {},
+    metadata: {previousCursor: null, nextCursor: null, total: 0},
+    currentFilter: {},
+  },
+  action: FilterAction | Action
+): FilteredState {
+  switch (action.type) {
+    // case getType(filterActions.setFilteredConversationsAction):
```
chrismatix, line 206: Commented out code

infrastructure/airy.conf
```diff
@@ -0,0 +1,23 @@
+global:
```
chrismatix, line 1: This should not be checked in

chrismatix (commented):
Left some comments. Good work 🎉

frontend/chat-plugin/src/airyRenderProps/AiryInputBar/index.tsx
bitboxer: Also since this is a InputChange, we could add the correct type, or do I miss something?

frontend/demo/src/components/Dialog/index.tsx
```diff
@@ -1,6 +1,6 @@
 import React, {Fragment, useEffect, useCallback} from 'react';
 
-import styles from './Dialog.module.scss';
+import styles from './index.module.scss';
```
bitboxer, line 3: 👏

bitboxer (commented):
Added a few comments

frontend/demo/src/actions/conversationsFilter/index.ts
lucapette: this doesn't work but for a different reason: we have to `join(' AND ')`

AitorAlgorta: this was fixed :)

frontend/chat-plugin/src/airyRenderProps/AiryInputBar/index.tsx
AitorAlgorta: Actually I was trying to type it right, but the event types in preact are a bit weird and it wasn't working... I will try again @bitboxer @chrismatix

frontend/demo/src/components/SimpleIconChannel/index.tsx
AitorAlgorta: Yeah I agree

frontend/demo/src/components/DialogCustomizable/index.tsx
```diff
@@ -0,0 +1,51 @@
+import React, {Fragment, useEffect, useCallback} from 'react';
```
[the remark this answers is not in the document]
AitorAlgorta: I wasn't sure and because the PR is already really big I didn't want to make changes here

PR #753, opened 2021-01-27
[#654] Replace scss files with scss modules

### LINKED ISSUE #654: Replace scss files with scss modules
In some places of the demo app we have `import './index.scss';` instead of the `import styles from './index.module.scss';`.

We need to check where and replace them all with the proper modules to reduce the possibilities of naming problems.

ogbeche77:
Fixes #654


bitboxer (approved):
LGTM

Issue #706, opened 2021-01-21
Make ngrok tunneling optional when deploying core


ljupcovangelski:
Currently ngrok sidecars are deployed in webhook pods for the sources, so that one can connect a source when running core locally, in a private or local network.

However, when core is deployed in a public network, then the ngrok tunnelling is not needed as the ingestion webhooks for the sources can be configured directly on the public interfaces.

We need to make the ngrok deployment optional inside the helm charts.

ljupcovangelski:
Issue resolved.

Issue #739, opened 2021-01-26
Put the appropriate version of the airy cli in the provisioning script


ljupcovangelski:
For the initial version of the airy cli, it was downloaded from AWS S3, instead from the released version.
https://github.com/airyhq/airy/blob/develop/infrastructure/scripts/provision/core.sh#L21

We need to fix this, so that it is always pulled from the binaries in the release.

PR #756, opened 2021-01-27
[#706] Make NGrok optional

ljupcovangelski:
Conditional deployment of the NGrok sidecars, based on a configuration item in `airy.yaml`.

By default, the installer starts ngrok during the bootstrap process.
If people install the charts outside of the bootstrap process, ngrok is disabled by default.
One can dissable ngrok during bootstrap with
```
NGROK_ENABLED=false ./scripts/bootstrap.sh
```


infrastructure/airy.tpl.yaml
```diff
@@ -52,5 +53,6 @@ core:
       path: "path"
 ingress:
   apiHost: api.airy
+  webhooksHost: "webhooks.airy"
```
ljupcovangelski, line 56: This would be available by default, but we don't need it in /etc/hosts I think.

infrastructure/scripts/status.sh
```diff
@@ -16,19 +16,28 @@ wait-for-service startup-helper api-auth 80 10 api-auth
...
+if [ ${NGROK_ENABLED} = "true" ]; then
+    FACEBOOK_WEBHOOK_PUBLIC_URL="https://fb-${CORE_ID}.tunnel.airy.co/facebook"
+    GOOGLE_WEBHOOK_PUBLIC_URL="https://gl-${CORE_ID}.tunnel.airy.co/google"
+    TWILIO_WEBHOOK_PUBLIC_URL="https://tw-${CORE_ID}.tunnel.airy.co"
+else
+    WEBHOOKS_HOSTNAME=$(kubectl get configmap HOSTNAMES -o jsonpath='{.data.WEBHOOKS_HOST}')
+    FACEBOOK_WEBHOOK_PUBLIC_URL="https://${WEBHOOKS_HOSTNAME}/facebook"
+    GOOGLE_WEBHOOK_PUBLIC_URL="https://${WEBHOOKS_HOSTNAME}/google"
+    TWILIO_WEBHOOK_PUBLIC_URL="https://${WEBHOOKS_HOSTNAME}/twilio"
+fi
```
ljupcovangelski, line 30: The status script prints either the webhook configured in the ingress or the randomly generated Ngrok tunnel url.

infrastructure/helm-chart/charts/apps/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
 appVersion: "1.0"
 description: A Helm chart for the Airy Core Platform application
-name: core
-version: 0.5.0
+name: apps
+version: 1.0
```
ljupcovangelski, line 5: `0.1.0` didn't mean much sense I think, so I put `1.0` as this is required, so that at least it is the same as the `apiVersion`

infrastructure/scripts/provision/core.sh
```diff
@@ -12,20 +11,19 @@ source ${INFRASTRUCTURE_PATH}/scripts/lib/k8s.sh
...
 
-
 echo "Deploying the Airy Core Platform with the ${AIRY_VERSION} image tag"
 
 if [[ -f ${INFRASTRUCTURE_PATH}/airy.yaml ]]; then
     yq eval '.global.appImageTag="'${AIRY_VERSION}'"' -i ${INFRASTRUCTURE_PATH}/airy.yaml
-    helm install core ${INFRASTRUCTURE_PATH}/helm-chart/ --values ${INFRASTRUCTURE_PATH}/airy.yaml --timeout 1000s > /dev/null 2>&1
-    wget -qnv https://airy-core-binaries.s3.amazonaws.com/alpine/airy.gz
+    helm install core ${INFRASTRUCTURE_PATH}/helm-chart/ --values ${INFRASTRUCTURE_PATH}/airy.yaml --set global.ngrokEnabled=${NGROK_ENABLED} --timeout 1000s > /dev/null 2>&1
+    wget -qnv https://airy-core-binaries.s3.amazonaws.com/${AIRY_VERSION}/alpine/amd64/airy.gz
```
ljupcovangelski, line 19: @pascal-airy do you think this will solve https://github.com/airyhq/airy/issues/739 ? 
I created a `develop` directory, where we would need to put the new binary when we merge to develop (if the cli has changed)

pascal-airy: yeah but we should find out how we can build the alpine version with bazel and then the  develop will automatically go into the "*-alpha" folder

infrastructure/scripts/provision/core.sh
```diff
@@ -12,20 +11,19 @@ source ${INFRASTRUCTURE_PATH}/scripts/lib/k8s.sh
 cd ${INFRASTRUCTURE_PATH}/scripts/
 wait-for-service-account
 
-
 echo "Deploying the Airy Core Platform with the ${AIRY_VERSION} image tag"
 
 if [[ -f ${INFRASTRUCTURE_PATH}/airy.yaml ]]; then
     yq eval '.global.appImageTag="'${AIRY_VERSION}'"' -i ${INFRASTRUCTURE_PATH}/airy.yaml
```
pascal-airy, line 17: I think you have to do the same as me here for NGROK or else when you enable ngrok via the env var but have it disabled in your airy.yaml it will be disabled

pascal-airy: ah no sorry that was just before we had the controller 👌

pascal-airy (dismissed):
lgtm!

infrastructure/airy.tpl.yaml
```diff
@@ -3,54 +3,56 @@ global:
   appImageTag: develop
   containerRegistry: ghcr.io/airyhq
   namespace: default
```
pascal-airy, line 5: but isn't  ngrok setting missing here?

ljupcovangelski: We can also put it there, but i think It is not needed as the only place we can overwrite it is during the bootstrap. And there is a default `false` inside the chart.

infrastructure/scripts/provision/core.sh
ljupcovangelski: Yes, that we must solve anyway, agreed :+1: But I think we must have the folder `develop`, otherwise we should push the binaries with the suffix, to a future-release-folder? Or is there a `future-release-alpha` folder created by default?

pascal-airy (approved):

PR #770, opened 2021-01-29
Fix/750 rename response metadata

### LINKED ISSUE #750: Rename ResponseMetadata
As we now have a metadata concept/model, this name is confusing. Suggestion: `PaginationData`

AudreyKj:
resolves #750 

BACKEND 
- Renamed response_metadata to pagination_data / paginationData

FRONTEND 
- Renamed all fields response_metadata to pagination_data
- Changed types in Actions of paginated endpoints (conversations + messages) using PaginatedPayload
- Changed types in Reducers of paginated endpoints (conversations + messages) 

HTTP TS LIBRRAY 
- Deleted ResponseMetadata from the lib 
- Exposed (exported) PaginatedPayload from the lib 

DOCS 
- Renamed response_metadata to pagination_data


chrismatix (dismissed):
Ah, so that's why IntelliJ 😄  Good work!

lib/typescript/httpclient/payload/PaginatedPayload.ts
```diff
@@ -1,4 +1,4 @@
 export interface PaginatedPayload<T> {
   data: T[];
-  response_metadata: {previous_cursor: string; next_cursor: string; total: number};
+  pagination_data: {previous_cursor: string; next_cursor: string; total: number};
```
lucapette, line 3: I was waiting for this pr so we could fix the problems here. It's maybe a little annoying though. We'd need:

- change the casing so that it reflects JS conventions
- have a mapper in the library so that we can convert between two (even though... maybe there's already mappers and we can enrich them)

AudreyKj: ok! fixing now

lucapette (approved):
:shipit:

Issue #491, opened 2020-12-14
Make 'edit page' link depend on the version of the docs you're looking at


lucapette:
it should point to `develop` for `develop.docs.airy.co` and to `main` for `docs.airy.co`. We must mention the docs version in the readme and show a different footer depending on the version (something like 'you're looking at x.v.z version of the docs, See this for beta version' and the other way around)

PR #771, opened 2021-01-29
[#491] Do not show the edit button for docs

lucapette:
Until we sort out how we want to handle versioned docs


sgarroum (approved):
Keeping it clean till we have a proper setup running :)

PR #795, opened 2021-01-31
[#791] Message Input UI/Design Improvement

### LINKED ISSUE #791: Message Input bar UI Improvement
The font and text area design needs to be similar to what we have on platform

ogbeche77:
Fixes #791

### DIFF
--- a/frontend/demo/src/pages/Inbox/MessageInput/index.tsx
+++ b/frontend/demo/src/pages/Inbox/MessageInput/index.tsx
@@ -41,7 +41,7 @@ const MessageInput = (props: MessageInputProps & ConnectedProps<typeof connector
   };
 
   return (
-    <div className={`${styles.container} ${styles.flexWrap}`}>
+    <form className={`${styles.container} ${styles.flexWrap}`}>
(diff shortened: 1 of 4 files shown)


frontend/demo/src/pages/Inbox/MessageInput/index.tsx
```diff
@@ -41,25 +41,29 @@ const MessageInput = (props: MessageInputProps & ConnectedProps<typeof connector
   };
 
   return (
-    <form className={styles.container}>
```
bitboxer, line 44: it was a form because of accesbility resons. If something looks like a form, it should be a form. This helps screen readers a lot. Not that we have a super focus on this, but we should at least try to do a few things into the right direction :D

bitboxer (changes requested):
See comment below.

frontend/demo/src/pages/Inbox/MessageInput/index.tsx
ogbeche77: Thanks for the insight

AitorAlgorta (approved):

bitboxer (approved):
LGTM

Issue #722, opened 2021-01-25
AKHQ deployment


pascal-airy:
AKHQ or formerly known KafkaHQ is a useful tool for Kafka debugging that we would like to add as an optional core component

ljupcovangelski:
Deployment done.

Issue #846, opened 2021-02-02
Introduce configuration document in the deployment section


lucapette:
We describe `airy.yaml` and how it's used via the `airy config apply` and link back to this document from the rest of the section (and wherever else we mention config)

PR #847, opened 2021-02-02
[#722] Add AKHQ as optional tool

ljupcovangelski:
AKHQ is deployed if `tools.akhq.enabled=true` in the `airy.yaml` file.
Basic authentication is enabled. Default username is `admin`. The password can be retrieved from a configMap:
```
kubectl get configmap akhq-config -o jsonpath="{.data.password}"
```

### DIFF
--- a/infrastructure/airy.tpl.yaml
+++ b/infrastructure/airy.tpl.yaml
@@ -5,7 +5,7 @@ global:
   namespace: default
   ingress:
     apiHost: api.airy
-    webhooksHost: "webhooks.airy"
+    webhooksHost: webhooks.airy
     uiHost: ui.airy
     chatpluginHost: chatplugin.airy
 prerequisites:
@@ -56,3 +56,6 @@ apps:
         bucket: "changeme"
         region: "changeme"
         path: "path"
+tools:
+  akhq:
+    enabled: false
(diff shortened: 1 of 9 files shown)


infrastructure/helm-chart/values.yaml
```diff
@@ -8,3 +8,4 @@ global:
     webhooksHost: "webhooks.airy"
     uiHost: "ui.airy"
     chatpluginHost: "chatplugin.airy"
+    toolsHost: "tools.airy"
```
ljupcovangelski, line 11: Akhq will be deployed and accessible through akhq.{toolsHost} or `akhq.tools.airy`.
I didn't add the DNS entry for the Vagrantbox, as `/etc/hosts` doesn't accept wildcard certificates.

Open to suggestions for this, maybe one option would be to do `tools.airy/akhq`, but this doesn't work out of the box and some tweak would be needed for traefik to strip the `/akhq` prefix, before the request is sent to the app.

pascal-airy (approved):
glad you made the auth work :)

infrastructure/airy.tpl.yaml
```diff
@@ -56,3 +56,6 @@ apps:
         bucket: "changeme"
         region: "changeme"
         path: "path"
+tools:
+  akhq:
+    enabled: true
```
lucapette, line 61: I wouldn't have this enabled by default

lucapette (changes requested):
I like it! but it's strange that it's an optional component and it's enabled by default.

We should also provide some docs about it somewhere but I suppose that can be done in the #846 

We do need a followup ticket to make this available in the vagrant instance though

lucapette (approved):
:shipit:

PR #850, opened 2021-02-03
[#575] Fix window: any type definition

### LINKED ISSUE #575: Fix window: any type definition
We have two places where we define `declare const window: any;`. This has to be fixed. The reason is that we added the `airy` object to the window object and need to extend it. 

The best way would be to globally extend the window interface like so:

```js
declare global {
    interface Window {
        userObject: UserInterface
    }
}
```

see [here](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#global-augmentation) for details.

bitboxer:
closes #575

### DIFF
--- a/frontend/ui/src/env.ts
+++ b/frontend/ui/src/env.ts
@@ -2,7 +2,11 @@ export interface Env {
   API_HOST?: string;
 }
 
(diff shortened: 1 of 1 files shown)


AitorAlgorta (approved):
🏆

Issue #827, opened 2021-02-02
Chatplugin uses its own payload


paulodiniz:
Currently the `chat plugin` has it's own response mapper, causing the response of `/chatplugin.send` to be different from the other sources.

Should `/source.send` endpoints have different response payloads?

lucapette:
In general, I'm not so sure about it. But for the message payload, I think we're kind of forced into sharing the same payload by the decisions we took (and I think make sense) about the way we render content on the frontend

chrismatix:
Agree with what Luca said. For messagse the chat plugin client will have to use the message rendering library (otherwise the source template tickets don't make a lot of sense) so conceptually the chat plugin message has to be the same as the core message.

lucapette:
We can/should use the message repository for this then

PR #853, opened 2021-02-03
[#827] Chatplugin uses its own payload

paulodiniz:
Fixes https://github.com/airyhq/airy/issues/827


backend/model/message/src/main/java/co/airy/model/message/MessageMapper.java
```diff
@@ -0,0 +1,4 @@
+package co.airy.model.message;
+
+public class MessageMapper {
```
chrismatix, line 3: delete empty class

backend/model/message/src/main/java/co/airy/model/message/dto/MessageResponsePayload.java
```diff
@@ -0,0 +1,44 @@
...
+
+import static co.airy.date.format.DateFormat.isoFromMillis;
+import static co.airy.model.message.MessageRepository.resolveContent;
+
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class MessageResponsePayload {
+    private String id;
```
chrismatix, line 17: cc @AitorAlgorta  @lucapette there was a conversation about having the `source` field on every message that comes in so that they can be rendered even without the conversation (e.g. websocket). Is that still relevant?

backend/model/message/src/main/java/co/airy/model/message/MessageMapper.java
paulodiniz: true 👍

backend/model/message/src/main/java/co/airy/model/message/dto/MessageResponsePayload.java
paulodiniz: Good point!

lucapette (approved):
:shipit:

PR #857, opened 2021-02-03
[#841] Logout if user auth token is wrong

### LINKED ISSUE #841: Logout if user auth token is wrong
If we rebuild the vagrant box and create a new box, the browser still has the login auth token for the UI app. If the UI cannot do anything with the token it has, it should realize it is broken and logout the current user. Currently it ends up in a weird state where most clicks only scream errors in the console, but don't work in the UI.

bitboxer:
This also removes the catch and return stuff. We do not want
to return non okay values as objects because those won't fit
the datamodel we expect in the promise chain later on.

Also: we might want to type the results somehow. We should have
seen that the return value differs via typescript?

closes #841

### DIFF
--- a/frontend/ui/src/InitializeAiryApi.ts
+++ b/frontend/ui/src/InitializeAiryApi.ts
@@ -1,7 +1,14 @@
 import {HttpClient} from 'httpclient';
 import {getAuthToken} from './cookies';
 import {env} from './env';
+import {store} from './store';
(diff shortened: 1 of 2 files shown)


chrismatix (approved):
Very good solution 👏

lucapette (dismissed):
:shipit:

lucapette (approved):

AitorAlgorta (approved):

PR #864, opened 2021-02-03
Fix facebook text render

### LINKED ISSUE #844: Rendering from Facebook doesn't work
Text messages from facebook are not working

chrismatix:
fixes #844


tsconfig.json
```diff
@@ -1,7 +1,10 @@
 {
   "extends": "./bazel.tsconfig.json",
+  "compilerOptions": {
+    "noEmit": true
```
chrismatix, line 4: This is helpful for development, since it prevents you from accidentally writing a massive `.js` and `.js.map` file tree

frontend/ui/src/selectors/conversations.ts
```diff
@@ -3,35 +3,42 @@ import {filter, reverse, sortBy, values} from 'lodash-es';
 import {Conversation} from 'httpclient';
 import {StateModel} from '../reducers';
 import {ConversationMap} from '../reducers/data/conversations';
+import {ConversationRouteProps} from "../pages/Inbox";
+
+export const getCurrentConversation = (state: StateModel, props: ConversationRouteProps) =>
```
chrismatix, line 8: @ogbeche77 This is what I meant in our call the other week :) This selector allows every component within the inbox page that is connected to the router using `withRouter` to access the currently open conversation in the store.

lib/typescript/dates/format.ts
```diff
@@ -1,37 +1,36 @@
...
-  today.setHours(0, 0, 0, 0);
-  const beginningOfWeek = new Date(today.getTime() - today.getDay() * 24 * 60 * 60 * 1000);
-  return date > beginningOfWeek;
+    const today = new Date();
+    today.setHours(0, 0, 0, 0);
+    const beginningOfWeek = new Date(today.getTime() - today.getDay() * 24 * 60 * 60 * 1000);
+    return date > beginningOfWeek;
 }
 
-export function formatTimeOfMessage(message: Message) {
```
chrismatix, line 14: No need for a dependency on message and therefore httplib

chrismatix:
So the layout idea for the render library has three constraints:
- Sources need to be pluggable
- Rendering components should be re-usable e.g. to avoid writing one text component per source
- Content mapping and rendering within a source should be decoupled for better structure

image

As you can see in the diagram the source message is a simple react component and the only interface of the library. Under the hood it determines the source of the message and picks the correct render provider. Providers do two things:
- Parse the content and map it to a source model
- Render that source model to a react component

lib/typescript/httpclient/model/Message.ts
```diff
@@ -47,6 +47,12 @@ export enum SenderType {
   sourceUser = 'source_user',
   appUser = 'app_user',
 }
+
+export function isFromContact(message: Message) {
```
chrismatix, line 51: Let's try and keep all methods that only operate on the content model right in the `model/` folder :) This way we can easily replace these methods if the logic changes and don't have go though UI code to change it.

lib/typescript/render/index.tsx
```diff
@@ -1,26 +1,22 @@
...
+    const provider = renderProviders[source];
+
+    try {
+        return provider(props);
+    } catch (e) {
+        console.error(e);
+        return <Text
+            {...getSharedComponentProps(props)}
+            text="Could not render this content"
+        />
```
chrismatix, line 20: For resilience reason we by default assume that we cannot render content. After all, we don't control the content schema so the render library is a "best effort" approach to solving the problem.

lib/typescript/render/index.tsx
```diff
@@ -1,26 +1,19 @@
...
+  const source = getSource(props.conversation);
 
-export default RenderLibrary;
+  const provider = renderProviders[source];
+
+  try {
+    return provider(props);
+  } catch (e) {
+    console.error(e);
+    return <Text {...getSharedComponentProps(props)} text="Could not render this content" />;
```
chrismatix, line 17: For resilience reason we by default assume that we cannot render content. After all, we don't control the content schema so the render library is a "best effort" approach to solving the problem.

lib/typescript/dates/BUILD
```diff
@@ -5,7 +5,6 @@ package(default_visibility = ["//visibility:public"])
 ts_library(
     name = "dates",
     deps = [
-        "//lib/typescript/httpclient",
```
lucapette, line 8: 🙈

AitorAlgorta (approved):

lucapette (approved):
Amazing work 👏 👏 👏

PR #873, opened 2021-02-04
[#846] Docs for airy config yaml

pascal-airy:


docs/docs/getting-started/deployment/configuration.md
```diff
@@ -0,0 +1,71 @@
+---
+title: Configuration
+sidebar_label: Configuration
+---
+
+## Airy Core Configuration File
+
+The `infrastructure/airy.tpl.yaml` file contains examples of all possible configuration options. Copy and rename it to `infrastructure/airy.yaml` and modify it according to your environment:
```
lucapette, line 8:
```suggestion
The `infrastructure/airy.tpl.yaml` file contains examples of all possible configuration options. Create your own `airy.yaml` like so:
```

docs/docs/getting-started/deployment/configuration.md
```diff
@@ -0,0 +1,71 @@
...
+
+```bash
+cd infrastructure
+cp airy.tpl.yaml airy.yaml
+```
+
+We will guide you through the different sections so you can make the changes you are looking for. Keys can also be omited for services which you do not wish to configure.
```
lucapette, line 15:
```suggestion
We will guide you through the different sections so you can make the changes you are looking for. Keys can also be omitted for services which you do not wish to configure.
```

docs/docs/getting-started/deployment/configuration.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Global
+
+- **appImageTag**
+
+  image tag of the container images for the `Airy Components`
```
lucapette, line 21: We should probably specific what kind of value they can expect here (as in the "official released" tags plus develop)

docs/docs/getting-started/deployment/configuration.md
```diff
@@ -0,0 +1,71 @@
...
+
+- **containerRegistry**
+
+  The images of the `Airy Components` are stored in the [Github Container Registry](https://docs.github.com/en/packages/guides/about-github-container-registry).
```
lucapette, line 25: The subject here should be something like "The URL of the container registry" I think (unless I'm missing something)

docs/docs/getting-started/deployment/configuration.md
```diff
@@ -0,0 +1,71 @@
...
+
+- **namespace**
+
+  Kubernetes namespace the `Airy Core` will use
```
lucapette, line 29:
```suggestion
  the Kubernetes namespace the `Airy Core` will use
```

docs/docs/getting-started/deployment/configuration.md
```diff
@@ -0,0 +1,71 @@
...
+- **redis**
+- **postgres**
+
+  We recommend that you create a new database if you are reusing a PostgreSQL server to avoid name collisions.
+
+### Components
+
+- **api**
+
+  `mail*` settings are needed if you want the Airy Core to send emails out
```
lucapette, line 48: as there's no "public feature" using this (and I believe... we will actually never have one as there's better ways to achieve this integration), I would remove this one

docs/docs/getting-started/deployment/configuration.md
```diff
@@ -0,0 +1,71 @@
...
+
+  `jwtSecret` should be set to a long secure secret in production environments
```
lucapette, line 50: this is the HTTP api JWT signing key right? I think the name has to be more explicit (I get we'd have to change the code... but if not now when? :))

docs/docs/getting-started/deployment/configuration.md
```diff
@@ -0,0 +1,71 @@
...
+
+  `allowedOrigins` set this to your sites origin
+
+- **sources**
+
+  The `Airy Controller` will only start `Airy Components` when they are configured here. So to keep system load to a minimum only add the sources you are using.
```
lucapette, line 56:
```suggestion
  The `Airy Controller` only starts configured sources. To keep system load to a minimum, only add the sources you are using.
```

docs/docs/getting-started/deployment/configuration.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Applying the configuration
+
+If you made changes in `airy.yaml` and want to apply it to your instance you can use the [CLI](getting-started/cli.md) by running the following command.
+
+```bash
+airy config apply --config ./airy.yaml --kube-config /path/to/your/kube.conf
+```
+
+The Airy CLI considers that the Kubernetes configuration file is located under
```
lucapette, line 69: this info belongs to the CLI docs I think (which also makes me think we should follow up with a ticket and have a `--config-dir` option). I will create the ticket!

docs/sidebars.js
```diff
@@ -9,6 +9,7 @@ module.exports = {
             'getting-started/deployment/introduction',
             'getting-started/deployment/vagrant',
             'getting-started/deployment/production',
+            'getting-started/deployment/configuration',
```
lucapette, line 12: Logically, I feel like this should come after introduction but maybe a good idea to also ask @sgarroum

lucapette (commented):
Amazing work!

docs/docs/getting-started/deployment/configuration.md
pascal-airy: love it!

docs/docs/getting-started/deployment/configuration.md
pascal-airy: I just don't want people to encourage changing it but yeah it's their fault if they do so I'll change it

docs/docs/getting-started/deployment/configuration.md
pascal-airy: ok I'll link it here instead then

docs/docs/getting-started/deployment/production.md
```diff
@@ -141,37 +141,26 @@ to proceed with deploying the apps, we assume that you have a running Kubernetes
...
-network connectivity to the required services.
+the `Airy Components` inside a Kubernetes cluster.
+
+We provided a Helm chart to deploy the `Airy Components`. Before you can run helm, you
+must configure the system via the `airy.yaml` file, then you can proceed:
 
-The file `infrastructure/airy.tpl.yaml` contains an example of all possible
-configuration parameters. This file should be copied to `airy.yaml` and edited
-according to your environment:
+See our [Configuration Docs](/getting-started/deployment/configuration.md) for detailed instructions.
```
ljupcovangelski, line 153: Perhaps we move this line inside the previous sentence?

docs/docs/getting-started/deployment/configuration.md
```diff
@@ -0,0 +1,79 @@
...
+  - `storage`
+    - `s3` set these to your AWS S3 config to store source specific user data
+
+## Applying the configuration
+
+If you made changes in `airy.yaml` and want to apply it to your instance you can use the [airy config apply](/cli/airy_config_apply.md) by running the following [CLI](getting-started/cli.md) command.
+
+```bash
+airy config apply --config ./airy.yaml --kube-config /path/to/your/kube.conf
+```
```
ljupcovangelski, line 79: I think that the content of the document is good, but this last command, I would put it inside the `Vagrant` and `Run-in-production` docs, so that all the commands which one needs to run are in a single document.

ljupcovangelski (approved):
Nice :+1:

PR #902, opened 2021-02-05
[#852] Feature/852 add google rich text to render library

### LINKED ISSUE #852: Render Rich Text Message from Google
```
{
   "conversation_id": "UUID",
   "message": {
      "text": "some string with **markdown**",
      "fallback": "some string without markdown",
      "containsRichText": "true"
   }
}
```

thorstenairy:
closes #852


bitboxer (approved):
I like it, but let's wait till @chrismatix had time to have a look at it. He is the visionary behind the api design 😉 .

lib/typescript/render/components/RichText/index.module.scss
```diff
@@ -0,0 +1,59 @@
...
+  height: 40px;
+  margin: 6px 8px 0 0;
+}
+
+.messageListUserContainer {
+  display: flex;
+  flex-direction: row;
+}
+
+.messageListItemUser {
```
chrismatix, line 23: I'd rename the classes since this is no longer in the message list item

lib/typescript/render/components/RichText/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+type RichTextRenderProps = DefaultMessageRenderingProps & {
+  message: Message;
+};
+
+export const RichText = (props: RichTextRenderProps) => {
+  const {message, fromContact, sentAt, contact} = props;
+  const messageJSON = JSON.parse(message.content);
+  const messageText = messageJSON.text;
+  const messageFallback = messageJSON.fallback;
+  const messageIsRichText = messageJSON.containsRichText;
```
chrismatix, line 16: All these values should be on the properties and the properties should be injected in the [render](https://github.com/airyhq/airy/pull/902/files#diff-c11b3c1a17a44fc8a03e882cce37071b72413f235bea2a7617b98d7b42629c85R15) function. Right now this UI component is aware of message content structure.

lib/typescript/render/providers/chatplugin/ChatPluginRender.tsx
```diff
@@ -1,24 +1,59 @@
...
+function chatpluginInbound(message: Message): ContentUnion {
+  const messageContent = JSON.parse(message.content);
+  if (messageContent && messageContent.containsRichText) {
+    return {
+      type: ContentType.richText,
+      text: messageContent.text,
+      fallback: messageContent.fallback,
+      containsRichtText: messageContent.containsRichText,
+    };
+  } else {
```
chrismatix, line 36: nitpick: Here and later you can save yourself the `else` since the `if` has a return

chrismatix (changes requested):

lib/typescript/render/components/RichText/index.module.scss
```diff
@@ -0,0 +1,59 @@
+@import 'assets/scss/fonts.scss';
+@import 'assets/scss/colors.scss';
+
+.messageContainer {
```
chrismatix, line 4: another nitpick: If every class in the file has the same prefix, the prefix doesn't add anything. It's a case of [context duplication](https://github.com/kettanaito/naming-cheatsheet#avoid-context-duplication)

lib/typescript/render/providers/chatplugin/ChatPluginRender.tsx
```diff
@@ -23,19 +38,54 @@ function render(content: ContentUnion, props: MessageRenderProps) {
...
         />
       );
+
+    // TODO render more chatplugin models
   }
 }
 
-function mapContent(message: Message): ContentUnion {
+// TODO map more string content to chatplugin models
+function chatpluginInbound(message: Message): ContentUnion {
```
chrismatix, line 47: I think for chatplugin in-/ outbound are the same so you can have one method instead

chrismatix (commented):
Great improvements! Left some more comments

lib/typescript/render/components/RichText/index.module.scss
thorstenairy: so true!

AitorAlgorta (approved):
LGTM

chrismatix (approved):

PR #904, opened 2021-02-05
[#854] Render File from Facebook

### LINKED ISSUE #854: Render File from Facebook
```
{
   "conversation_id": "UUID",
   "message": {
      "attachment":{
        "type":"image", 
        "payload":{
          "is_reusable": true,
          "url":"http://www.messenger-rocks.com/image.jpg"
      }
    }
}
```

bitboxer:
closes #854

### DIFF
--- a/lib/typescript/render/components/Image/index.module.scss
+++ b/lib/typescript/render/components/Image/index.module.scss
@@ -31,13 +31,11 @@
   position: relative;
 }
 
-.messageListItemUserText {
+.messageListItemUserImage {
   display: inline-flex;
-  padding: 10px;
   margin-top: 5px;
-  background: var(--color-background-blue);
-  color: var(--color-text-contrast);
   border-radius: 8px;
+  overflow: hidden;
 }
 
 .messageListItemMember {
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "eslint": "eslint frontend/"
+    "eslint": "eslint ./"
```
bitboxer, line 6: Sorry @chrismatix , but I really had to fix this 😉 .

lib/typescript/render/providers/facebook/facebookModel.ts
```diff
@@ -1,12 +1,42 @@
...
+}
+
+export enum ContentType {
+  text = 'text',
+  image = 'image',
+}
+
 export interface Content {
-  type: 'text';
+  type: ContentType.image | ContentType.text;
```
chrismatix, line 28: So this is a very subtle difference, but I think I would prefer string unions here for the reason that Typescript enums actually get compiled to a Javascript snippet. This snippet we don't really need since we control the intermediary content payloads. For string unions, it'd be statically checked at compile time only.

Also this may be a question of taste, but I feel like it's easier to grasp the available file types [this way](https://github.com/airyhq/airy/blob/develop/lib/typescript/types/content.ts#L11)

lib/typescript/render/components/Image/index.module.scss
```diff
@@ -0,0 +1,63 @@
+@import 'assets/scss/fonts.scss';
+@import 'assets/scss/colors.scss';
+
+.wrapper {
+  display: flex;
+  flex: none;
+}
+
+.messageListItem {
```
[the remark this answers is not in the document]
bitboxer: I will keep it like this because it makes refactoring and extrating it easier :D.

chrismatix (approved):

Issue #883, opened 2021-02-04
Add metadata field to channels API 


paulodiniz:
Add `metadata` to the channels API.

We can use this library to output the JSON https://github.com/airyhq/airy/pull/891

lucapette:
the thing with the naming here is that the endpoints should be called `userdata.upsert` (which is 100% horrible :D). So I can't really propose anything better than:

- `/metadata.upsert` 
- `{metadata: {contact, tags..., custom: {}}`

I don't love it... but less confusing than what we have already. My fav solution would be:

`{name_i_dont_have_yet: {contact, tags, ..., metadata: {}}}` (but maybe there's really not a good name for this

chrismatix:
`userMetadata` is overly descriptive. Perhaps that works?

lucapette:
@chrismatix it does... but my point is also that it should stay in sync with the endpoint. So I'm not sure you're suggesting `/usermetadata.update` (or maybe this is a non-requirement? That's also a valid answer I think)

chrismatix:
It seems intuitive, but if the request payload already contains `userMetadata`, I think we can keep the endpoint's name `/metadata.upsert` and it still makes sense.

lucapette:
yeah I think it's reasonable!

PR #908, opened 2021-02-08
WIP: Channel metadata API changes

chrismatix:
channels part for #883


backend/api/communication/src/main/java/co/airy/core/api/communication/Mapper.java
```diff
@@ -18,10 +20,13 @@
 public class Mapper {
 
     public ConversationResponsePayload fromConversation(Conversation conversation) {
+
         return ConversationResponsePayload.builder()
                 .channel(ChannelPayload.builder()
                         .id(conversation.getChannelId())
-                        .name(conversation.getChannel().getName())
+                        // TODO https://github.com/airyhq/airy/issues/909
```
chrismatix, line 27: So to get the metadata here we would have to change the topology the same way as required for #909. I will default the name on the frontend and we finish this up after doing #909

backend/model/channel/src/main/java/co/airy/model/channel/dto/ChannelContainer.java
```diff
@@ -0,0 +1,33 @@
...
+@AllArgsConstructor
+public class ChannelContainer implements Serializable {
+    private Channel channel;
+    private MetadataMap metadataMap;
+
+    public MetadataMap getMetadataMap() {
+        return this.metadataMap != null ? this.metadataMap : getDefaultMetadata();
+    }
+
+    private MetadataMap getDefaultMetadata() {
```
chrismatix, line 27: This ensures that there will always be a `channel.metadata.name` for the clients to use

lib/typescript/httpclient/model/Channel.ts
```diff
@@ -1,8 +1,14 @@
+import {Metadata} from './Metadata';
+
+export type ChannelMetadata = Metadata & {
+  name: string;
+  image_url?: string;
```
bitboxer, line 5: Shouldn't this be `imageUrl` ? We decided to have camelCase for every interface that exposes a structure to the outside of the http client lib

bitboxer (commented):
One question, otherwise the JS code looks nice.

lib/typescript/httpclient/model/Channel.ts
chrismatix: Yes I will change this

lib/typescript/httpclient/mappers/channelMapper.ts
```diff
@@ -4,5 +4,9 @@ import {ChannelApiPayload} from '../payload/ChannelApiPayload';
 export const channelMapper = (payload: ChannelApiPayload): Channel => ({
   ...payload,
   sourceChannelId: payload.source_channel_id,
+  metadata: {
+    ...payload.metadata,
+    imageUrl: payload.metadata.image_url,
```
bitboxer, line 9: Nitpick: doesn't this mean you will end up with `image_url` and `imageUrl` in the data?

chrismatix: Yes, that's a good point actually :/ The problem is that `metadata` can contain structures provided by the user. Ok so I will make the mapping here explicit for now and we figure out how to do this at some other point.

lucapette (approved):

Issue #858, opened 2021-02-03
Render Rich Card (Google style)


AitorAlgorta:
```
{
   "conversation_id": "UUID",
   "message": {
    "fallback": "Hello, world!\n\nReply with \"A\" or \"B\"",
    "richCard": {
      "standaloneCard": {
        "cardContent": {
          "title": "Hello, world!",
          "description": "Sent with Business Messages.",
          "media": {
            "height": "TALL",
            "contentInfo":{
              "altText": "Google logo",
              "fileUrl": 'https://www.google.com/images/branding/googlelogo/2x/googlelogo_color_272x92dp.png',
              "forceRefresh": "false"
            }
          },
          "suggestions": [
            {
              "reply": {
                "text": "Suggestion #1",
                "postbackData": "suggestion_1"
              }
            },
            {
              "reply": {
                "text": "Suggestion #2",
                "postbackData": "suggestion_2"
              }
            }
          ]
        }
      }
    }      
   }
}
```

lizhutter:
![103911413-4d17ef80-5106-11eb-8d76-561c35fb8da6](https://user-images.githubusercontent.com/74183162/106778500-55504580-6646-11eb-8439-d76d239e173d.png)

PR #912, opened 2021-02-08
fixed rich card rendering with product requirements

### LINKED ISSUE #911: fix rendering of rich cards with product requirements
requirements:
- The media template should be designed on the UI to have at least one image and one button
- The button link could be an external link or a trigger for a suggested reply
- Title and description are optional fields, so the UI should not break in case these fields were missing

AudreyKj:
resolves #911 / related to #858

note: atm buttons contain external links, perhaps this would have to be fixed in the future if we want the buttons to link to another template in addition to external links

changes: 
- Title, description, and one suggestion button are optional fields
 - The template is designed to have at least one image and one button
- time is removed from the component (as we will externalize this)

### DIFF
--- a/lib/typescript/render/components/RichCard/index.tsx
+++ b/lib/typescript/render/components/RichCard/index.tsx
@@ -3,40 +3,52 @@ import styles from './index.module.scss';
 import {Media, MediaRenderProps} from '../Media';
 import {DefaultMessageRenderingProps} from '../index';
 
-interface Suggestion {
-  reply: {
-    text: string;
-    postbackData: string;
-  };
-}
+type Suggestions = [
+  {
+    reply: {
+      text: string;
+      postbackData: string;
+    };
+  },
+  {
+    reply?: {
+      text: string;
+      postbackData: string;
+    };
+  }
+];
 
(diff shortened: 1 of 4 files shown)


lib/typescript/render/components/RichCard/index.tsx
```diff
@@ -3,40 +3,52 @@ import styles from './index.module.scss';
...
 import {DefaultMessageRenderingProps} from '../index';
 
-interface Suggestion {
-  reply: {
-    text: string;
-    postbackData: string;
-  };
-}
+type Suggestions = [
+  {
```
bitboxer, line 7: Short question: can there only be  two suggestions? Or is this possibly an array of suggestions?

bitboxer (approved):
LGTM, one little question.

lib/typescript/render/components/RichCard/index.tsx
AudreyKj: there can only be two suggestions maximum (=2 buttons) and at least one suggestion

bitboxer: Ah, understood.

PR #928, opened 2021-02-09
New Airy websocket using Airy events

### LINKED ISSUE #884: New Airy Event for Websocket
We need to introduce an `AiryEvent` to the websocket payload that contains the following information

- `event_type`: `message|metadata|channel`
- `data`: event payload

Websocket clients  can determine which events they want to subscribe.

To avoid a big PR or breaking the clients, we should create a new app `api-websocket`, instead of modifying the current `api-communication`

chrismatix:
resolves #884

### DIFF
--- a/docs/docs/api/websocket.md
+++ b/docs/docs/api/websocket.md
@@ -20,30 +20,25 @@ To execute the handshake with `/ws.communication` you must set an
 `Authorization` header where the value is the authorization token obtained [from
 the API](/api/introduction#authentication).
 
-## Outbound Queues
+## Event Types
 
-Outbound queues follow the pattern `/queue/:event_type[/:action}]` and deliver
-JSON encoded payloads.
+All event updates are sent to the `/queue/updates` queue as JSON encoded payloads. The `type`
+field informs the client of the kind of update that is encoded in the payload.
 
 ### Message
 
-`/queue/message`
-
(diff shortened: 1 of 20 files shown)


docs/docs/api/websocket.md
```diff
@@ -20,93 +20,71 @@ To execute the handshake with `/ws.communication` you must set an
...
 
 ```json5
 {
-  conversation_id: "{UUID}",
-  //unique conversation id
-  unread_message_count: 42,
-  //the number of unreaded messages in this conversation
-  timestamp: "{string}"
-  //'yyyy-MM-dd'T'HH:mm:ss.SSSZ' date in UTC form, to be localized by clients
+  "type": "metadata",
```
chrismatix, line 60: @lucapette and @paulodiniz So there are two ways of doing this:

- Send every single metadata record that we get immediately using a consumer group (advantage: very fast)

- Aggregate metadata to a map and emit complete metadata objects on every websocket event (advantage: clients always get the latest metadata document state, disadvantage: slower, can't use a consumer group)

What are your thoughts? Locally I have the former version, but I feel like the latter will need to a nicer API for clients.

chrismatix: Resolved offline: The latter approach is better for the reasons listed above

lucapette (approved):

PR #937, opened 2021-02-10
[#910] Add message metadata API documentation

### LINKED ISSUE #910: Add message metadata API
All message API payloads should have a `metadata` field. For now only the message list has access to metadata, so for all other payloads the value will default to `{}`.

We can use this library to output the JSON https://github.com/airyhq/airy/pull/891

paulodiniz:
Fixes #910

### DIFF
--- a/docs/docs/api/endpoints/messages.md
+++ b/docs/docs/api/endpoints/messages.md
@@ -39,8 +39,12 @@ latest.
       // See glossary
       "sent_at": "{string}",
       //'yyyy-MM-dd'T'HH:mm:ss.SSSZ' date in UTC form, to be localized by clients
-      "source": "{String}"
+      "source": "{String}",
       // one of the possible sources
+      "metadata": {
+        "sentFrom": "iPhone"
+      }
+      // metadata object of the message
     }
   ],
   "pagination_data": {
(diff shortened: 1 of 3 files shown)


docs/docs/api/endpoints/messages.md
```diff
@@ -39,8 +39,16 @@ latest.
       // See glossary
       "sent_at": "{string}",
       //'yyyy-MM-dd'T'HH:mm:ss.SSSZ' date in UTC form, to be localized by clients
-      "source": "{String}"
+      "source": "{String}",
       // one of the possible sources
+      "metadata": {
+         "source": {
```
chrismatix, line 45: nitpick: I'd use a different example here since this is conversation metadata: Perhaps something like `metadata.sentFrom=iphone` or something like that :)

paulodiniz: I actually picked this property for this reason specifically. It's metadata that the user didn't input so it doesn't seem that she/he missed a step.

I can definitely change it, it's just "funny" how we had different perceptions of the same data

paulodiniz:
I don't understand why the CI is failing *now* complaining about using `Map<string, object>`

```

lucapette (approved):

chrismatix (approved):
:shipit:

Issue #914, opened 2021-02-08
Fix the ingress URL for AKHQ in Vagrant


ljupcovangelski:
Currently, AKHQ is deployed with the following host configuration in the ingress: `akhq.tools.airy`. This is not provisioned in the `/etc/hosts` file because we don't want to add entries there for every tool we put. It would be better if we can use `tools.airy/akhq`, so that all the tools are like a sub-path, under the same domain name.

ljupcovangelski:
URL fixed with path instead of a subdomain.

PR #940, opened 2021-02-10
[#914] Add ingress and docs for AKHQ

ljupcovangelski:


docs/docs/getting-started/deployment/configuration.md
```diff
@@ -70,6 +70,13 @@ These settings are used to connect the **Airy Components** to your Kafka cluster
   - `storage`
     - `s3` set these to your AWS S3 config to store source specific user data
 
+### Tools
+
+These settings are used to enable or disable some external tools, used to monitor or debug the **Airy Core Platform**.
```
pascal-airy, line 75:
```suggestion
These settings are used to enable or disable some external tools, used to monitor or debug the **Airy Core**.
```
saying platform is so last week 😎

pascal-airy (commented):
I like that you call it akhq now everywhere. We should also get used to the new name 😎 
I can also create a ticket for the `platform`string and remove it there from all places since I saw it still pops up here and there

lucapette (approved):

PR #941, opened 2021-02-10
Add metadata to conversations API

### LINKED ISSUE #909: Add metadata field to conversation APIs
The metadata must contain `contact`, `tags`, `unread_message_count` and `user_metadata` (old `contact.info` field)

Currently `unread_message_count` is not yet in metadata, so it has to be moved.

We can use this library to output the JSON https://github.com/airyhq/airy/pull/891

TODO:
- [ ] move unread count to metadata
- [ ] move tags to metadata API response
- [ ] consume conversational metadata on clients

chrismatix:
resolves #909


backend/api/communication/src/main/java/co/airy/core/api/communication/Stores.java
```diff
@@ -212,6 +218,32 @@ public void deleteMetadata(Subject subject, String key) throws ExecutionExceptio
         producer.send(new ProducerRecord<>(applicationCommunicationMetadata, getId(subject, key).toString(), null)).get();
     }
 
+    public MetadataMap getMetadata(String subjectId) {
+        final ReadOnlyKeyValueStore<String, MetadataMap> metadataStore = getMetadataStore();
+        return metadataStore.get(subjectId);
+    }
+
+    public List<Conversation> addChannelMetadata(List<Conversation> conversations) {
```
chrismatix, line 226: @paulodiniz and @lucapette So we are avoiding a join on the channels metadata and instead look it up on the fly.

Best case would be that we can also get rid of the channel join, but then it will become much harder to build the lucene index

bitboxer (commented):
Really nice refactoring of the typescript part ❤️

paulodiniz (approved):
This is great 👍

PR #943, opened 2021-02-11
[309] homebrew cli tap

pascal-airy:


scripts/upload-cli-binaries.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+set -eo pipefail
+IFS=$'\n\t'
+
+for os in "linux" "darwin" "windows"
+do
+    sha256sum bazel-bin/infrastructure/cli/airy_"$os"_bin > bazel-bin/infrastructure/cli/airy_"$os"_bin.sha
```
ljupcovangelski, line 7: I would recommend the following filename: `airy_"$os"_sha256sum.txt`

pascal-airy: right that is also how kubectl does it. fixed.

ljupcovangelski (approved):

pascal-airy:
Fixes #309

PR #948, opened 2021-02-11
[#675] Add Templates Endpoints

### LINKED ISSUE #675: Add Templates Endpoints
### Templates

We must provide endpoints for:
- creating templates
- listing all templates
- listing one template
- updating templates
- deleting templates

In other words, we need the ability to store templates and the ability to retrieve templates. These will all be basic operations to manage data of a specific kind into the system.

It is important to state that we will not manage them, but rather the CMS will be external.

There should be no limit to how many templates can be saved or retrieved.

We need to offer the option to filter saved replies when they are being listed. This filtering system should work as either:
- filter by name
- filter by parts of a name

We need to offer the ability to store JSON that is source specific and then send it to conversations, if possible.

paulodiniz:
Fixes #675

### DIFF
--- a/backend/api/admin/src/main/java/co/airy/core/api/admin/TemplatesController.java
+++ b/backend/api/admin/src/main/java/co/airy/core/api/admin/TemplatesController.java
@@ -1,0 +1,104 @@
+package co.airy.core.api.admin;
+
+import co.airy.avro.communication.Template;
+import co.airy.core.api.admin.payload.CreateTemplateRequestPayload;
+import co.airy.core.api.admin.payload.DeleteTemplateRequestPayload;
+import co.airy.core.api.admin.payload.GetTemplateRequestPayload;
+import co.airy.core.api.admin.payload.ListTemplatesRequestPayload;
+import co.airy.core.api.admin.payload.TemplatesResponsePayload;
+import co.airy.core.api.admin.payload.UpdateTemplateRequestPayload;
+import co.airy.model.template.TemplatePayload;
+import co.airy.spring.web.payload.EmptyResponsePayload;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RestController;
+
+import javax.validation.Valid;
+import java.util.List;
+import java.util.UUID;
+import java.util.concurrent.ExecutionException;
+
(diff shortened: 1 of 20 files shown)


paulodiniz:
Points to discuss
-`/templates.list` is not paginated
-`/templates.list` filters by name in memory, and not by a lucene query as we wrote on the docs PR

If we decide to go with these 2 approaches, we also need to update the docs PR https://github.com/airyhq/airy/pull/872

backend/avro/communication/template.avsc
```diff
@@ -0,0 +1,11 @@
+{
+  "namespace": "co.airy.avro.communication",
+  "name": "Template",
+  "type": "record",
+  "fields": [
+    {"name": "id", "type": "string"},
+    {"name": "name", "type": "string"},
+    {"name": "content", "type": "string"},
+    {"name": "variables", "type": "string"}
```
paulodiniz, line 9: I'm thinking if there's a better way to store `variables`

backend/model/template/src/main/java/co/airy/model/template/TemplatePayload.java
```diff
@@ -0,0 +1,37 @@
...
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class TemplatePayload {
+    private String id;
+    private String name;
+    private String content;
+    private JsonNode variables;
+    public static TemplatePayload fromTemplate(final Template template) {
+        final ObjectMapper mapper = new ObjectMapper();
```
paulodiniz, line 23: Maybe we can use the objectmapper we. already have

chrismatix: Hmm, I wouldn't. The templates are source specific message content. So we want to pass them as unmodified as possible, right?

backend/api/admin/src/main/java/co/airy/core/api/admin/TemplatesController.java
```diff
@@ -0,0 +1,104 @@
...
+            return ResponseEntity.notFound().build();
+        }
+        stores.deleteTemplate(template);
+        return ResponseEntity.ok(new EmptyResponsePayload());
+    }
+
+    @PostMapping("/templates.list")
+    ResponseEntity<?> listTemplates(@RequestBody @Valid ListTemplatesRequestPayload payload) {
+        final List<TemplatePayload> response = stores.getTemplates().stream()
+                .filter(t -> t.getName().contains(payload.getName()))
```
chrismatix, line 78: Will this work, if the user does not pass the filter? 🤔

chrismatix (approved):
LGTM

backend/model/template/src/main/java/co/airy/model/template/TemplatePayload.java
paulodiniz: Good point 👍

chrismatix (approved):

PR #962, opened 2021-02-12
[#671] Render Rich Card (Google Style) on Live…

### LINKED ISSUE #671: Render Rich Card (Google Style) on Live Chat Plugin
### Media template with buttons

- The media template should be designed on the UI to have at least one image and one button
- The button link could be an external link or a trigger for a suggested reply
- Title and description are optional fields, so the UI should not break in case these fields were missing

image

bitboxer:
…Chat Plugin

closes #671


frontend/chat-plugin/src/App.module.scss
```diff
@@ -5,7 +5,7 @@
   right: 0;
   bottom: 0;
   max-height: 500px;
-  max-width: 300px;
+  max-width: 320px;
```
bitboxer, line 8: I hope it is okay to make the chat plugin 20px wider. The google rich card needs that space. Otherwise I can also tweak the rich cards, but this was the easiest fix 😉

bazel.tsconfig.json
```diff
@@ -35,10 +35,16 @@
         "./lib/typescript/types/*"
       ],
       "dates": [
-        "./lib/typescript/types"
+        "./lib/typescript/dates"
```
bitboxer, line 38: Fixed a small copy and paste error here

AitorAlgorta (dismissed):
Awesome stuff!

bitboxer:
Will wait for @chrismatix to check the API design and then merge it.

lib/typescript/types/content.ts
```diff
@@ -32,8 +32,13 @@ export interface Text {
   text: string;
 }
 
+export interface SuggestionResponse {
```
chrismatix, line 35: Sorry: I forgot to delete this file. It only really makes sense to be on `types` if it's a universal content model, but since we don't have that anymore I would add this type directly to the code where it's needed

lib/typescript/render/shared.ts
```diff
@@ -2,19 +2,35 @@ import {isFromContact, Message, Contact} from 'httpclient';
 import {formatTime} from 'dates';
 import {DefaultMessageRenderingProps} from './components';
 
+export interface Command {
+  type: string;
+}
+
+export interface SuggestedReplyCommand extends Command {
```
chrismatix, line 9: Nice, this is how I imagined it. However don't you think that `SuggestedReply` is a bit too source specific? :) Perhaps we can break it down to "send message" with a generic payload.

frontend/chat-plugin/src/components/websocket/index.ts
```diff
@@ -1,7 +1,7 @@
 import {Client, messageCallbackType, IFrame} from '@stomp/stompjs';
 import 'regenerator-runtime/runtime';
 import {start, getResumeToken, sendMessage} from '../api';
-import {Text} from 'types';
+import {SuggestionResponse, Text} from 'types';
```
chrismatix, line 4: See my comment on the `content.ts` file: We should move any types to the chatplugin itself

chrismatix (commented):
Already looking good! Started some minor discussions

lib/typescript/render/shared.ts
bitboxer: I think I will keep it this specific till we find another usecase, then we can open it up to be more generic if we understand more about the data 😉

chrismatix (dismissed):

chrismatix (approved):

Issue #934, opened 2021-02-10
Parse (but don't map) API message response content


chrismatix:
**Current state**

We deliver in-/outbound message content on the APIs as a string, which is the way we store it internally. I.e. a text message looks like this:

```json5
{ //...
    "content": "{\"text\":\"on the road again\"}",
}
```

**Desired state**

We still do not want to maintain a schema for the source content. However, it would be useful to clients if we at least parse the message content to an untyped JSON object to make it easier to consume. The same message would look like this:

```json5
{ //...
    "content": {"text":"on the road again"},
}
```

This works for all sources that support JSON. For Twilio, this is not the case as we get messages in a url-encoded string. Therefore we can keep representing message content that we cannot parse to JSON as a string like so:

```json5
{
  "content": "ApiVersion=2010-04-01&Body=hello%20how%20are%20you%0A"
}
```

TODO:
- [ ] If possible parse the content for the `MessageResponsePayload` to a `JsonNode`
- [ ] Remove `JSON.parse` code from the render library

bitboxer:
It would be really nice if you could parse the string and extract a json object out of it 🙏

chrismatix:
@bitboxer Yes, that's the spirit of the ticket.

lucapette:
What do we do with twilio?

AitorAlgorta:
I think we need a special renderer for Twilio

lucapette:
that's not my point. My point is that the API looses consistency:

- it's a JSON object
- unless it's twilio 

That's _really_ strange.

What's the advantage of this not being always a string? Which, just to be sure, it's by far the simplest approach and makes the most consistent api

chrismatix:
Yes, Twilio is "funny" because inbound is URL encoded while outbound is json. So the rendered already has to make a distinction.

@lucapette makes a good point, which is why I initially went for a String in the payload. But I also get cc @steffh point that having a JSON for most cases makes the content look a bit more natural for heavy users of sources that support JSON.

paulodiniz:
I think I get what @lucapette is saying. It's a principle of least surprise.

Meaning: it's a surprise that the API returns one type and then returns another type depending on the source. I agree with that.
We *could* transform the twilio data to a json before delivering it, but that would break the whole principle of it (not change content)

bitboxer:
[...]

PR #977, opened 2021-02-15
[#934] Parse (but don't map) API message response…

paulodiniz:
…content

### DIFF
--- a/backend/model/message/src/main/java/co/airy/model/message/MessageRepository.java
+++ b/backend/model/message/src/main/java/co/airy/model/message/MessageRepository.java
@@ -3,9 +3,12 @@ package co.airy.model.message;
 import co.airy.avro.communication.DeliveryState;
 import co.airy.avro.communication.Message;
 import co.airy.model.metadata.dto.MetadataMap;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
 
 import java.time.Instant;
-import java.util.Map;
+import java.util.List;
 
 public class MessageRepository {
     public static Message updateDeliveryState(Message message, DeliveryState state) {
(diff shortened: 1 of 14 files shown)


backend/model/message/src/main/java/co/airy/model/message/MessageRepository.java
```diff
@@ -29,5 +37,18 @@ public static String resolveContent(Message message, Map<String, String> metadat
...
+
+        try {
+            jsonNode = new ObjectMapper().readTree(resolvedContent);
+        } catch (JsonProcessingException e) {
+            ObjectMapper objectMapper = new ObjectMapper();
+            try {
+                jsonNode = objectMapper.readTree("{\"text\": \"" + content + "\", \"type\": \"text\"}");
+            } catch (JsonProcessingException ex) {
+                ex.printStackTrace();
+            }
```
chrismatix, line 49: In the ticket we agreed on returning the content as is if we cannot parse it to a string so I would simply `return resolvedContent;` in the `catch` clause.

chrismatix (commented):
The Typescript and doc changes are good 👍

backend/model/message/src/main/java/co/airy/model/message/MessageRepository.java
paulodiniz: We cannot because the return type of `content` is `JsonNode`

paulodiniz: This discussion is now on https://github.com/airyhq/airy/issues/934#issuecomment-779759231

frontend/ui/src/pages/Inbox/ConversationListItem/index.tsx
```diff
@@ -38,8 +38,7 @@ const connector = connect(mapStateToProps, mapDispatchToProps);
 
 const FormattedMessage = ({message}: FormattedMessageProps) => {
   if (message && message.content) {
-    const messageJSON = JSON.parse(message.content);
-    return <>{messageJSON.text}</>;
+    return <>{message.content.text}</>;
```
paulodiniz, line 41: This will break with Twilio because `message.content` is a String

chrismatix: I thought this was already using the render library to be honest.

chrismatix (approved):

PR #988, opened 2021-02-16
[#918] Consume Airy Events from new websocket in the UI

### LINKED ISSUE #918: Consume Airy Events from new websocket
#884 introduces a new Websocket that delivers `Airy Events` using a single queue. This means that the payload will have a `type` field, which can be used to identify and read it.

The task is to switch to that Websocket.

chrismatix:
resolves #918


backend/api/communication/src/main/java/co/airy/core/api/communication/Stores.java
```diff
@@ -109,6 +110,7 @@ private void startStream() {
 
         // produce unread count metadata
         messageStream.selectKey((messageId, message) -> message.getConversationId())
+                .filter((conversationId, message) -> isFromContact(message))
```
chrismatix, line 113: Fixes a bug wherein unread counts where increased for messages sent from Airy

.eslintrc
```diff
@@ -16,7 +16,7 @@
 
   "rules": {
     "react/prop-types": 0,
-    "@typescript-eslint/no-unused-vars": ["error", {"varsIgnorePattern": "_"}],
```
chrismatix, line 19: We should always provide a reason when changing the rules: This is very useful when you want to remove properties from an object by destructuring. I.e.

```typescript
const {name, ...dataWithOutName} = contact;
```
You don't need the name here, but need to state it explicitly

bitboxer (commented):
Really awesome refactoring ❤️ . The typescript stuff is 👍

bitboxer (approved):
:shipit:

paulodiniz (approved):

PR #991, opened 2021-02-16
Fix Netlify docs previews

chrismatix:
Currently `develop.docs.airy.co` and the branch previews do not work, because they docusaurus build has a static `baseUrl` set to `/docs/core`


ljupcovangelski (approved):
Nice :+1:

PR #1011, opened 2021-02-17
Feature/905 extract avatar and time from message render library

thorstenairy:


frontend/ui/src/pages/Inbox/Messenger/MessageList/index.tsx
```diff
@@ -165,12 +168,18 @@ const MessageList = (props: MessageListProps) => {
                   {formatDateOfMessage(message)}
                 </div>
               )}
-              <SourceMessage
-                source={getSource(conversation)}
-                message={message}
-                contact={contactToShow}
-                lastInGroup={lastInGroup}
-              />
+              <AvatarTime
```
bitboxer, line 171: Somehow this name does not feel right for me? Maybe someone here has a better suggestion?

lib/typescript/render/components/RichCard/index.tsx
```diff
@@ -51,27 +59,58 @@ export const RichCard = ({title, description, suggestions, media, cardWidth, com
...
-              key={idx}
-              className={styles.suggestionButton}
-              onClick={() => {
-                clickSuggestion(suggestion);
-              }}>
-              {suggestion.reply ? suggestion.reply.text : suggestion.action.text}
-            </button>
-          ))}
+    <>
+      {fromContact ? (
```
bitboxer, line 63: Maybe we could build it in a way that it's not longer needed to differentiate that stuff in the components?

bitboxer (commented):
I have the feeling that there is still stuff in the other renderers that could be removed? I also left some questions for the team to discuss maybe?

frontend/ui/src/pages/Inbox/Messenger/MessageList/index.tsx
thorstenairy: always up for other suggestions :)

bitboxer: `MessageWrapper` sounds super boring and generic...maybe `MessageRow` because this is what it actually should represent?

lib/typescript/render/components/RichCard/index.tsx
thorstenairy: Do you have any good solution for this already in mind? @bitboxer

bitboxer: the `div` surrounding the component could be a flex layout and use the flex attributes to render something on the left or right?

frontend/ui/src/pages/Inbox/Messenger/MessageList/index.tsx
```diff
@@ -165,12 +168,18 @@ const MessageList = (props: MessageListProps) => {
...
-              />
+              <MessageInfoWrapper
+                fromContact={isFromContact(message)}
+                contact={conversation.metadata.contact}
+                sentAt={sentAt}
+                lastInGroup={lastInGroup}>
+                <SourceMessage
+                  source={getSource(conversation)}
+                  message={message}
+                  contact={contactToShow}
```
chrismatix, line 179: Is the `contact` still needed for the `SourceMessage` after this PR? If not I would remove it here and from the `DefaultMessageRenderingProps` :)

lib/typescript/render/components/RichCard/index.tsx
chrismatix: I agree with @bitboxer: From the looks of it the only differentiator between the ternary branches is a `flex` attribute. So since everything is supposed to be wrapped in `MessageInfo` (good name btw 👍 ) it should go there so we can completely remove the `fromContact` dependency.

AitorAlgorta: then we need `MessageInfo` also in the chat plugin. Because the chat plugin also need to render messages in both sides.

bitboxer (dismissed):
Awesome stuff.

bitboxer (approved):

PR #1012, opened 2021-02-17
Docs/1001 improve google facebook sources

AudreyKj:
changes on FB and Google sources: 
- added more info in configuration of Fb and Google sources, fixed typo, replaced config FB image + added image for google (see below)

Screenshot 2021-02-17 at 11 49 18 2

Screenshot 2021-02-18 at 10 49 31


docs/docs/sources/facebook.md
```diff
@@ -99,4 +120,120 @@ On the `User or Page` option, select `Get Page Token` and click on `Generate Acc
...
+icon={() => <BoltSVG />}
+title='Channels endpoint'
+description='Connect a Facebook source to your Airy Core instance through the Channels endpoint'
+link='api/endpoints/channels#facebook'
+/>
+
+<br />
+
+```
+POST /channels.facebook.connect
```
chrismatix, line 145: If we want to duplicate the endpoint documentation here, I would add them as a `connect-facebook.mdx` component so that we make it easier to change in the future.

lucapette: Agreed.

To be fair we moved this away from the sources document on the suggestion of @steffh but I suppose then @sgarroum suggested we introduced this back?

I think what @chrismatix suggested is the best of both words as, in fact, both places are "right" for this kind of doc. 

@AudreyKj we do something along the line of this with twilio config so you can use that as a reference (do ping me if you run into any trouble)

sgarroum: Yes, I am highly in favor of duplicating it (@chrismatix unification approach makes sense) and thus making the source connection doc self sufficient - from the first connection command to testing it by sending and receiving a message.

lucapette (approved):
👏 👏 👏

docs/docs/api/endpoints/channels.md
```diff
@@ -134,73 +134,15 @@ POST /channels.chatplugin.connect
...
+<ConnectFacebook />
 
 ### Google
 
-Connects a Google Business Account to Airy Core.
-
-```
-POST /channels.google.connect
-```
+import ConnectGoogle from './connect-google.mdx'
```
pascal-airy, line 143: cool 😎

bitboxer (approved):
🥳

bitboxer (approved):

PR #1024, opened 2021-02-18
[#957] Add ErrorBoundaries for the RenderLibrary

### LINKED ISSUE #957: Add ErrorBoundaries for the RenderLibrary
We don't want unexpected behaviours in the render library, and if we have, we would like to have a backup plan for this and not crash the page or render something undesirable. So the best approach is using ErrorBoundaries and wrap the component in it, so we are sure that even in case of crash or error, we will display something that is under our control.

bitboxer:
Sadly I found cases where the try..catch failed. And it was not possible
to catch them  with the hooks approach. I needed to transform this to a
Component. See also [this faq](https://reactjs.org/docs/hooks-faq.html#do-hooks-cover-all-use-cases-for-classes).

closes #957

### DIFF
--- a/lib/typescript/render/index.tsx
+++ b/lib/typescript/render/index.tsx
@@ -6,13 +6,39 @@ import {getDefaultMessageRenderingProps, MessageRenderProps} from './shared';
 
 export * from './shared';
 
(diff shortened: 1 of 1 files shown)


AitorAlgorta (approved):
🚀

PR #1027, opened 2021-02-18
[#1007] Bug: Cookies + 2 Chat Plugins

### LINKED ISSUE #1007: Bug: Cookies + 2 Chat Plugins
Have two chat plugins - one called `testing6` and the other one `aditest`.

When I initialize `aditest`, it pulls up the previous conversation with `testing6` and adds the message to the other conversation.

screenshot_2021-02-16_at_18 31 02

**Possible problem:** might be the cookies -- they are too infringing and disregard the url parameters. Once I delete the cookie, it initializes the right plugin and creates a new conversation.

bitboxer:
I refactored/moved a few things:

* everything is now `camelCase` in the chat plugin
* things that are no react components are now moved out of the
  `components` directory
* added an eslint rule that warns if `snake_case` is used.

closes #1007

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,26 +1,52 @@
 {
-  "extends": ["eslint:recommended", "plugin:react/recommended", "plugin:@typescript-eslint/recommended"],
-  "plugins": ["react", "@typescript-eslint"],
+  "extends": [
+    "eslint:recommended",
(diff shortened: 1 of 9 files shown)


chrismatix (approved):
Good stuff 👍

PR #1053, opened 2021-02-22
Ensure that the Facebook token is set on the channel

chrismatix:


AitorAlgorta (dismissed):

backend/sources/google/connector/src/main/java/co/airy/core/sources/google/services/Mapper.java
```diff
@@ -15,6 +16,14 @@ public JsonNode fromSendMessageRequest(SendMessageRequest sendMessageRequest) th
         final Message message = sendMessageRequest.getMessage();
         final JsonNode messageNode = mapper.readTree(message.getContent());
         ((ObjectNode) messageNode).put("messageId", message.getId());
+
+        // Default human representative
+        if (messageNode.get("representative") == null) {
```
AitorAlgorta, line 21: I would actually remove this if we are going full "transparent" with what we send.

lucapette (approved):

AitorAlgorta (approved):

Issue #1017, opened 2021-02-17
Introduce the airy create command


ljupcovangelski:
This issue is about introducing the `airy create` command. At first we ignore how the Kubernetes cluster is created and we are only expecting the path of the `kubeconfig`. So the command should roughly look like this

```
airy create --kubeconfig=./kube.conf
```
The default location for the `kubeconfig` should be `~/.airy/kube.conf`, which is now created by the `./scripts/bootstrap.sh`.

We will further discuss the flags later in the comments to the issue.

In order to accomplish this, the whole bash provisioning needs to be rewritten. Here are the required steps and the other relqted issues:

- [x] Provision the kafka topics with a kubernetes job #1016 
- [x] Deploy the kubernetes resources from helm templates #1014
- [x] Create traefik ingress controller in the kubernetes cluster
- [x] Change the hostnames for core #1022 
- [x] Apply scheduling and waiting for resources to be there #1015 
- [ ] Print out messages about how to access and start using `Airy Core`
- [x] Introduce the minikube provider #1041 

After this issue is finished, we can proceed with how do we create the kubernetes cluster, both locally and in the cloud.

_(The original issue was intended for a different purpose. The description was added after the first round of discussions)._

lucapette:
from #1020 

After the successful creation of Airy Core, we need to print-out the output for how to access the platform.

 Currently we print out:

    The webhooks
    The API endpoint
    Info about how to create the first user.


Definitely part of the requirement here is defining the output. But, to be fair, I think we should start from the way the command UI looks like. Here's some early questions:

- We can do providers in two ways (at least I can think of two but probably there's more):
   - `airy create --provider=local --args etc...`   
   - `airy create local --args etc..`
   When I thought about this, I assumed I'd like the former more but now that I see them written down I think the latter is better: it's more explicit, no need for funky work on args that are provider dependent, each command gets its own docs/description.
- What does the command exactly do? From what I'm reading in #1015 I think there's some confusion about this and lack of clarity (I didn't even think we'd have an airy start command at first as I personally assumed airy create would do the exact same thing that we have now with the bootstrap. after all that's the main goal here: get rid of the bootstrap script in a way we can have multiple providers much more easily)
- What's the output? I think we should define this a little upfront. It will also clarify in detail: the steps we have to build, the different responsibilities of each step. I will propose something about this later today. I want to think bout it for a minute and the previous questions are more important

bitboxer:
[...]

Issue #1051, opened 2021-02-22
Remove airy init 


lucapette:
As `airy create` will effectively do all the work `airy init` does, the command is redundant from a UX perspective. 

We use this issue as an "excuse" to also restructure the CLI looks for config. Now we have different flags for different things (actual cli config, kube.conf.... airy.yaml file). I suggest we let users only choose a `config-dir` which will contain all the configuration of an airy core instance. 

That way we can also add some nice docs that suggest how to deal with multiple instances as you'd only need to point your cli to different directories and "everything just works" then (proposing some aliases in the docs would make this very nice to work with)

In the context of this change, we can also restructure the integration tests of the CLI a little:

- the runner generates a temp dir
- we put "golden files" in the temp dir
- we point the cli to this temp dir
- run all the tests in a single integration file 
- delete the temp dir

After the conversation in the comments, here's what we plan to do:

- [ ] introduce the concept of `configdir` while removing the file config flags (we point the CLI to a dir with `airy.yaml`, `kube.conf` `cli.yaml`)
- [ ] Because of the previous point, we refactor the integration tests as described (look up in this description)
- [x] move cli to the root directory (#1401)
- [ ] re-organise the packages internally (this point can be further discussed at pr time I think)
- [ ] remove the init command
- [ ] add the --init flag to the create command (so we cover the case @ljupcovangelski brought up)
- [ ] Update all the docs :)

ljupcovangelski:
I agree, however I think we must take into consideration that there is an already created `Airy core` and somebody is downloading the cli for the first time on a new machine, to interact with this existing `Airy core` instance.
But this can be made part of the `airy config` workflow, to add configuration for an already created instance.

lucapette:
that's such a good point! I didn't think about it. But I feel like we're going toward the right direction anyway:

- we remove `airy init` anyway
- `airy create --name instance` creates a dir `instance` with the config files according to the creation process the user choose
- we add `airy create --init instance` which only does the scaffolding of the dir part (in case you need to attach the cli to an existing instance you have the config for already somewhere else)

That should cover all the use cases. There's one more thing I would probably add to this process since we're at it. The CLI accepts a dir config as input or defaults to `./` (instead of `$HOME/.airy`). If we do that, one can achieve a very nice workflow where you have things like:

- `cd staging-instance; airy whatever # runs commands in staging`
- `cd prod-instance; airy whatever # runs commands in prod`
- `airy --config-dir foo whatever # runs commands in foo`

[...]

PR #1056, opened 2021-02-22
[#1017] Introduce the airy create command

paulodiniz:


infrastructure/cli/cmd/create/create.go
```diff
@@ -0,0 +1,31 @@
...
+	"fmt"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+var (
+	provider  string
+	CreateCmd = &cobra.Command{
+		Use:   "create",
+		Short: "Creates an instance of the airy core",
```
lucapette, line 13:
```suggestion
		Short: "Creates an instance of Airy Core",
```

lucapette (approved):
I'm not sure this "fixes the issue" :D but I think we should merge already as there's nothing wrong with this :)

infrastructure/cli/cmd/create/create.go
```diff
@@ -0,0 +1,32 @@
...
+
+func init() {
+	CreateCmd.Flags().StringVarP(&provider, "provider", "", "", "One of the supported providers (aws|local)")
+	CreateCmd.MarkFlagRequired("provider")
+	viper.BindPFlag("provider", CreateCmd.PersistentFlags().Lookup("provider"))
+}
+
+func create(cmd *cobra.Command, args []string) {
+	fmt.Println("⚙️  Creating core with provider", provider)
+	fmt.Println("🚀 Starting core with default config")
```
ljupcovangelski, line 27: Not sure about default config. I would recommend: `Starting core with default components`, as we don't actually provide any configuration for the components by default (the values which are in `airy.yaml`), apart from what is on or off.

docs/docs/cli/reference.md
```diff
@@ -81,6 +81,31 @@ airy config apply [flags]
...
+
+```
+  -h, --help              help for create
+      --provider string   One of the supported providers (aws|local)
+```
+
+#### Options inherited from parent commands
+
+```
+      --apihost string      Airy Core HTTP API host (default "http://api.airy")
```
chrismatix, line 104: I'd also split this into `api-host` to keep it consistent

lucapette: I wouldn't do this here. We're reworking the whole config in #1051 and that's the best place for this

lucapette (approved):

ljupcovangelski (approved):

PR #1076, opened 2021-02-24
[#814] Add cmd+enter and better disabled state to…

### LINKED ISSUE #814: Add cmd+enter and better disabled state to input bar
We want to enhance the input bar of the demo app a little bit. For that it:  

- [x] sends a message when pressing ⌘ + ↵ on a mac or the equivalent for windows.
- [x] if the message input field is empty, completely disable the send button. On hover it should stay gray and the cursor should not indicate a clickable field. 
- [x] if the message input has content, enable the send button, make it blue to hint that it can be clicked now

bitboxer:
…input bar

closes #814

### DIFF
--- a/frontend/ui/src/pages/Inbox/MessageInput/index.tsx
+++ b/frontend/ui/src/pages/Inbox/MessageInput/index.tsx
@@ -1,4 +1,4 @@
-import React, {useState, useEffect, useRef} from 'react';
+import React, {useState, useEffect, useRef, KeyboardEvent} from 'react';
 import {connect, ConnectedProps} from 'react-redux';
 import {useParams} from 'react-router-dom';
 import styles from './index.module.scss';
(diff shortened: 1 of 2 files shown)


frontend/ui/src/pages/Inbox/MessageInput/index.tsx
```diff
@@ -36,10 +36,20 @@ const MessageInput = (props: MessageInputProps & ConnectedProps<typeof connector
   const conversationIdParams = useParams();
   const currentConversationId: string = conversationIdParams[Object.keys(conversationIdParams)[0]];
 
-  const handleClick = () => {
+  const sendMessage = () => {
     props.sendMessages(getTextMessagePayload(channelSource, currentConversationId, input)).then(() => setInput(''));
   };
 
+  const handleClick = () => {
```
chrismatix, line 43: nitpick: What's the point of the extra wrap? :)

chrismatix (approved):

AitorAlgorta (approved):

frontend/ui/src/pages/Inbox/MessageInput/index.tsx
bitboxer: Could have called the method down below, but we will add more stuff to it soon. Would have to introduce the method again.

Issue #1041, opened 2021-02-19
Introduce minikube provider for airy create 


lucapette:
We require minikube to be present on the machine already. We make it as explicit as possible in the docs and provide suggestions for CPU/RAM.

- [x] provision Kubernetes cluster
- [x] store path to the kubeconfig file and kubecontext 
- [x] provision endpoint
- [ ] remove vagrant bootstrap code
- [ ] Update docs

lucapette:
About minikube, I'm not sure what the advantage of specifically require minikube would be. But if there's any, all up for it as it'a lovely project. For sure, we should recommend people trying this locally to install kubernetes via minikube (probably with some specific requirements for cpu/ram).

I think `local` is somewhat a strange name, so here's what I'm thinking:

- if we require minikube (maybe this is a good excuse to do so) I would call it minikube
- k8s may be another name... but maybe it's a little too generic (and I'm not sure we should have this door open via the airy create command of "installing it in a bare kubernetes cluster". Being in control of the env, via the face we know the provider, can simplify how you'd interact with your airy core instance).

And now after I wrote this, I'm leaning toward 1) calling it minikube 2) require minikube 3) leverage minikube commands to expose the airy instance locally

lucapette:
One more argument for making minikube required (and thefore calling the provider minikube):  we can offer a better development experience by using minikube locally as it's somewhat more comfortable to create a container registry and therefore providing a better workflow for developers (change something, compile, ship to minikube). Obviously this is out of the scope of the provider itself but it'd help us long term

chrismatix:
I agree with the points mentioned above: So we go with `--provider=minikube` and fail if it's not installed.

paulodiniz:
👍 On calling it `minikube`

@lucapette When you say we `require` minikube it means that we don't bundle it in the cli correct? (I like that idea)

If we agree on that the checklist should be:

- [ ] Change the provider name to `minikube`
- [ ] Update the docs to convey that minikube is a dependency that should be installed beforehand
- [ ] Spit out minikube specific commands

lucapette:
@chrismatix @paulodiniz  I would agree with that. We write the docs of the provider so that we explain people how to install minikube (also helpful for us so we can communicate basic requirements CPU/RAM to actually have a decent experience with core locally).

I will update the issue now 🎉

ljupcovangelski:
I agree that `minikube` is a good option for the local provider. I don't quite understand which option are we gonna go for:
[...]

PR #1084, opened 2021-02-25
[#1014] Deploy the helm charts

### LINKED ISSUE #1014: Deploy the helm charts
As we have the helm charts templates in our repo (for a particular version, bundled in the CLI), we would want to generate all the manifests with cli and apply them to the Kubernetes cluster.

This needs to be part of the `airy create` command and we would need to use some `helm` libraries, so that we don't depend on the `helm` binary.

paulodiniz:
resolves #1014


infrastructure/cli/cmd/create/create.go
```diff
@@ -4,25 +4,51 @@ import (
...
 
+const defaultKubeConfig = "$HOME/.kube/config"
+
 func init() {
+	if home := homedir.HomeDir(); home != "" {
+		CreateCmd.Flags().StringVar(&kubeConfig, "kubeconfig", defaultKubeConfig, "(optional) absolute path to the kubeconfig file")
+	} else {
+		CreateCmd.Flags().StringVar(&kubeConfig, "kubeconfig", "", "absolute path to the kubeconfig file")
+	}
+
```
lucapette, line 32: we should hide this flag from the public UI

infrastructure/cli/cmd/create/create.go
```diff
@@ -4,25 +4,51 @@ import (
...
 func create(cmd *cobra.Command, args []string) {
 	fmt.Println("⚙️  Creating core with provider", provider)
+
+	if kubeConfig == defaultKubeConfig {
+		kubeConfig = filepath.Join(homedir.HomeDir(), ".kube", "config")
+	}
+
+	helm := New(kubeConfig, "develop", "default")
+	if err := helm.Setup(); err != nil {
+		log.Fatalf("setting up Helm failed with err %v", err)
```
lucapette, line 47: we don't use log anywhere as it comes with its own format and right now we don't have an opinion about it. So Printf plus exit is how we're doing this (not sure we don't have any log left... I did remove the few we had a few weeks ago but things move fast :))

infrastructure/cli/cmd/create/helm.go
```diff
@@ -0,0 +1,218 @@
...
+	name      string
+	version   string
+	namespace string
+	clientset *kubernetes.Clientset
+}
+
+func New(kubeConfigPath string, version string, namespace string) Helm {
+	config, err := clientcmd.BuildConfigFromFlags("", kubeConfigPath)
+	if err != nil {
+		log.Fatal(err)
```
lucapette, line 37: also here (in the whole file) same comment about logs

infrastructure/cli/cmd/create/helm.go
```diff
@@ -0,0 +1,218 @@
...
+
+const airyYamlConfigMap = "airy-yaml-config"
+const serviceAccountName = "helm-account"
+
+const defaultAiryConfig = `
+global:
+  appImageTag: develop
+  containerRegistry: ghcr.io/airyhq
+  namespace: default
+`
```
lucapette, line 25: I get it was a little easier to do it this way but I'm not so comfortable making such an important concept just a string somewhere. As I understand it, it's also not so much work as now go supports embededding data into a binary natively (long overdue feature, https://golang.org/pkg/embed/ this should get us going).

I suggest we introduce `infrastructure/defaults.yml` (we can debate our guidelines about the configuration in a different pr, probably the best chance is doing that right before we eliminate the init command) and embed the file here

infrastructure/cli/cmd/create/helm.go
```diff
@@ -0,0 +1,218 @@
...
+	_, err = roleBindingClient.Create(context.TODO(), roleBinding, metav1.CreateOptions{})
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (h *Helm) InstallCharts() error {
+	return h.runHelm([]string{"install", "--values", "/apps/config/airy.yaml", "core", "/apps/helm-chart/"})
```
lucapette, line 94: this airy.yaml is not "our official" config file right? Because if it is, it's confusing, if it isn't then the name is confusing :sorry:

infrastructure/cli/cmd/create/helm.go
```diff
@@ -0,0 +1,218 @@
...
+		_, err := h.clientset.CoreV1().ConfigMaps(h.namespace).Create(context.TODO(),
+			&corev1.ConfigMap{
+				ObjectMeta: v1.ObjectMeta{
+					Name:      airyYamlConfigMap,
+					Namespace: h.namespace,
+				},
+				Data: cmData,
+			}, v1.CreateOptions{})
+		return err
+	} else {
```
lucapette, line 200: we should be able to get by without the else (not the code, just the actual branching)

lucapette (changes requested):
Very solid work!

I'm requesting changes because of the UI flag but I admit I have a somewhat strong opinion also about the embedding of the configuration

infrastructure/cli/cmd/create/helm.go
chrismatix: We tried to go that route briefly in our pairing session, but noticed a few pitfalls right away, because of Gazelle. I would recommend that we tackle this in [this issue](https://github.com/airyhq/airy/issues/1051#issuecomment-788041906), which can serve as a follow-up where we also embed an entire directory and not just the Airy yaml

infrastructure/cli/cmd/create/helm.go
chrismatix: What do you mean by "official" config file? 🤔 Currently, we use the `airy.yaml` as the helm values file so it felt naturally to call the mount the same.

infrastructure/cli/cmd/create/helm.go
lucapette: while I understand the argument and we'd like to move forward with this:

- it's not ready anyway
- before we get to that issue it's going to be a while (as we have two providers between this and the command) 

And I'm afraid it's too much of an important concept to hide it even for a week or so. to be fair the argument "we tried and it seems tricky" works more in favor of doing it right away than waiting the whole codebase has more code.  

Now having said all of it, my opinion is not so strong I'd prevent us fro merging this (that's something I tend to do only with the "public api" and the docs). So if the argument is not convincing, happy to add one more step to the issue you mention

infrastructure/cli/cmd/create/helm.go
lucapette: we're embedding the default config in the CLI so I'm not exactly sure what's this file for

chrismatix: I see: This is that default file and we mount it in the job's container using a config map. So I used `airy.yaml` as a generic name for that config. It's not the one from the infrastructure directory or any other file.

infrastructure/cli/cmd/create/helm.go
paulodiniz: I don't think we can. Creating a config map with the same name will yield an error. I'm probably not getting this correctly

infrastructure/cli/cmd/create/helm.go
paulodiniz: I changed the code so it uses a `infrastructure/defaults.yaml` as @lucapette mentioned

infrastructure/cli/cmd/create/helm.go
```diff
@@ -0,0 +1,221 @@
...
+				Name:      airyConfigMap,
+				Namespace: h.namespace,
+			},
+			Data: cmData,
+		}, v1.CreateOptions{})
+	return err
+}
+
+func (h *Helm) getAiryConfigMap() string {
+	content, err := ioutil.ReadFile("./infrastructure/defaults.yaml")
```
chrismatix, line 204: I am not sure if this is going to work if you run the CLI from anywhere else than the repository. really think we need to embed this file for this to work.

paulodiniz: Can we come up with a test scenario for this? Or: how can we test this?

chrismatix: You could build and copy the binary outside of the repository workspace and try running the `airy create` from there.

infrastructure/scripts/provision/core.sh
```diff
@@ -14,14 +14,9 @@ wait-for-service-account
...
-    airy init
-    airy config apply --kube-config /etc/rancher/k3s/k3s.yaml --config "${INFRASTRUCTURE_PATH}"/airy.yaml
-else
-    helm install core "${INFRASTRUCTURE_PATH}"/helm-chart/ --set global.appImageTag="${AIRY_VERSION}" --set global.ngrokEnabled="${NGROK_ENABLED}" --timeout 1000s > /dev/null 2>&1
-fi
+wget -qnv https://airy-core-binaries.s3.amazonaws.com/"${AIRY_VERSION}"/linux/amd64/airy
+chmod +x airy
+mv airy /usr/local/bin/
+airy init
+airy create --provider=local
```
ljupcovangelski, line 21: I thought this will be something like
```
airy create --kubeconfig /etc/rancher/k3s/k3s.yaml
```
And the helm charts are deployed.

I think we agreed on building the first version with the kubeconfig file there and k3s already setup, but I might be wrong.

chrismatix: Yes that's correct, the flag is missing from this line. But all this will go away of course after https://github.com/airyhq/airy/issues/1041

ljupcovangelski: Thank you :+1: 
Maybe we don't even need the `--provider` flag for now.

chrismatix (approved):

lucapette (approved):

Issue #1047, opened 2021-02-22
Automated test of the chatplugin


pascal-airy:
The chat plugin is currently the only source that we control so we will start writing integration tests for it first.

The test should connect a chat plugin source and write messages from both sides also using Cypress for the UI testing and receiving of messages.

PR #1087, opened 2021-02-25
[#1047] Chatplugin testing

pascal-airy:


thorstenairy (dismissed):
LGTM

integration/chatplugin/create_conversation.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+      cy.visit('/login');
+      cy.get('form')
+        .within(() => {
+          cy.get('input[type=email]').type(Cypress.env('username'));
+          cy.get('input[type=password]').type(Cypress.env('password'));
+        })
+        .submit();
+      cy.get('[data-cy=messageTextArea]').type(messageInbox);
+      cy.get('[data-cy=messageSendButton]').click();
+      cy.get('[data-cy=messageList]').children().its('length').should('be.gte', 2);
```
chrismatix, line 27: This is nice. I have a suggestion to make it even better: In the `frontend/ui` package you could add a single file library that exports all these handles like so:

```typescript
export const cyMessageList = "messageList";
```
```python
ts_library(
name = "handles",
srcs = ["handles.ts"]
)
```

and then here and in the UI code you can reference it by importing 
```typescript
import {cyMessageList} from "handles"
```

This way we can check at build time that the names are in sync

lucapette: I was about to suggest this :) 💯

integration/ui/BUILD
```diff
@@ -1,11 +1,10 @@
...
-load("@npm//@bazel/typescript:index.bzl", "ts_library")
+load("@com_github_airyhq_bazel_tools//web:typescript.bzl", "ts_web_library")
 
-ts_library(
+ts_web_library(
     name = "specs",
     srcs = glob(["*.spec.ts"]),
-    tsconfig = ":tsconfig.json",
     deps = [
+        "//frontend/ui:app",
```
pascal-airy, line 7: This should not depend on the app target but on the `:handler` but then the build fails. I think because the handler.js is picked up by both rules. Not sure how to fix that. Please help

frontend/chat-plugin/src/components/chat/index.tsx
```diff
@@ -119,11 +120,11 @@ const Chat = (props: Props) => {
 
   const inputBar = props.inputBarProp
     ? () => props.inputBarProp(ctrl)
-    : () => <AiryInputBar sendMessage={sendMessage} />;
+    : () => <AiryInputBar sendMessage={sendMessage} cyTextareaId={cyInputbarTextarea} cyButtonId={cyInputbarButton} />;
```
bitboxer, line 123: Why not `dataCyTextareaId` ?

frontend/chat-plugin/src/airyRenderProps/AiryInputBar/index.tsx
```diff
@@ -4,10 +4,14 @@ import style from './index.module.scss';
 
 type Props = {
   sendMessage: (text: string) => void;
+  cyButtonId?: string;
+  cyTextareaId?: string;
 };
 
 const AiryInputBar = (props: Props) => {
   const [messageString, setMessageString] = useState('');
+  const cyButtonId = props.cyButtonId;
```
chrismatix: Since these are global variables, I would not pass them via props, but require them right here

frontend/ui/tsconfig.json
```diff
@@ -0,0 +1 @@
+{}
```
chrismatix, line 1: Remove file

integration/chat-plugin/create_conversation.spec.ts
```diff
@@ -0,0 +1,19 @@
+import {cyBubble, cyInputbarTextarea, cyInputbarButton} from '../../frontend/chat-plugin/handles';
```
chrismatix, line 1: Should be `from 'handles'` since we are treating it as its own package.

integration/ui/BUILD
chrismatix: So the idea was to do a separate `ts_library` only for the labels that gets used as an import in `//frontend/ui:app` and here.

pascal-airy: yes! I had this but I got a weird error that the handles.js is targeted twice from the :app and the :handles target
https://github.com/airyhq/airy/pull/1087/commits/dd87b9be16e460188de45118fd18ab4c834d8217

frontend/chat-plugin/src/airyRenderProps/AiryInputBar/index.tsx
pascal-airy: fair

integration/ui/BUILD
chrismatix: Yes, then you have to exlcude the `handles.ts` in the source glob pattern of the `:app` target :)

chrismatix: Or move it to a completely different module

pascal-airy:
@chrismatix hope you like my Hendles
image

chrismatix (approved):
Good job with all the handles 😄

lucapette (approved):

PR #1098, opened 2021-02-26
replaced manual mapping with npm lib

AudreyKj:
resolves #1098 

changes: 
- replaced mappers with npm library ([camelcase-keys](https://www.npmjs.com/package/camelcase-keys)), mappers are completely deleted from the http library
- affects: http library + websocket library + chat plugin


lib/typescript/httpclient/index.ts
```diff
@@ -16,21 +16,12 @@ import {
...
-  userMapper,
-  paginatedPayloadMapper,
-  messageMapper,
-  conversationMapper,
-} from './mappers';
-import {TagColor, Tag} from './model';
+
+import {TagColor, Tag, Message} from './model';
+/* eslint-disable @typescript-eslint/no-var-requires */
+const camelcaseKeys = require('camelcase-keys');
```
AudreyKj, line 24: here we have to use _require_ and not _import__ because the library supports TS but is not in TS

lib/typescript/httpclient/index.ts
```diff
@@ -97,30 +88,38 @@ export class HttpClient {
     return this.parseBody(response);
   }
 
+  private mapMessage = (payload: MessagePayload): Message => {
+    return {...camelcaseKeys(payload, {deep: true}), sentAt: new Date(payload.sent_at)};
+  };
+
```
AudreyKj, line 94: I added _mapMessage_  as a private method because it is used x3 in the library

chrismatix (approved):
Awesome! 💥

chrismatix (changes requested):
So this is good already, but there are a few places where we should not map the keys and use the libraries [stop path](https://www.npmjs.com/package/camelcase-keys#stoppaths) feature:
- On every payload that has a `metadata.userData` path because that is user-supplied
- Every `message.content` field, because that schema is source specific

frontend/ui/src/pages/Inbox/ConversationsFilter/index.tsx
```diff
@@ -111,7 +111,7 @@ const ConversationsFilter = (props: ConversationsFilterProps) => {
     if (filteredPaginationData.loaded) {
       return (
         <div className={styles.filterCount}>
-          {`Filtered: ${filteredPaginationData.filtered_total} Total: ${filteredPaginationData.total}`}
+          {`Filtered: ${filteredPaginationData.filteredTotal} Total: ${filteredPaginationData.total}`}
```
AudreyKj, line 114: fixed this snake case in the UI, I am not so sure why this was not in camelCase...

lib/typescript/httpclient/payload/PaginatedResponse.ts
```diff
@@ -1,4 +1,4 @@
 export interface PaginatedResponse<T> {
   data: T[];
-  paginationData: {previousCursor: string; nextCursor: string; total: number};
+  paginationData: {previousCursor: string; nextCursor: string; total: number; filteredTotal: number};
```
AudreyKj, line 3: updated PaginatedResponse type with _filteredTotal_ field

frontend/ui/src/reducers/data/conversations/index.ts
```diff
@@ -29,7 +29,7 @@ export type AllConversationPaginationData = {
   total: number;
   loading?: boolean;
   loaded?: boolean;
-  filtered_total?: number;
+  filteredTotal?: number;
 };
```
AudreyKj, line 33: fixed snake_case in reducer for conversation data

chrismatix (approved):
Great work 👍

PR #1123, opened 2021-03-03
[#868] Templates manager

### LINKED ISSUE #868: Template UX
### Template UX

Although templates will be managed with a CMS externally, we should be able to render them within Airy Core. Towards the righthand side of the text input bar, we should include a template modal. We should display all authorized templates within this modal so that when a system user wants to view or select a template, the templates are readily available.

Screenshot 2021-02-03 at 16 32 30
Screenshot 2021-02-03 at 16 32 56

### Template Search Endpoint

We need to provide this endpoint to allow the search of authorized templates. This search should not be limited to a certain amount of templates - meaning that this search must be conducted thoroughly regardless of how many templates are in the system.

Screenshot 2021-02-03 at 16 32 44

AitorAlgorta:
resolves #868 

related to #1228 #1206 #1204 #1126

FOLLOW-UP: https://github.com/airyhq/airy/issues/1239

**changes** 
- Template type in Model (http lib)
- listTemplates endpoint / reducer / action 
- Templates modal + emoji modal 
- Select and send template and emoji
- Filter templates based on their name 
- Error / Not Found / No Templates messages in Template modal 
- Render library renders message and templates (type RenderedContent)

(rather big PR, sorry!)

**screenshots**

Screenshot 2021-03-11 at 15 22 46

Screenshot 2021-03-11 at 15 22 37
(here templates from different sources are displayed in the modal - this is fixed with the change in listTemplates' endpoint. Only templates from the same source as the conversation will be displayed)

Screenshot 2021-03-11 at 18 05 32

Screenshot 2021-03-11 at 15 22 16

Screenshot 2021-03-11 at 15 19 18

template selected (goes in input bar and then is sent)
Screenshot 2021-03-11 at 18 07 18


frontend/ui/src/components/Templates/RenderTemplate.tsx
```diff
@@ -0,0 +1,21 @@
...
+  switch (templateContent.blueprint) {
+    case 'text':
+      return <TextTemplate styleVariant={styleVariant} template={templateContent} isMember={true} />;
+
+    default:
+      return null;
+  }
+};
+
+export default RenderTemplate;
```
AudreyKj, line 21: yeah so I know it makes sense the render library to render the templates, but I guess we would need to tweak it to accept something else than messages

lib/typescript/httpclient/model/RenderedContent.ts
```diff
@@ -0,0 +1,13 @@
...
+  id: string;
+  content: any;
+  deliveryState?: MessageState;
+  senderType?: SenderType;
+  sentAt?: Date;
+}
+
+export function isFromContact(message: RenderedContent) {
+  return message?.senderType === SenderType.sourceContact;
+}
```
AudreyKj, line 13: To render templates inside the templates modal, we needed to tweak the render library so that it renders both messages and templates that are not messages. 

I've created a type for the Render library called _RenderedContent_ which works for both messages and templates. I think this solution is not the most  ideal (as we lose some type checking) and the naming could also be improved, but I couldn't think of anything else

lib/typescript/render/components/Carousel/index.tsx
```diff
@@ -48,11 +48,14 @@ export const Carousel = ({children}) => {
...
-    carouselChildren.current.addEventListener('scroll', () => {
+
+    if (carouselChildren && carouselChildren.current) {
+      resizeObserver.observe(carouselChildren.current);
       resetScrollButtons();
-    });
+      carouselChildren.current.addEventListener('scroll', () => {
+        resetScrollButtons();
+      });
+    }
```
AudreyKj, line 58: fixed this in chat plugin carousel because got error in the UI because of the ref

lib/typescript/httpclient/endpoints/listTemplates.ts
```diff
@@ -0,0 +1,15 @@
...
+export default HttpClient.prototype.listTemplates = async function listTemplates(
+  requestPayload: ListTemplatesRequestPayload
+) {
+  const response: ListTemplatesPayload = await this.doFetchFromBackend('templates.list', {
+    name: requestPayload.name,
+    source_type: requestPayload.sourceType,
+  });
+
+  return camelcaseKeys(response.data);
+};
```
AudreyKj, line 15: I know @bitboxer  made some changes to the TS http library (will rebase/edit when this is merged)

lib/typescript/render/components/Carousel/index.module.scss
```diff
@@ -5,7 +5,7 @@
   display: flex;
   align-items: flex-end;
   flex-direction: column;
-  margin-top: 8px;
+  margin-top: 5px;
   overflow: hidden;
   position: relative;
```
AudreyKj, line 10: the different render components had different values for the _margin-top_ (which displayed weirdly in the modal template) so I've unified everything at 5px (following the Text components)

lib/typescript/render/providers/chatplugin/ChatPluginRender.tsx
```diff
@@ -96,6 +96,11 @@ function mapContent(message: Message): ContentUnion {
       text: messageContent.text,
     };
   }
+
+  return {
+    type: 'text',
+    text: 'Unknown message type',
+  };
 }
```
AudreyKj, line 104: added this return value when the message type is unknown in the chat_plugin render (otherwise we'll get an error for unknown message types)

lib/typescript/render/shared.ts
```diff
@@ -17,19 +17,19 @@ export interface SuggestedReplyCommand extends Command {
...
 export interface MessageRenderProps {
-  message: Message;
+  message: RenderedContent;
   source: string;
   contact?: Contact;
-  lastInGroup: boolean;
+  lastInGroup?: boolean;
   invertSides?: boolean;
   commandCallback?: (command: CommandUnion) => void;
 }
```
AudreyKj, line 27: used the type _RenderedContent___ for the type that is rendered in the library and made the MessageRenderProps exclusive to Messages optional, so that it would work for both Messages & Templates

lib/typescript/httpclient/model/RenderedContent.ts
```diff
@@ -0,0 +1,13 @@
+import {MessageState, SenderType} from './Message';
+
+export interface RenderedContent {
+  id: string;
+  content: any;
+  deliveryState?: MessageState;
+  senderType?: SenderType;
+  sentAt?: Date;
+}
```
AudreyKj, line 9: RenderedContent  type

lib/typescript/httpclient/payload/ListTemplatesRequestPayload.ts
```diff
@@ -0,0 +1,4 @@
+export interface ListTemplatesRequestPayload {
+  name?: string;
+  sourceType: string;
+}
```
AudreyKj, line 4: newest change on _templates.list___: the sourceType is required so that only templates from the source will be fetched

lib/typescript/httpclient/model/RenderedContent.ts
lucapette: Probably already Content is a better name?

chrismatix: Good idea. This is a hint that the render library is too message-specific. An ideal solution would be to extract the rendering logic that is specific to messages (sender information) to a wrapper and have it only expose content rendering logic. 

But that can be done in a separate issue I think.

lucapette: The type is also strange though as it seems still message specific. I will read the whole thing more carefully then

lib/typescript/render/providers/facebook/FacebookRender.tsx
```diff
@@ -139,14 +139,14 @@ function facebookOutbound(message: Message): ContentUnion {
     };
   }
 
-  if (messageJson.attachment || messageJson.message?.attachments) {
```
chrismatix, line 142: This one I am a bit confused about: How would the `RenderedContent` abstraction change the structure of the `content` field? 🤔

lib/typescript/httpclient/model/RenderedContent.ts
lucapette: agree on the first part @chrismatix but honestly wouldn't wait for a separate issue unless we pick it up immediately (which means I'd do here if it was for me)

AudreyKj: yeah I thought about Content, but feel that Content could be confusing, as it doesn't reflect that the type is exclusive to the render library (but yeah RenderedContent is not great)

lib/typescript/httpclient/model/Template.ts
```diff
@@ -0,0 +1,9 @@
+import {MessageSource} from './Message';
+
+export interface Template {
+  id: string;
+  name: string;
+  sourceType: MessageSource;
```
lucapette, line 6: The naming here is off (not sure it was already like that... but we also noticed when discussing channels yesterday). I suppose this should be called `SourceType`

chrismatix: It's too late since the backend PR was already merged, but I just reckoned that since this field is called `source` everywhere else this should also be the case for the field on the templates.

lib/typescript/render/providers/facebook/FacebookRender.tsx
AudreyKj: the `RenderedContent` abstraction does not change the structure of the content field: when we render templates, there is always going to be a message field inside content (whereas it's not always the case for messages)
 
I've just refactored the code so that we don't always have to check for `messageJson.message`. I've added `const messageJson = message.content.message ?? message.content;` (above in the file) to check if the content field contains `message.content.message`

chrismatix: Ah, right I missed that!

bitboxer (approved):

PR #1130, opened 2021-03-03
[#659] Enable to connect via facebook

### LINKED ISSUE #659: Enable users to connect via Facebook Messenger
### Summary
To enhance overall experience, we should offer the ability to connect Facebook Messenger for Facebook business pages.

### Requirements
On Airy Core, it should be an option to connect several Facebook business pages as a means of communication via Facebook Messenger. Before offering the option to connect Facebook pages within channels, we should display the configuration requirements that must be met.
6 0 0 Connect channel Messenger@2x
When connecting a new Facebook business page, the required fields should include:
- Facebook Page ID
- Token
When connecting a new Facebook business page, the optional fields should include:
- Name
- Image URL

If no data is entered into the optional fields, then the standard name and picture should be set to be the same as the name and picture displayed on the corresponding Facebook page.
6 1 0 Connect Messenger@2x

It should be possible for several Facebook business pages to be added within the Facebook Messenger channel, and these pages will all be visibly listed. It should also be possible to see whether the page is properly connected. If it is not, we should display an error message next to the affected page.
6 2 0 Connect Messenger@2x

If an error message is displayed, it should be possible to access the page details previously entered. Here, the field containing incorrect data should be outlined in red with another error message to show that Airy cannot connect with the current token. It should be possible to enter the correct token and then connect the page properly.
6 3 0 Connect Messenger@2x

It should be possible to filter and search within all connected Facebook business pages within the Facebook Messenger channel.

thorstenairy:
resolves #659


frontend/ui/src/pages/Channels/ChannelsMainPage/SourceInfo/index.tsx
```diff
@@ -59,9 +60,9 @@ const SourceInfo = (props: SourceInfoProps) => {
                   );
                 })}
               </div>
-              <div className={styles.extraChannel}>
+              <div className={styles.extraChannel} onClick={() => props.onSourceInfoClick()}>
                 {hasExtraChannels && (
-                  <LinkButton onClick={props.onMoreChannelsClick}>
+                  <LinkButton onClick={() => props.onSourceInfoClick()}>
```
lucapette, line 65: shouldn't we bind in the constructor? This is not recommended by the react team (as it puts additional load on the runtime by creating a function _every time_ the component renders)

Probably also a more general point (as I've seen this quite a bit in the codebase already I think) /cc @AitorAlgorta

frontend/ui/src/pages/Channels/ChannelsMainPage/Sources/FacebookConnect.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const [buttonTitle, setButtonTitle] = useState('Connect Page');
+  const [errorMessage, setErrorMessage] = useState('');
+
+  const buttonStatus = () => {
+    if (id.length > 5 && token != '') {
+      return false;
+    } else {
+      return true;
+    }
+  };
```
lucapette, line 42: `return !(id.length > 5 && token != '')`

frontend/ui/src/pages/Channels/ChannelsMainPage/index.tsx
```diff
@@ -29,8 +56,14 @@ const ChannelsMainPage = (props: ChannelsConnectProps) => {
...
       <div className={styles.wrapper}>
+        {displayDialogFromSource !== '' && openRequirementsDialog(displayDialogFromSource)}
         {props.config.components['sources-chatplugin'].enabled && <ChatPluginSource pluginSource={props.channels} />}
-        {props.config.components['sources-facebook'].enabled && <FacebookSource facebookSource={props.channels} />}
+        {props.config.components['sources-facebook'].enabled && (
+          <FacebookSource
+            facebookSource={props.channels}
+            showDialogAction={(source: string) => setDisplayDialogFromSource(source)}
+          />
+        )}
```
lucapette, line 66: this code reads always the same no matter the source, that's a hint we can simplify the code a bit (also because the naming here is off... why are we passing channels to something called `xSource`?).

One way of doing this creating an helper function that takes the component name and the relevant props and switches on the right component. Then we can `props.config.components.map((name, component) => component.enabled && getSource(name, props))`

frontend/ui/src/pages/Channels/ConnectedChannelsList/ChannelListItem.tsx
```diff
@@ -0,0 +1,151 @@
...
+    }
+  };
+
+  const disconnectChannel = (channelSource: string) => {
+    let source: string;
+    if (channelSource === 'chat_plugin') {
+      source = 'chatplugin';
+    } else {
+      source = channel.source;
+    }
```
lucapette, line 93: a sign we should probably do something about the naming here to avoid those conditionals (we control the names after all)

frontend/ui/src/pages/Channels/ConnectedChannelsList/ChannelListItem.tsx
```diff
@@ -0,0 +1,151 @@
...
+      case ChannelSource.google:
+        break;
+      case ChannelSource.twilioSMS:
+        break;
+      case ChannelSource.twilioWhatsapp:
+        break;
+      case ChannelSource.chatPlugin:
+        return {pathname: CHANNELS_CHAT_PLUGIN_ROUTE + `/${channel.id}`, state: {channel: channel}};
+    }
+  };
```
lucapette, line 85: the only thing changing here is going to be the path right? so we can/should make that clear by mapping channel names to paths (then this becomes one line of code)

frontend/ui/src/pages/Channels/ConnectedChannelsList/ChannelListItem.tsx
```diff
@@ -0,0 +1,151 @@
...
+      case ChannelSource.chatPlugin:
+        if (channel?.metadata?.imageUrl) {
+          return <img className={styles.imageUrlLogo} src={channel.metadata.imageUrl} />;
+        } else {
+          return <AiryLogo />;
+        }
+      default:
+        return <AiryLogo />;
+    }
+  };
```
lucapette, line 70: If we extract the imageUrl if on top (and do an early return) the code will look much better there. But I'd go further and map sources to logos into a function `getSourceLogo`

lib/typescript/httpclient/payload/ExploreChannelRequestPayload.ts
```diff
@@ -1,4 +1,3 @@
 export interface ExploreChannelRequestPayload {
-  source: string;
```
lucapette, line 2: There's no generic explore endpoint anymore so since we're changing this we should rename this one

lib/typescript/httpclient/payload/ConnectChannelRequestPayload.ts
```diff
@@ -1,7 +1,6 @@
-export interface ConnectChannelRequestPayload {
-  source: string;
-  sourceChannelId: string;
-  token: string;
+export interface ConnectChannelFacebookRequestPayload {
+  pageId: string;
+  pageToken: string;
```
lucapette, line 3: I think now I understand a bit better why we're doing this (but the code  reads wrong anyway in the connect page as that's still a generic place).

We should rename the file  though

lucapette (changes requested):
I did a first pass and I think there's good work we can do to improve the way we handle sources. 

Not directly related to your work here @thorstenairy but more a general comment for the @airyhq/codeowners-frontend team

I also believe this is a short term effort and we need to rethink a little how we're doing those pages. The naming is a little confusing (as we have clear definitions of sources/providers/channels and I think the UI code doesn't reflect this well. sources/channels naming is especially mixed). The folder structure of the channels page also doesn't reflect well the nature of the problem, just to clarify what I mean, I think you'd expect this part of the UI to look like this more or less:

- `pages/Channels/index.js` ("home page" of the section)
- `pages/Channels/Facebook/index.js` (for the facebook flow)

Where the top level `pages/Channels` would also contain all the "generic" components this part of the UI needs (and the specialised code would only live under the corresponding source).

The face we have to resort to so many mappings is also a sign the whole design can be improved upon. I promised @AitorAlgorta to propose something about this but didn't have the time to go over it just yet.

frontend/ui/src/pages/Channels/ChannelsMainPage/Sources/FacebookSource.tsx
```diff
@@ -16,11 +23,30 @@ const FacebookSource = (props: FacebookSourceProps) => {
...
+        id={ChannelSource.facebook}
+        onAddChannelClick={() => props.showDialogAction(ChannelSource.facebook)}
       />
 
-      <SourceInfo source="facebook" channels={channels} connected="CONNECTED" isConnected="connected" />
+      <SourceInfo
+        source="facebook"
+        channels={channels}
+        connected="CONNECTED"
+        isConnected="connected"
```
chrismatix, line 34: Something about these props doesn't seem right if they reflect the same state, right? Is there a way that we can merge them?

frontend/ui/src/pages/Channels/ConnectedChannelsList/ChannelsList.tsx
```diff
@@ -0,0 +1,131 @@
...
+  const {channels} = props;
+  const [name, setName] = useState('');
+  const [path, setPath] = useState('');
+  const [searchText, setSearchText] = useState('');
+  const [showingSearchField, setShowingSearchField] = useState(false);
+  const source = props.match.params.source;
+
+  const filteredChannels = channels
+    .filter((channel: Channel) => channel?.source === source)
+    .filter((channel: Channel) => channel?.metadata?.name?.toLowerCase().includes(searchText.toLowerCase()));
```
chrismatix, line 35: You can extract this filter code to the `mapStateToProps` to clean up the component code a bit.

Also, I think you can remove the `?` in `channel?.source`, since all channels in that list are present.

frontend/ui/src/pages/Channels/ChannelsMainPage/SourceInfo/index.tsx
AitorAlgorta: It's true that we miss-used this feature in some places, but I think here it makes sense. We don't want to use the HTML event that we get in the `onClick` we just want to trigger a function when the `onClick` gets triggered, if we would do it differently we would have to pass back the `event` that the `onClick` give us and we won't use it. That's how I see it here @lucapette

frontend/ui/src/pages/Channels/ChannelsMainPage/SourceInfo/index.tsx
```diff
@@ -70,7 +71,7 @@ const SourceInfo = (props: SourceInfoProps) => {
           </div>
 
           <div className={styles.channelButton}>
-            <button type="button" className={styles.addChannelButton} onClick={props.onAddChannelClick}>
+            <button type="button" className={styles.addChannelButton} onClick={() => props.onAddChannelClick()}>
```
AitorAlgorta, line 74: We couldn't re-use the function here because the `onClick` event here is different. @lucapette

frontend/ui/src/pages/Channels/ChannelsMainPage/Sources/FacebookSource.tsx
AitorAlgorta: the prop `connected` has a bad naming, since I think this is the string that we show when we have more than X items in the components. Fixing this would be nice, maybe we don't even new a prop for this and can be inside the component itself.

frontend/ui/src/pages/Channels/ChannelsMainPage/index.tsx
```diff
@@ -15,7 +18,29 @@ type ChannelsConnectProps = {
   config: ConfigPayload;
 };
 
-const ChannelsMainPage = (props: ChannelsConnectProps) => {
+const ChannelsMainPage = (props: ChannelsConnectProps & RouteComponentProps) => {
+  const [displayDialogFromSource, setDisplayDialogFromSource] = useState('');
+
+  const openRequirementsDialog = (source: string): JSX.Element => {
```
AitorAlgorta, line 24: Shouldn't this be a component?

frontend/ui/src/pages/Channels/ChannelsMainPage/index.tsx
AitorAlgorta: I think we should pass the channels already filtered to the component, this way it would be way more cleaner. But maybe we can do this in a different PR, I don't want to be too annoying. If it's not fixed in this PR, please create the issue before merging.

frontend/ui/src/pages/Channels/ConnectedChannelsList/ChannelListItem.tsx
AitorAlgorta: Agree 100% this doesn't look pretty

frontend/ui/src/pages/Channels/ConnectedChannelsList/ChannelsList.tsx
AitorAlgorta: I just wrote the same in a different place before reading this comment. Yes 100%

AitorAlgorta (approved):
LGTM, please create the issue for the improvements

lucapette (approved):

PR #1149, opened 2021-03-04
[#1142] Backbutton & search aligned

### LINKED ISSUE #1142: backButton and searchField are not aligned in inbox
Screenshot 2021-03-04 at 10 52 46

ogbeche77:
Fixes #1142
Screen Shot 2021-03-04 at 12 45 17

### DIFF
--- a/frontend/ui/src/pages/Inbox/ConversationListHeader/index.module.scss
+++ b/frontend/ui/src/pages/Inbox/ConversationListHeader/index.module.scss
@@ -74,15 +74,14 @@
 }
 
 .searchIcon {
-  margin-top: 7px;
+  width: 22px;
+  height: 22px;
+  padding-top: 2px;
+  margin-right: 4px;
+  position: relative;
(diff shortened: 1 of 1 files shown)


frontend/ui/src/pages/Inbox/ConversationListHeader/index.module.scss
```diff
@@ -75,7 +75,14 @@
 
 .searchIcon {
   margin-top: 7px;
+
   path {
     fill: var(--color-text-gray);
+    width: 22px;
```
bitboxer, line 81: Does it also work when you attach it to the `.searchIcon` and not the path? Somehow it feels strange to me to style the path that way.

ogbeche77: It should be in the `.searchIcon`. Fixing that

bitboxer (approved):
LGTM

PR #1199, opened 2021-03-10
[#871] Httpclient methods need return value

### LINKED ISSUE #871: httpclient methods need return value
also maybe we can configure typescript that it yells when the return value is not defined?

bitboxer:
closes #871


frontend/ui/src/actions/conversations/index.ts
```diff
@@ -87,7 +87,7 @@ export const getConversationInfo = (conversationId: string, retries?: number) =>
     });
 
 export const readConversations = (conversationId: string) => (dispatch: Dispatch<any>) => {
-  HttpClientInstance.readConversations(conversationId).then(() =>
+  HttpClientInstance.markConversationAsRead(conversationId).then(() =>
```
AitorAlgorta, line 90: Having the same name as the endpoint was intended. I think it's good for consistency. I would like to keep it that way. What do you think @lucapette @airyhq/codeowners-frontend ?

lib/typescript/httpclient/endpoints/updateTag.ts
```diff
@@ -1,7 +1,8 @@
 import {Tag} from '../model';
 import {HttpClient} from '../client';
 
-export default HttpClient.prototype.updateTag = async function (tag: Tag) {
-  await this.doFetchFromBackend('tags.update', {...tag});
-  return Promise.resolve(true);
+export default HttpClient.prototype.updateTag = function (tag: Tag) {
+  return this.doFetchFromBackend('tags.update', {...tag})
+    .then(() => true)
+    .catch(() => false);
```
AitorAlgorta, line 7: Can't we do this in a different way? It doesn't feel right @bitboxer

frontend/ui/src/actions/conversations/index.ts
bitboxer: The method name looked not right for me. More a "get a list of conversations" and not "mark this one conversation as read"

lib/typescript/httpclient/endpoints/updateTag.ts
bitboxer: It always returned `true` before. I just changed the thing to actually return `false` if something went wrong. But I can totally refactor it and not make it return a boolean but a real promise object

frontend/ui/src/actions/conversations/index.ts
lucapette: I agree it doesn't look right but I would also not open the door for diverging the api methods names from the actual endpoints (the convention is "reverse the namespace") right now. 

So what I think we should do is keeping the convention and rediscuss the naming of `/conversations.read` (because I would argue that's the wrong name?)

lib/typescript/httpclient/endpoints/updateTag.ts
```diff
@@ -1,7 +1,6 @@
 import {Tag} from '../model';
 import {HttpClient} from '../client';
 
-export default HttpClient.prototype.updateTag = async function (tag: Tag) {
-  await this.doFetchFromBackend('tags.update', {...tag});
-  return Promise.resolve(true);
+export default HttpClient.prototype.updateTag = function (tag: Tag) {
+  return this.doFetchFromBackend('tags.update', {...tag});
```
AitorAlgorta, line 5: ❤️

frontend/ui/src/actions/conversations/index.ts
AitorAlgorta: Yeah, I agree. The problem is the name of the endpoint, maybe `/conversations.markAsRead` or maybe `/conversations.markAs.read` and in the future we can have more endpoints like `/conversations.markAs.XXXX`

AitorAlgorta: or without the `markAs`, just `mark`

bitboxer: https://github.com/airyhq/airy/issues/1211 created a ticket for it, reverted the change here for now.

chrismatix (approved):
🎉

AitorAlgorta (approved):
It looks way better! 🚀

AitorAlgorta (approved):

PR #1207, opened 2021-03-10
Chatplugin overview section

lucapette:


docs/docs/sources/chatplugin/overview.md
```diff
@@ -4,13 +4,57 @@ sidebar_label: Overview
...
+automate FAQs.
+
+Airy’s Live Chat Plugin comes out of the box fully working, but thanks to its
+open-source nature and React Render Props you can customize everything about it.
+
+Out of the box Airy’s Live Chat Plugin supports:
+
+- Full customization of look, feel and features
+- All message types, including emojis
+- Rich Messaging: Templates, cards & carousels
```
AitorAlgorta, line 31: Templates and cards are the same right?

sgarroum: templates in Facebook, cards in Google. better to have all mentioned for people who only know one of the terms

sgarroum (approved):
perfection. and with perfection I mean okay first start :D

PR #1216, opened 2021-03-11
[#1214] Duplicate messages in the redux store

### LINKED ISSUE #1214: Duplicate messages in the redux store
I am not sure how I ended up with this situation, but I have this state in my app right now:

Screen Shot 2021-03-11 at 09 28 11

We need to make sure that we don't add a message more than once to our store. The uuid should guarantee us that this message is unique.

bitboxer:
closes #1214

### DIFF
--- a/frontend/ui/src/reducers/data/messages/index.ts
+++ b/frontend/ui/src/reducers/data/messages/index.ts
@@ -28,27 +28,6 @@ function mergeMessages(oldMessages: Message[], newMessages: Message[]): Message[
 export default function messagesReducer(state = initialState, action: Action): Messages {
   switch (action.type) {
     case getType(actions.loadingMessagesAction):
-      if (state.all[action.payload.conversationId]) {
-        return {
-          ...state,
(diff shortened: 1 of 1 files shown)


frontend/ui/src/reducers/data/messages/index.ts
```diff
@@ -28,27 +28,6 @@ function mergeMessages(oldMessages: Message[], newMessages: Message[]): Message[
...
   switch (action.type) {
     case getType(actions.loadingMessagesAction):
-      if (state.all[action.payload.conversationId]) {
-        return {
-          ...state,
-          all: {
-            ...state.all,
-            [action.payload.conversationId]: [
-              ...mergeMessages([...action.payload.messages], []),
-              ...state.all[action.payload.conversationId],
```
bitboxer, line 38: This line was the mistake. Adding existing messages without running them through the `mergeMessage` function.

AitorAlgorta (approved):
💯

Issue #677, opened 2021-01-18
Render Suggested Replies


lizhutter:
### Summary

The following criteria must be met for each suggested reply to be eligible to be rendered:
- ID
- name
- source type (This is necessary because we have to know for which source they are. We need to make it possible to have some replies that are source agnostic.)
- content (This will be very source specific, so our system should accept whatever JSON is saved because there is no CMS feature on our side.)

### Requirements

If a question is received by a system user on Airy Core, it should be possible to view suggested replies. If a system user clicks the suggested reply modal, we should offer the option of one or several reply options. Once the system user selects a suggested reply to send, it should auto-fill the input bar and the system user should be able to send this message.

![02 0 Suggested Replies one option@2x](https://user-images.githubusercontent.com/74183162/106909707-d66c1300-6700-11eb-8b3a-9c49618fbd3c.png)
![02 1 Suggested Replies one option@2x](https://user-images.githubusercontent.com/74183162/106909713-d79d4000-6700-11eb-8642-215244eb41d3.png)
![02 2 Suggested Replies one option@2x](https://user-images.githubusercontent.com/74183162/106909720-d835d680-6700-11eb-8ef8-632bffeec243.png)
![01 0 Suggested Replies – many options@2x](https://user-images.githubusercontent.com/74183162/106907850-fc90b380-66fe-11eb-93a6-195649f2f51c.png)
![01 1 0 Suggested Replies – many options@2x](https://user-images.githubusercontent.com/74183162/106907864-ff8ba400-66fe-11eb-9b94-6006a6490112.png)
![01 1 1 Suggested Replies – many options@2x](https://user-images.githubusercontent.com/74183162/106907872-00bcd100-66ff-11eb-921d-0aed573d8439.png)
![01 2 Suggested Replies@2x](https://user-images.githubusercontent.com/74183162/106907877-01edfe00-66ff-11eb-84e7-e0c65d90ca24.png)
![01 3 Suggested Replies@2x](https://user-images.githubusercontent.com/74183162/106907881-031f2b00-66ff-11eb-828c-6069d173c250.png)

We need to be able to render different types of templates regardless of whether they are composed of text, media, or buttons.

![03 0 Suggested Replies – Templates@2x](https://user-images.githubusercontent.com/74183162/106909903-01566700-6701-11eb-937a-e9b3c5f4975a.png)
[...]

PR #1224, opened 2021-03-11
[#677] add missing ingress definition for suggest replies endpoint

chrismatix:
required for #677


lucapette (approved):
amazing. We never remember 😂

Issue #1018, opened 2021-02-17
Introduce AWS provider for airy create


ljupcovangelski:
## Open questions

- [x] Do we use eks? 
> Yes
- [x] What are the required credentials and how do we require them? (env seems the most common approach there)
> There will be a `aws-profile` flag on the command
- [x] What else do we need to actually make this provider work? (For example: load balancer?)
> If a VPC and 2 subnets are already created, then we need an EKS cluster, 1 Node and 1 Load Balancer

lucapette:
this depends on #1017 (I know we have a branch about this... but it's really not a good plan to do this first)

paulodiniz:
About the topics that @ljupcovangelski mentioned

> We use eks?

I think so. Or better put: I don't see a reason not to

> What are the required credentials and how do we require them? (env seems the most common approach there)

I think having a valid `~/.aws/config` could be a requirement for this provider. What do you think?

> What else do we need to actually make this provider work? (For example: load balancer?)

From my AWS understanding

1. Create a VPC
2. Create two subnets in different regions (`us-east-1a` and `us-east-1b` for example). Both subnets must be contained in the VPC
3. Create a Arn Role
4. Create the cluster
5. Create a Load Balancer

ljupcovangelski:
I agree with @paulodiniz that those are the steps needed to make this provider work. I just want to add two notes:
- Apart from creating the EKS cluster, we also need to add nodes/instances to the cluster. How many instances and which size would we recommend? Would be good to make this configurable and in correlation with some resources request/limits of the components.
- Perhaps we can leave an option for people to use an existing VPC, which will facilitate connecting of the EKS cluster (and the components) to the prerequisites, in case somebody is running Kafka/PostgreSQL and Redis on their own and in an already running cluster.

For the other two points:
- I think we should use an EKS
- For the profile, I think we can introduce a flag `--aws-profile`, connected also to an ENV variable `AWS_PROFILE`

paulodiniz:
@ljupcovangelski Good point about adding the nodes!

On adding the `vpc` options, I think it's a nice to have, but we could go without it.

I don't understand why the `AWS_PROFILE`, isn't it easier to rely on the aws config file itself?

ljupcovangelski:
> I don't understand why the `AWS_PROFILE`, isn't it easier to rely on the aws config file itself?

Because in the config file, one can have multiple profiles configured.

paulodiniz:
[...]

PR #1240, opened 2021-03-12
[#1018] Introduce aws provider for airy create

ljupcovangelski:
Steps to provision

- [x] Create IAM role for EKS
- [x] Create VPC or use existing one
- [x] Create two subnets in the VPC
- [x] Create the EKS cluster
- [x] Wait for cluster to be in a ready state
- [x] Create an NodeGroup
- [x] Start one node in the NodeGroup
- [x] Generate the kube config file
- [x] Retrieve the url of the loadbalancer and inject it in a configMap post installing


infrastructure/cli/pkg/providers/aws/aws.go
```diff
@@ -4,14 +4,31 @@ import (
 	"cli/pkg/kube"
 	"context"
 	"fmt"
+	"log"
```
lucapette, line 7: we shouldn't use log yet I think (we're trying hard not to use it anywhere short term so we can do one coherent pass at the output

WORKSPACE
```diff
@@ -139,7 +139,28 @@ go_register_toolchains(
     version = "1.16",
 )  # airy_nogo is in the top-level BUILD file of this workspace
 
-load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies")
+load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies", "go_repository")
```
chrismatix, line 142: Sometimes gazelle does not seem to pick up the `to_macro` directive, which tells it to write these to the `go_repositories.bzl`. I fixed this for the `update-deps` command on `develop`. So if you remove this, rebase and run `update-deps` they should show up in the repositories file :)

docs/docs/getting-started/installation/aws.md
```diff
@@ -0,0 +1,127 @@
...
+
+import Script from "@site/src/components/Script";
+
+<Script data-cols="90" id="asciicast-KHz6yTASgqwnKT5FB7Gt7jF1k" src="https://asciinema.org/a/KHz6yTASgqwnKT5FB7Gt7jF1k.js"></Script>
+
+If you want to customize your `Airy Core` instance please see our [Configuration Section](configuration.md).
+
+## Public webhooks
+
+The public webhooks will be accessible on the public LoadBalancer which is created by the Ingress loadBalancer Kubernets service.
```
lucapette, line 44:
```suggestion
The public webhooks will be accessible on the public LoadBalancer which is created by the Ingress loadBalancer Kubernetes service.
```

docs/docs/getting-started/installation/aws.md
```diff
@@ -0,0 +1,127 @@
...
+
+Export your AWS_PROFILE as described in the [AWS documentation](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html).
+
+Now you can run this command, which will create Airy core in your AWS account:
+
+```bash
+airy create --provider=aws
+```
+
+By default, the installation will create a single EC2 Kubernetess node, as part of a single node group. You can scale your EKS cluster by adding more nodes or node groups, through the AWS web console or the AWS cli.
```
lucapette, line 32:
```suggestion
By default, the installation will create a single EC2 Kubernetes node, as part of a single node group. You can scale your EKS cluster by adding more nodes or node groups, through the AWS web console or the AWS CLI.
```

docs/docs/getting-started/installation/aws.md
```diff
@@ -0,0 +1,127 @@
+---
+title: AWS
+sidebar_label: AWS
+---
+
+import useBaseUrl from '@docusaurus/useBaseUrl';
+
+The goal of this document is to provide an overview of how to run Airy Core on
```
lucapette, line 8: Right before this, so that @sgarroum is happy :), I would add a TLDR component with content along the lines of "Install Airy Core on AWS with one command"

docs/docs/getting-started/installation/aws.md
```diff
@@ -0,0 +1,163 @@
...
+## Where to go from here
+
+Now that you have a running local installation of Airy core on AWS you can connect it to messaging sources. Check out the
+[source documentation](/sources/introduction.md) to learn more.
+
+## Third party tools
+
+Third party tools can be activated in the `airy.yaml` configuration file, under the `tools` section.
+For more details please see our [Configuration Section](configuration.md).
+
```
ljupcovangelski, line 71: For the Uninstall part, it can be also a script which gets the id and removes all the resources.
However, for the first version, I thought better to have separate steps, as sometimes something fails to delete and it is not fully stable yet. When something fails, it should be deleted from the AWS console.

infrastructure/cli/pkg/kube/context.go
```diff
@@ -21,10 +20,6 @@ func New(kubeConfigPath, contextName string) KubeCtx {
 
 func (c *KubeCtx) GetClientSet() (*kubernetes.Clientset, error) {
 	if c.ContextName == "" {
-		if c.KubeConfigPath == "" {
```
chrismatix, line 24: Why the removal? 🤔

chrismatix: I agree that the nested `if` condition doesn't look great, but if we let this pass `BuildConfigFromFlags` will return the current context which means that it could talk to whatever current profile the user is using.

infrastructure/cli/pkg/providers/aws/aws.go
```diff
@@ -1,104 +1,578 @@
...
-	"log"
-	"os"
+	eksTypes "github.com/aws/aws-sdk-go-v2/service/eks/types"
+	"github.com/aws/aws-sdk-go-v2/service/iam"
+	iamTypes "github.com/aws/aws-sdk-go-v2/service/iam/types"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/mitchellh/go-homedir"
 )
 
+const RolePolicyDocument = `{
```
chrismatix, line 26: We could embed this in a file. Checkout the `workspace` module for an example :)

infrastructure/cli/pkg/providers/aws/aws.go
```diff
@@ -1,104 +1,578 @@
...
+
+var letters = []rune("abcdefghijklmnopqrstuvwxyz")
+
 type Aws struct {
+	context kube.KubeCtx
 }
 
 func (a *Aws) GetHelmOverrides() []string {
-	return []string{}
+	return []string{"--set", "global.ngrokEnabled=false", "--set", "global.host=http://airy.core"}
```
chrismatix, line 49: I wouldn't set any global host if it is incorrect. I think instead we should make the host optional on the helm chart because we need to accommodate for the weakest provider constraint.

infrastructure/cli/pkg/providers/aws/aws.go
```diff
@@ -1,104 +1,578 @@
...
 	return nil
 }
 
+type KubeConfig struct {
+	ClusterName     string
+	EndpointUrl     string
+	CertificateData string
+}
+
+const KubeConfigTemplate = `
```
chrismatix, line 85: Same point about embedding here.

infrastructure/cli/pkg/providers/aws/aws.go
```diff
@@ -1,104 +1,578 @@
...
+        - "get-token"
+        - "--cluster-name"
+        - "{{.ClusterName}}"
+`
+
 func (a *Aws) Provision() (kube.KubeCtx, error) {
 	cfg, err := config.LoadDefaultConfig(context.TODO())
 	if err != nil {
-		log.Fatal(err)
+		fmt.Println(err.Error())
```
chrismatix, line 116: The method has an `error` signature, but we treat every error like a hard crash. Imho post-installation errors are expected and can thus be relayed back to the caller.

chrismatix (commented):
Good stuff!

infrastructure/cli/pkg/kube/context.go
paulodiniz: I actually can't imagine why. Do you know something about it @ljupcovangelski ?

ljupcovangelski: I'm not sure, I don't recall changing anything in this file. Will bring it back.

lucapette (commented):
Ok did a first pass.

Want to read the go code one more time before commenting that

infrastructure/cli/pkg/providers/aws/aws.go
```diff
@@ -1,104 +1,534 @@
...
+		fmt.Println("error parsing template", err)
+		os.Exit(1)
 	}
 
-	log.Println(output)
-	fmt.Println("aws provider not yet implemented")
-	os.Exit(1)
-	return kube.KubeCtx{}, nil
+	home, err := homedir.Dir()
+	kubeConfigFilePath := path.Join(home, "/.airy/kube.conf")
```
lucapette, line 409:
```suggestion
	kubeConfigFilePath := path.Join(home, ".airy", "kube.conf")
```

infrastructure/cli/pkg/providers/aws/aws.go
```diff
@@ -1,104 +1,534 @@
...
-		log.Fatal(err)
+		fmt.Println("error parsing template", err)
+		os.Exit(1)
 	}
 
-	log.Println(output)
-	fmt.Println("aws provider not yet implemented")
-	os.Exit(1)
-	return kube.KubeCtx{}, nil
+	home, err := homedir.Dir()
```
lucapette, line 408: we want to check this error I think

infrastructure/cli/pkg/providers/aws/aws.go
```diff
@@ -1,104 +1,534 @@
...
+	_, err := ec2Client.ModifySubnetAttribute(context.TODO(), &ec2.ModifySubnetAttributeInput{
+		SubnetId: subnetId,
+		MapPublicIpOnLaunch: &ec2Types.AttributeBooleanValue{
+			Value: true,
+		},
+	})
+	if err != nil {
+		return err
+	}
+	return nil
```
lucapette, line 343:
```suggestion
	return err
```

infrastructure/cli/pkg/providers/aws/aws.go
```diff
@@ -1,104 +1,534 @@
...
+
+func (a *Aws) associateSubnetToRouteTable(ec2Client *ec2.Client, subnetId *string, routeTableId *string) error {
+	_, err := ec2Client.AssociateRouteTable(context.TODO(), &ec2.AssociateRouteTableInput{
+		RouteTableId: routeTableId,
+		SubnetId:     subnetId,
+	})
+	if err != nil {
+		return err
+	}
+	return nil
```
lucapette, line 354:
```suggestion
	return err
```

infrastructure/cli/pkg/providers/aws/aws.go
ljupcovangelski: We do not know the host upfront and the helm chart requires this to be set, otherwise will exit with en error.
I think we should set a default for the `global.host` as the chart as a whole should be deplorable with `helm install` without any needed overrides (--values/--set).
For example, the ingress resource doesn't need any specific host to work, but the `hostnames` configMap requires this value, otherwise the apps that use it will crash.
I will do it like that now (put default for `global.host`) and we can open an issue to review this and do it properly in the future.

lucapette (approved):

PR #1248, opened 2021-03-15
[#1247] Optional variables for templates creation

lucapette:


paulodiniz (dismissed):

chrismatix (changes requested):
Discussed with @lucapette  that it's perhaps even better to remove them all together

paulodiniz (approved):

chrismatix (approved):

PR #1251, opened 2021-03-15
fixed templates modal and fixed type template with api change

### LINKED ISSUE #1249: Small fix UI Templates
**to do**
- make modal window resizable, we don’t need a super big section if we don’t have any template

AudreyKj:
resolves #1249 

(i'm opening the PR, but we can add more small UI fix to it if needed)

changes:
- modal template resizable 
- templates api changes (variables optional)
Screenshot 2021-03-15 at 15 46 42

### DIFF
--- a/lib/typescript/httpclient/model/Template.ts
+++ b/lib/typescript/httpclient/model/Template.ts
@@ -5,5 +5,5 @@ export interface Template {
   name: string;
   sourceType: SourceType;
   content: string;
-  variables: any;
+  variables?: any;
 }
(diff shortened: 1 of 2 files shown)


lib/typescript/httpclient/model/Template.ts
```diff
@@ -5,5 +5,5 @@ export interface Template {
   name: string;
   sourceType: SourceType;
   content: string;
-  variables: any;
+  variables?: any;
```
lucapette, line 8: I'm removing this from the api as its introduction was premature (I planned to remove it from here as well but since we're at it we can do here)

AudreyKj: ahh ok, I thought it was only becoming optional! thanks for the update

lucapette (approved):

Issue #1090, opened 2021-02-25
Add URL check for image while connect a new channel


thorstenairy:
When you add an invalid URL for an image, while connecting a new channel, the not found image appears next to the channel name and shows the alt of the image which covers the channel name itself.

PR #1254, opened 2021-03-15
[#1090] Add fallback image to channels

thorstenairy:


frontend/ui/src/services/image/index.ts
```diff
@@ -0,0 +1,6 @@
+import _, {SyntheticEvent} from 'react';
+
+export const fallbackImage = (event: SyntheticEvent<HTMLImageElement, Event>, channelSource: string) => {
```
lucapette, line 3: source better name

frontend/ui/src/services/image/index.ts
```diff
@@ -0,0 +1,6 @@
...
+  event.currentTarget.src = `https://s3.amazonaws.com/assets.airy.co/${channelSource}_avatar.svg`;
+  event.currentTarget.alt = 'fallback image';
```
lucapette, line 5: `default ${source} image` maybe?

lucapette (commented):
Minor comments :)

frontend/ui/src/services/image/index.ts
thorstenairy: as we use the _avatar name exactly for those images, I would keep it.

AitorAlgorta (approved):

Issue #1270, opened 2021-03-17
Installation: Toggle broken & Update for Airy Create


sgarroum:
Toggle is broken: Homebrew cannot be selected, it switches between macOS & Linux. All 3 should be selectable and show the right content for each.

This document also needs an update now that Airy Create is here, could be done in one issue.

PR #1273, opened 2021-03-17
[#1270] Installation: Toggle broken & Update for…

pascal-airy:
…Airy Create


sgarroum (approved):
perfection (we will add the update for Airy Create later)

Issue #1265, opened 2021-03-16
CLI Doc Cleanup: Merge Cheatsheet from Intro into Reference, delete intro


sgarroum:
Intro of the CLI is currently only the cheatsheet and TLDR, so we should move the TLDR to the installation part, and the cheatsheet to the reference to clean it up

lucapette:
done in #1277

PR #1277, opened 2021-03-18
[#1265] Merge cheatsheet from introduction into reference

ogbeche77:
Fixess #1265


docs/docs/cli/reference.md
```diff
@@ -54,6 +54,19 @@ airy api signup [flags]
...
+| Command             | Effect                                                                             |
+| ------------------- | ---------------------------------------------------------------------------------- |
+| `airy init`         | Inits your airy configuration                                                      |
+| `airy status`       | Reports the status of an Airy Core instance                                        |
+| `airy config apply` | Applies configuration values from airy.yaml configuration to an Airy Core instance |
+| `airy ui`           | Opens the Airy Core UI in your local browser                                       |
+| `airy api signup`   | Creates a default user                                                             |
+| `airy api login`    | Returns a valid jwt token                                                          |
+| `airy api login`    | Returns a valid jwt token                                                          |
+| `airy --help`       | Prints all available commands with parameters                                      |
```
lucapette, line 69: the file is auto generated so this will break the build.

But the actual plan for this issue was to remove the cheat sheet all together (I know it was not clear from the issue... but I "forgot" to comment there... and I assumed I'd do this myself :)) so fortunately the fix is pretty simple. We keep all the other changes you make (because they're good!) and revert changes to this file.

lucapette (changes requested):
As per comment, we have to revert changes to the reference file

lucapette (approved):
:shipit:

PR #1297, opened 2021-03-19
Fix/1239 message wrapper for render library

### LINKED ISSUE #1239: Message Wrapper for render library
follow up to issue #868, see discussion in PR #1123 

**Current State** 
The Render Library is too message-specific, while we need it to render other types of content that are not messages (eg. templates). 

Following @chrismatix's brilliant idea, we should extract the rendering logic that is specific to messages (sender information) to a wrapper and have it only expose content rendering logic.

**To do** 
- Extract rendering logic that is specific to messages (sender information) to a separate wrapper 

**Goal**
- Optimize the Render Library so that its core focus is only to render all types of content, across different sources

AudreyKj:
closes #1239 

**changes**
- updated typing related to the templates api (both in frontend and httpClient)
- created SharedServices to share fallbackImage across frontend and the render library 
- typing for rendering content
- render library revamp with better typing + cleaned its components from all unnecessary props
- moved all message rendering logic to `InfoMessageWrapper` and updated all the frontend

**screenshots**
--> as all the message-rendering logic is moved to the wrapper, the template modal renders nicely all types of templates + the fallback image improves user experience for all non-valid images 

Screenshot 2021-03-17 at 17 49 01
Screenshot 2021-03-17 at 15 00 48


lib/typescript/render/shared.ts
```diff
@@ -16,21 +15,28 @@ export interface SuggestedReplyCommand extends Command {
...
+
+export const getDefaultRenderingProps = (props: RenderPropsUnion): DefaultRenderingProps => {
+  const fromContact = isFromContact(props.renderedContent);
+
   return {
     fromContact,
-    contact: props.contact ?? null,
-    commandCallback: props.commandCallback,
-    sentAt: props.lastInGroup ? formatTime(props.message.sentAt) : null,
   };
```
AudreyKj, line 41: All the logic exclusive to rendering messages is moved to the wrapper <InfoMessageWrapper />, which displays the Avatar and sentAt time along with the message's content. This wrapper is also responsible for determining is which side the messages are displayed.

-  _RenderProps__ contains the props common to all the content  that the library needs to render (i.e. messages and templates)
-  _DefaultRenderingProps_  are needed by some components that the library needs to render, across all content (this is needed for all the components that use <Text />, as this props will determine the background-color of the message)

lib/typescript/httpclient/model/Content.ts
```diff
@@ -0,0 +1,13 @@
...
+export interface Content {
+  id: string;
+  content: any;
+}
+
+export type RenderedContentUnion = Message | Template;
+
+export function isFromContact(message: RenderedContentUnion) {
+  if (message && 'senderType' in message) return message?.senderType === SenderType.sourceContact;
+}
```
AudreyKj, line 13: typing for render library: _RenderedContentUnion_ represents the type for the content that the library is rendering

frontend/ui/src/reducers/data/templates/index.ts
```diff
@@ -22,7 +23,8 @@ export default function templatesReducer(state = defaultState, action: Action):
     case getType(actions.listTemplatesAction):
       return {
         ...state,
-        all: action.payload,
+        all: action.payload.templates,
+        source: action.payload.source,
       };
```
AudreyKj, line 28: This keeps track of the source of the templates listed,  which is needed because we fetch the list of templates when the conversation's source channel changes (as the api lists templates based on the source).

lib/typescript/render/shared.ts
```diff
@@ -1,6 +1,5 @@
-import {isFromContact, RenderedContent, Contact} from 'httpclient';
-import {formatTime} from 'dates';
-import {DefaultMessageRenderingProps} from './components';
+import {isFromContact, RenderedContentUnion} from 'httpclient';
```
chrismatix, line 1: nitpick: `shared.ts` was never a great name for this file. Looking at it, what do you think about `props.ts` or something else?

frontend/sharedServices/fallbackImage.ts
```diff
@@ -0,0 +1,14 @@
...
+    source === 'mediaImage'
+      ? `https://s3.amazonaws.com/assets.airy.co/fallbackMediaImage.svg`
+      : `https://s3.amazonaws.com/assets.airy.co/${source}_avatar.svg`;
+
+  if (source === 'mediaImage') {
+    event.currentTarget.style.objectFit = 'contain';
+  }
+
+  event.currentTarget.alt = 'fallback image';
+};
```
AudreyKj, line 14: With @thorstenairy we had the idea of creating _SharedServices_ that contains helper functions needed by both the frontend and our libraries. Here fallbackImage serves for images in both the frontend and the render library.

lib/typescript/httpclient/model/Message.ts
```diff
@@ -1,13 +1,6 @@
 /* eslint-disable @typescript-eslint/no-var-requires */
 const camelcaseKeys = require('camelcase-keys');
-
-export enum Source {
-  facebook = 'facebook',
-  google = 'google',
-  chatplugin = 'chatplugin',
-  smsTwilio = 'twilio.sms',
-  whatsappTwilio = 'twilio.whatsapp',
-}
```
AudreyKj, line 10: the enum SourceType was duplicated in both Message and Channels in the httpClient library. I've kept the one in Channels.

lib/typescript/render/components/RichCard/Media/index.tsx
```diff
@@ -18,5 +19,6 @@ export const Media = ({height, contentInfo: {altText, fileUrl}}: MediaRenderProp
     className={`${styles.mediaImage} ${
       height === MediaHeight.tall ? styles.tall : height === MediaHeight.medium ? styles.medium : styles.short
     }`}
+    onError={(event: React.SyntheticEvent<HTMLImageElement, Event>) => fallbackImage(event, 'mediaImage')}
   />
```
AudreyKj, line 23: added fallbackImage for all the images inside the render library. This fallback does not work for video, I've created an issue for this #1299

frontend/sharedServices/fallbackImage.ts
lucapette: yeah so this was brought up many times and I should probably write down the answer eheh

the tl;dr is "shared" "util" is always bad as then we mix things that have nothing to do with each other. (I think last I talked to @thorstenairy exactly about this because he introduced a shared lib with was a date lib)

A better approach is to have small libraries that solve a specific problem.

to me, it feels like here we're introducting an `avatar library`. So best is 1) call it what it does 2) that way we can move more "avatar related things" there 3) _by far_ themost important point: we do not open the door to catch all libraries

frontend/sharedServices/BUILD
```diff
@@ -0,0 +1,21 @@
+load("//tools/lint:web.bzl", "web_lint")
+load("@com_github_airyhq_bazel_tools//lint:buildifier.bzl", "check_pkg")
+load("@com_github_airyhq_bazel_tools//web:typescript.bzl", "ts_web_library")
+
+package(default_visibility = ["//visibility:public"])
```
chrismatix, line 5: I get the intention of this module, but I think the name is too generic and not quite accurate (It hosts a component and not a service). Only thing I can think of for now would be `frontend/components`

chrismatix (commented):
First pass. Amazing work already!

lib/typescript/httpclient/model/Message.ts
lucapette: source was a better name :D (sorry... I suggested source type somewhere else.... without realising we call the field source everywhere in the api)

lib/typescript/render/shared.ts
AudreyKj: yeah I think _props.ts_ makes sense :)

frontend/sharedServices/BUILD
lucapette: `frontend/lib/avatar` would be my suggestion (given the previous point I made about it)

frontend/sharedServices/fallbackImage.ts
AudreyKj: @lucapette the idea was to use this function also for the fallback images in the render library. I totally get the point of not falling in 'catch all libraries' so should we create 2 libraries: one `avatar library`  used for the channel avatars (and possibly also the Avatar component), and another library for the fallback images in the render library?

lucapette: duplicating a little of code is better than the wrong abstraction for sure

lucapette: ok we talked offline about it: we do not need to introduce libraries as the abstraction is not worth it. //c @thorstenairy

frontend/ui/src/pages/Channels/ConnectedChannelsBySourceCard/index.tsx
```diff
@@ -2,25 +2,25 @@ import React from 'react';
 import {RouteComponentProps, withRouter} from 'react-router-dom';
 
 import {LinkButton} from '@airyhq/components';
-import {Channel, SourceType} from 'httpclient';
-import {SourceTypeInfo} from '../MainPage';
+import {Channel, Source} from 'httpclient';
+import {SourceInfo} from '../MainPage';
 import {fallbackImage} from '../../../services/image/index';
```
AudreyKj, line 7: here updated naming to follow api naming **Source**

frontend/ui/src/pages/Inbox/Messenger/ConversationMetadata/index.tsx
```diff
@@ -184,7 +184,7 @@ const ConversationMetadata = (props: ConnectedProps<typeof connector>) => {
               <AvatarImage contact={contact} />
             </div>
 
-            <div className={styles.displayName}>{contact.displayName}</div>
+            <div className={styles.displayName}>{contact && contact.displayName}</div>
```
bitboxer, line 187: Like above:

```suggestion
            <div className={styles.displayName}>{contact?.displayName}</div>
```

infrastructure/.vagrant/machines/airy-core/virtualbox/action_provision
```diff
@@ -0,0 +1 @@
+1.5:b2a9215c-db31-4f1b-b3db-1e3a1e879bbe
```
bitboxer, line 1: I think the `.vagrant` folder is gone now because `minikube` ?

bitboxer (commented):
Awesome stuff 🥳 . Some minor comments. After the rebase is done I will 🟢  this.

infrastructure/.vagrant/machines/airy-core/virtualbox/action_provision
AudreyKj: yes! don't know how this snugged into my PR!

lucapette (changes requested):
Looking great!

I think we should merge, I'm requesting changes so we don't forget to remove all the vagrant files before we merge :)

bitboxer (dismissed):
:shipit:

lucapette (approved):

PR #1324, opened 2021-03-22
[#677] Render Suggested Replies

### LINKED ISSUE #677: Render Suggested Replies

bitboxer:
closes #677

### DIFF
--- a/frontend/ui/src/pages/Inbox/MessageInput/index.tsx
+++ b/frontend/ui/src/pages/Inbox/MessageInput/index.tsx
@@ -23,7 +23,6 @@ import {getCurrentMessages} from '../../../selectors/conversations';
 
 import styles from './index.module.scss';
 import SuggestedReplySelector from '../SuggestedReplySelector';
-import {SuggestionsContent} from 'render/providers/google/googleModel';
 
 const mapDispatchToProps = {sendMessages};
 
@@ -62,7 +61,7 @@ const MessageInput = (props: MessageInputProps & ConnectedProps<typeof connector
   useEffect(() => {
     if (input) {
       textAreaRef.current.style.height = '0px';
-      const scrollHeight = textAreaRef.current.scrollHeight;
+      const scrollHeight = Math.min(100, textAreaRef.current.scrollHeight);
       textAreaRef.current.style.height = scrollHeight + 'px';
     }
   }, [input]);
--- a/frontend/ui/src/pages/Inbox/MessageInput/index.tsx
+++ b/frontend/ui/src/pages/Inbox/MessageInput/index.tsx
@@ -1,28 +1,33 @@
-import React, {useState, useEffect, useRef, KeyboardEvent} from 'react';
+import React, {useState, useEffect, useRef, KeyboardEvent, useCallback} from 'react';
 import {connect, ConnectedProps} from 'react-redux';
-import {useParams} from 'react-router-dom';
-import styles from './index.module.scss';
-import {sendMessages} from '../../../actions/messages';
-import TemplateSelector from '../TemplateSelector';
-import 'emoji-mart/css/emoji-mart.css';
+import {withRouter, useParams} from 'react-router-dom';
+import {Button} from '@airyhq/components';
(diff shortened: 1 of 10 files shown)


bitboxer:
This is how it looks:

Screen Shot 2021-03-22 at 15 19 52

frontend/ui/src/pages/Inbox/MessageInput/index.tsx
```diff
@@ -189,54 +196,95 @@ const MessageInput = (props: MessageInputProps & ConnectedProps<typeof connector
...
 
+  const getLastMessageWithSuggestedReplies = useCallback(() => {
+    const lastMessages = props.messages
+      ?.filter((message: Message) => message.senderType == 'source_contact')
+      .slice(props.messages.length - 5)
+      .reverse();
+    return lastMessages?.find(
+      (message: Message) => message.metadata?.suggestions && Object.keys(message.metadata.suggestions).length > 0
+    );
+  }, [props.messages]);
```
chrismatix, line 207: So the idea here is that if the last message does not have suggestions we assume that any of the previous 5 message might have suggestions that will also work? :) 

I don't really have an opinion on whether or not that's good. Just want to see if I understand the idea.

frontend/ui/src/reducers/data/messages/index.ts
```diff
@@ -25,6 +26,63 @@ function mergeMessages(oldMessages: Message[], newMessages: Message[]): Message[
   return sortBy(messages, message => message.sentAt);
 }
 
+const findConversationId = (state: Messages, messageId: string) => {
```
chrismatix, line 29: Ok this is a funny consequence of the data model I had not anticipated. This means that the update becomes `O(n)` compared to `O(1)` compared to the other stores.

chrismatix: But I don't have a better suggestion other than populating a separate field on the store with a lookup table to amortize the cost

chrismatix (approved):
Good work!

frontend/ui/src/pages/Inbox/MessageInput/index.tsx
bitboxer: Yes, it was outlined in the ticket this way.

frontend/ui/src/reducers/data/messages/index.ts
bitboxer: ...or just adding the conversation id in the websocket message. That would make this also a `O(1)` thing again.

PR #1329, opened 2021-03-22
[#1267] Created generic logo component

thorstenairy:


AudreyKj (approved):
well done! looks 🔥

frontend/ui/src/components/SourceLogo/index.tsx
```diff
@@ -0,0 +1,56 @@
+import React, {CSSProperties, SyntheticEvent} from 'react';
+import {ReactComponent as GoogleLogo} from 'assets/images/icons/google_avatar.svg';
```
lucapette, line 2: I think we should rename all of them so that they resemble as close as possible the file name as discussed with you and @AitorAlgorta last week.

In this case though, I would probably renamed the files :) as the components seem to have the right name?

frontend/ui/src/components/SourceLogo/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+type SourceLogoProps = {
+  channel: Channel;
+  style?: CSSProperties;
+  imageUrl?: string;
+};
+
+const fallbackImageUrl = (event: SyntheticEvent<HTMLImageElement, Event>, source: string) => {
+  event.currentTarget.src = `https://s3.amazonaws.com/assets.airy.co/${source}_avatar.svg`;
+  event.currentTarget.alt = 'fallback image';
```
lucapette, line 18:
```suggestion
  event.currentTarget.alt = `${source} fallback image`;
```

frontend/ui/src/components/SourceLogo/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+      case SourceType.twilioWhatsapp:
+        return <WhatsappLogo />;
+      default:
+        return <AiryLogo />;
+    }
+  };
+
+  return (
+    <div className={styles.image} style={style}>
+      {channel?.metadata?.imageUrl || imageUrl ? getCustomLogo(channel) : getSourceLogo(channel)}
```
lucapette, line 51: the channel can't be nil right? so `channel?` shouldn't be needed

lucapette (changes requested):
I left a few minor comments.

I like the introduction of this component but I would simplify the API. The `SourceLogo` component doesn't need a channel, it needs a source and an optional

Issue #1312, opened 2021-03-21
Add Users List Endpoint


steffh:
As an API User trying to see all users of my Airy instance,
I want to send an authenticated request to a new endpoint such as `/users.list`,
so that I receive a response with a list containing information about all users in my Airy instance

### Example Request

POST `/users.list`

### Example Response

```JSON5
[
  { 
    "id": "3fba4558-465f-4784-9a40-0ff9dd9f4164",
    "first_name": "Hannibal",
    "last_name": "Lecter",
    "email": "hannibalsgourmet@gmail.com",
  },
  { 
    "id": "921de134-e2ae-4ea4-97e5-a1e672ee57cf",
    "first_name": "Clarice",
    "last_name": "Starling",
    "email": "agentcmstarling@gmail.com",
  }
]
```

PR #1333, opened 2021-03-22
[#1312] Add users.list

lucapette:


backend/api/auth/src/main/java/co/airy/core/api/auth/controllers/UsersController.java
```diff
@@ -161,50 +156,23 @@ private String getResetToken(String userId) {
...
-                .build();
+        ListResponsePayload listResponsePayload = new ListResponsePayload();
 
-        userDAO.insert(user);
+        listResponsePayload.setData(users
+                .stream()
+                .map(u -> UserPayload.builder()
+                        .firstName(u.getFirstName())
+                        .lastName(u.getLastName())
+                        .token(null)
```
chrismatix, line 170: Just for my understanding: Do we have to set this to null explicitly?

chrismatix (approved):

backend/api/auth/src/main/java/co/airy/core/api/auth/controllers/UsersController.java
lucapette: not that I know, I thought the clarity wouldn't hurt though :)

If this grows again, it's probably not good to use this `UserPayload` for all the responses... I thought it was a little overkill right now and that's my way of "documenting" it. (I can/should leave a comment maybe?)

PR #1340, opened 2021-03-23
[#1298] MessageTextArea in inbox doesn't shrink…

### LINKED ISSUE #1298: MessageTextArea in inbox doesn't shrink when extended it before with a long text
Screenshot 2021-03-19 at 10 03 44

bitboxer:
…when extended it before with a long text

closes #1298


ogbeche77 (approved):
LGTM

PR #1350, opened 2021-03-23
[#1097] Allow users to fetch a chat plugin resume token

### LINKED ISSUE #1097: Receive Resume Token for Chat Plugin via the Webhook
As a developer consuming events from the webhook,
having an Airy chat plugin running on one of my websites at the same time,
whenever an end user is in a conversation via the chat plugin on my website,
then I want to receive a resume token via the webhook together with incoming messages,
so that I can send the end user a notification (e.g. via SMS or email) including the resume token,
so that the relevant end user can continue the conversation 
(i.e. without relying on a cookie being there in the browser on the device that started the conversation)

Alternative solution: 

If you don't want to broadcast such resume token via the webhook for security reasons, please provide a secure route instead that I could use to query with an authenticated request for a resume token to represent a participant for a relevant conversation_id.

Additional requirement:

Some kind of documentation on how to use the resume token with the chat plugin (how to inject it via the url, etc.) so that we can lead the relevant end user back to the conversation.

chrismatix:
resolves #1097

### DIFF
--- a/backend/sources/chat-plugin/src/main/java/co/airy/core/chat_plugin/config/AuthConfig.java
+++ b/backend/sources/chat-plugin/src/main/java/co/airy/core/chat_plugin/config/AuthConfig.java
@@ -5,6 +5,7 @@ import org.springframework.context.annotation.Configuration;
 import org.springframework.core.env.Environment;
 import org.springframework.security.config.annotation.method.configuration.EnableGlobalMethodSecurity;
 import org.springframework.security.config.annotation.web.builders.HttpSecurity;
+import org.springframework.security.config.annotation.web.builders.WebSecurity;
 import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity;
 import org.springframework.security.config.annotation.web.configuration.WebSecurityConfigurerAdapter;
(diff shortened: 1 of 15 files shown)


backend/sources/chat-plugin/src/main/java/co/airy/core/chat_plugin/config/AuthConfig.java
```diff
@@ -35,11 +36,15 @@ protected void configure(final HttpSecurity http) throws Exception {
...
                 .addFilter(new JwtAuthenticationFilter(authenticationManager(), jwt))
                 .authorizeRequests()
-                .antMatchers("/actuator/**", "/ws.chatplugin").permitAll()
-                .mvcMatchers("/chatplugin.authenticate").permitAll()
                 .anyRequest().authenticated();
     }
 
+    @Override
+    public void configure(WebSecurity web) {
+        web.ignoring()
```
chrismatix, line 44: So TIL that this is how you really open up routes. `permitAll` still runs the authentication middleware but it allows any principal to pass.

paulodiniz (approved):

Issue #1318, opened 2021-03-22
Core Components: Change title, Add Explainer Image and TLDR


sgarroum:
Change Title from "Core Components" to "Airy Core Components"

Add TLDR:
Airy Core comes with all the components you need for a fully-featured conversational platform

![Airy_Explainer_Components_docs](https://user-images.githubusercontent.com/12533283/111981351-dafd5700-8b07-11eb-9940-04955bf1fdc1.png)

The Airy Core platform contains the following core components: -> Airy Core contains the following components:

lucapette:
<img width="1277" alt="Screenshot 2021-03-22 at 12 31 43" src="https://user-images.githubusercontent.com/124274/111983606-9d4dfd80-8b0a-11eb-9ec2-c2e5719442c3.png">

<img width="1221" alt="Screenshot 2021-03-22 at 12 31 51" src="https://user-images.githubusercontent.com/124274/111983612-9fb05780-8b0a-11eb-8a5b-62e794cf18f0.png">


It looks good in dark mode as his... but with the light theme the background is really heavy. Probably a transparent background image would be already better (even though I'm not so sure about the font...)

AitorAlgorta:
We could just turn the background transparent and then it just uses the background of the page. No need of changing the image depending on the mode. @lucapette

sgarroum:
I created a Lightmode Version with transparent BG & Dark Font, so we have the best of @AitorAlgorta & @lucapette's worlds

![Airy_Explainer_Highlevel_Lightmode](https://user-images.githubusercontent.com/12533283/112179575-17f64600-8bfb-11eb-8a6f-e2fba31a4698.png)

sgarroum:
I also made a lightmode version with white boxes that fits better:
![Airy_Explainer_Highlevel_Lightmode_Whiteboxes](https://user-images.githubusercontent.com/12533283/112310698-2a2cbe80-8ca5-11eb-83e5-d3c46b757a7b.png)

PR #1360, opened 2021-03-24
[#1318] Improve components page

lucapette:


docs/src/components/Image/index.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import useThemeContext from '@theme/hooks/useThemeContext';
+import useBaseUrl from '@docusaurus/useBaseUrl';
+
+const Image = ({path}) => {
```
chrismatix, line 5: This name seems a bit generic. `ContrastImage`, `ThemeImage`, `DarkModeImage`

Either that or we keep it as `Image` and make the props explicit. I.e. `{dark,light}ModePath=`

What do you think? :)

chrismatix: I think it would prefer the second option since currently it also forces the use of `png`

lucapette: I did not have a good name/idea for this so I "made it work" and hoped for such a comment.

So I like this:

```js
<Image lightModePath="..." darkModePath="..." />
```

it's a little verbose but maybe that's a good thing. It's also more flexible than the current component. I'd argue I should keep both props required so that it's always obvious which image is used where?

chrismatix: Yes I agree with that!

lucapette: will change!

chrismatix (approved):

Issue #1359, opened 2021-03-24
typescript does not find important errors


bitboxer:
This error in `lib/typescript/render/index.tsx` was not making our build red:

<img width="373" alt="Screen Shot 2021-03-24 at 13 53 54" src="https://user-images.githubusercontent.com/56195/112315300-4c750b00-8caa-11eb-8680-e78de6d61683.png">

And we need to figure out why.

PR #1364, opened 2021-03-24
[#1326] Move Carousel to components lib

### LINKED ISSUE #1326: Move Carousel to components lib
We have a `Carousel` component that is part of our `render` library. We need to move that to our components library.

bitboxer:
closes #1326


bitboxer:
This moves the `Carousel` into the `components` library. 

Also, it now figures out that the last component displayed is not fully shown and will only scroll to a position where it is visible. Before today, it didn't care and scrolled right pass it 😆 

https://user-images.githubusercontent.com/56195/112336277-3e7cb580-8cbd-11eb-9045-d2f4bd9e5007.mov

frontend/ui/src/pages/Inbox/MessageInput/index.tsx
```diff
@@ -199,7 +199,7 @@ const MessageInput = (props: MessageInputProps & ConnectedProps<typeof connector
   const getLastMessageWithSuggestedReplies = useCallback(() => {
     const lastMessages = props.messages
       ?.filter((message: Message) => message.senderType == 'source_contact')
-      .slice(props.messages.length - 5)
+      .slice(-5)
```
bitboxer, line 202: This was a stupid mistake 🤦

frontend/ui/src/pages/Inbox/Messenger/MessengerContainer/index.tsx
```diff
@@ -32,7 +32,7 @@ const MessengerContainer = ({conversations, currentConversation}: MessengerConta
       ) : (
         <div className={styles.messageDisplay}>
           <MessageList />
-          {currentConversation && <MessageInput Source={currentConversation.channel.source as Source} />}
+          {currentConversation && <MessageInput source={currentConversation.channel.source as Source} />}
```
bitboxer, line 35: I created another ticket for this: #1359

frontend/ui/src/pages/Inbox/MessageInput/index.tsx
chrismatix: This is a separate point, but if I re-read the ticket correctly this code also doesn't quite do what is asked.

The difference is subtle, but the original ticket read to me like they want the suggestions to be the union of all the suggestions of the previous 5 messages. Not the first one that has any.

chrismatix (approved):
lgtm with minor comment

PR #1393, opened 2021-03-26
[#1370] Use shellcheck rule from bazel-tools

### LINKED ISSUE #1370: Use shellcheck rule from bazel-tools
We should use https://github.com/airyhq/bazel-tools/blob/main/lint/shellcheck.bzl instead of our current shellcheck which is installed with apt-get

pascal-airy:
Resolves #1370


BUILD
```diff
@@ -28,6 +29,13 @@ multirun(
     visibility = ["//visibility:public"],
 )
 
+shellcheck(
+    scrs = glob([
```
chrismatix, line 33: This doesn't work, because you cannot access files in other packages using `glob`. I.e. this glob will only return the script files at the root package. You need to add the shellcheck rule to every package that has script files.

pascal-airy:
It's not working as expected will close for now and test again next week

Issue #1357, opened 2021-03-24
Rename the draft release with release.sh


pascal-airy:
In order to make the release process smoother, we should automate as much as possible including renaming the release which is important for the [changelog](https://airy.co/docs/core/changelog)

- [ ] rename to x.y.z
- [ ] tag with x.y.z

PR #1394, opened 2021-03-26
[#1357] Rename the draft release with release.sh

pascal-airy:


scripts/release.sh
```diff
@@ -89,16 +89,28 @@ merge_develop() {
...
 }
 
+rename_draft_release() {
+    release_id=$(
+        command curl -H "Authorization: token ${GITHUB_TOKEN}" -X GET \
+            https://api.github.com/repos/airyhq/airy/releases | jq '.[0].id'
+    )
+    command curl -H "Authorization: token ${GITHUB_TOKEN}" -X PATCH \
+        https://api.github.com/repos/airyhq/airy/releases/"$release_id" \
+        -d "{\"name\":\"${release_number}\", \"tag_name\":\"${release_number}}"
```
pascal-airy, line 99: This will actually create a new tag on the default branch of the repo. Because when we run this the tag won't be there yet. 
Screenshot 2021-03-26 at 12 47 02

pascal-airy: not sure what the full implications of this are. Should we remove `git create tag` and just keep this here in the curl?

chrismatix (approved):

PR #1398, opened 2021-03-26
[#1383] Change size of message bubble

### LINKED ISSUE #1383: Change size of message bubble
Screenshot 2021-03-25 at 17 07 20

It looks kinda odd that we use 100% width for the message bubble, we should set the max-width to anything around 50%!?

ogbeche77:
Fixes #1383

Screen Shot 2021-03-26 at 14 35 13


AitorAlgorta (approved):
LGTM 🚀

PR #1414, opened 2021-03-29
Remove Airy init and restructure cli

### LINKED ISSUE #1051: Remove airy init

chrismatix:
resolves #1051 

Check out the issue for a detailed list of changes.


cli/integration/BUILD
```diff
@@ -8,6 +8,13 @@ go_library(
         "runner.go",
         "test_file.go",
     ],
+    embedsrcs = [
```
chrismatix, line 11: > we put "golden files" in the temp dir
We can discuss how best to handle the golden files. I did not understand what would be the benefit of adding them to the temp config dir. Embedding them makes handling the paths a bit easier. 

@lucapette Could you help me understand the reasoning behind the quoted point a bit more? :)

lucapette: because right now it changes the git status

cli/pkg/providers/aws/aws.go
```diff
@@ -20,23 +21,29 @@ import (
...
 }
 
-func (a *Aws) GetHelmOverrides() []string {
+func New(w io.Writer) *provider {
+	return &provider{
+		w: w,
+	}
+}
+
+func (a *provider) GetHelmOverrides() []string {
```
lucapette, line 42: then you want to call it `p` :)

cli/pkg/providers/minikube/minikube.go
```diff
@@ -18,20 +19,27 @@ const (
...
 }
 
-func (m *Minikube) GetHelmOverrides() []string {
+func New(w io.Writer) *provider {
+	return &provider{
+		w: w,
+	}
+}
+
+func (m *provider) GetHelmOverrides() []string {
```
lucapette, line 33: then you want to call this one `p`

cli/pkg/workspace/init.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	for _, entry := range entries {
+		if err := recCopy(path, "template", entry); err != nil {
+			return ConfigDir{}, err
+		}
+	}
+
+	// Init viper config
+	err = viper.WriteConfigAs(filepath.Join(path, cliConfigFileName))
```
lucapette, line 74:
```
	if err != nil {
		return ConfigDir{}, err
	}
```

and then 

```
return ConfigDir{Path: path}, nil
```

would be more precise no? (as in you deal with the error right away)

docs/docs/cli/installation.md
```diff
@@ -129,15 +129,6 @@ Make sure the output matches the version number you expect:
...
 
-### Step 4: Initialize the setup
-
-The [airy init](cli/reference.md#init) will create a `cli.yaml` file which stores
-your `apiHost` and `apiJwtToken`:
-
-```bash
-airy init
-```
-
```
lucapette, line 140: 😍

lucapette: but there's some sort of TOC in this file so we must remove more

lucapette (commented):
Looking great!

cli/pkg/workspace/init.go
chrismatix: I will move the three commands in between above this one so the immediate return makes more sense :)

lucapette (approved):
:shipit:

PR #1424, opened 2021-03-31
[#1244] Display lightbulb icon for previous…

### LINKED ISSUE #1244: Display lightbulb icon for previous suggested replies
### Requirements

As a means of offering the ability to access suggested replies for previous messages, we should display a lightbulb icon on the righthand side of previous messages that have suggested replies. This requirement should not be limited to a certain amount of messages within the conversation, but rather all messages that contain suggested replies.

We can show the suggested replies on top of the input bar if one of the last 5 messages has suggested replies. If none of last 5 messages contain suggested replies, we will only show the lightbulb icon.

### Design Example

106909707-d66c1300-6700-11eb-8b3a-9c49618fbd3c

In the design above, the lightbulb would be shown to the righthand side of the first message ("Hey, when are you open?"), and the suggested replies prompt would already be available on top of the input bar for the most recent message.

### Design Notes

Please use [this design](https://icons.getbootstrap.com/icons/lightbulb/) for the lightbulb icon.

bitboxer:
…suggested replies

closes #1244


bitboxer:
https://user-images.githubusercontent.com/56195/113101951-9e171b80-91fd-11eb-94a3-7f4b57bbd308.mov

frontend/ui/src/reducers/index.ts
```diff
@@ -5,17 +5,20 @@ import * as authActions from '../actions/user';
...
 import data, {DataState} from './data';
+import ui, {UIState} from './ui';
 
 export * from './data';
 
 type Action = ActionType<typeof authActions>;
 
 export type StateModel = {
   data: DataState;
+  ui: UIState;
```
bitboxer, line 16: This PR adds a UI reducer to make it possible to change the ui state from different parts of the app.

chrismatix: What happens if I switch conversations? I.e. wouldn't the state persist so that you see the same suggestions on the message input of the other conversation?

Alternatively, this state could be handled in the `MessengerContainer`, because `MessageInput` and `MessageList` are siblings.

bitboxer: Sorry, i had the feeling I forgot something 🤦 . Thanks for finding that one. Yes, it could be in the `MessageContainer`, but that would mean sending callbacks and states around. But will do that now 😉

bitboxer: Ah and because my message was misleading: the overlay closes when clicking anywhere, so switching between chats would have worked flawlessly. Just the statemachine was a too much if one callback could do it.

AitorAlgorta (approved):
LGTM 🚀

AitorAlgorta (approved):
🚀

Issue #929, opened 2021-02-09
Airy Live Chat: End Chat option


lizhutter:
### Summary

To enhance the Airy Live Chat experience, we should offer contacts the option to end the current chat permanently. We could offer this option with an X at the top righthand side of the chat, next to a (-) symbol as a means of minimizing the conversation. If a contact chooses to click this button, the next interaction with this user will create an entirely new conversation.
![wine com end chat](https://user-images.githubusercontent.com/74183162/107388116-c6d03e00-6af5-11eb-8cd5-09ea4d168daf.png)

### Design Notes

These images are only for illustrative purposes - this feature does not need to be built to look identical to this.

lucapette:
What's the exact UI interaction we expect to happen when someone clicks the button?

lizhutter:
When someone clicks the button, we can prompt something like the following image:

<img width="341" alt="Screenshot 2021-02-12 at 15 53 17" src="https://user-images.githubusercontent.com/74183162/107786338-11e48e00-6d4e-11eb-93a3-d6abeae7d759.png">

And if the contact clicks "`End`", we can display a message saying something like "_Your conversation has ended. Thank you for chatting with us today_."

Below this, we can also offer the option to `Start a New Conversation` if the contact would like to.

lizhutter:
To reiterate, the X will prompt the contact to end the chat, and the minus symbol will prompt the contact to minimize the chat. Please use the minus symbol I've attached from [this website](https://www.zwicon.com/cheatsheet.html) for the design.

![Screenshot 2021-03-23 at 18 13 16](https://user-images.githubusercontent.com/74183162/112188656-76272700-8c03-11eb-919b-1e01784f62ec.png)

bitboxer:
Also: what will the agent see in the app? Do we send a special crafted message type saying "conversation ended" / "conversation restarted" ?

lizhutter:
I think that would make the most sense in this situation. This way, the agent is aware of how to approach the conversation from that point forward (whether to treat it as an extension of the current conversation or the beginning of a new conversation).

bitboxer:
As discussed in #1436 we might want to change the "auto open" setting if the chat has ended. So don't open the chat automatically when opening the page and the chat is defined as closed.

PR #1436, opened 2021-03-31
Fix/1413 expand chat plugin by default

### LINKED ISSUE #1413: Expand Chat Plugin by default
Currently, the chat plugin is collapsed/minimized by default. We should create a custom behavior for returning to conversations so that the chat plugin is expanded by default.

For example: when a contact clicks on the link provided to resume the conversation, it would be ideal if the plugin would already be open and ready, so that the contact does not have to click once again to open the chat and begin typing.

AudreyKj:
closes #1413 

The chat is expanded when there is a resume token, it is hidden when there is no resume token.


bitboxer:
🤔  Then it is always open as soon as I once talked to someone, right? Might be a bit annoying to go back to a page 3 months later and see that old message pop up .

AudreyKj:
@bitboxer yeah the chat will appear as open as soon as the conversation started. @lizhutter what do you think?

bitboxer:
Asking because there is no way for a customer to end the chat and this means we can't not open the popup.

lizhutter:
@AudreyKj @bitboxer There is currently a ticket for end chat (https://github.com/airyhq/airy/issues/929) that needs to be discussed, but it will help with this situation once it's out of discussion and has been worked on.

Issue #1304, opened 2021-03-19
Wait for all the necessary resources before printing status on create


ljupcovangelski:
When we run `airy create` the command prints the status after the helm chart completes. However, some of the components might not be running yet.

Components which we should consider:
- ui
- chatpluin / source-chatplugin
- api-communication
- api-auth

This will also concern the future improvement of the `airy status` command.

PR #1442, opened 2021-03-31
[#1304] Wait for core components during create

ljupcovangelski:
Four jobs to wait for the core components, before releasing the helm hook. This way, when `airy create` finishes, we ensure to have a running system.


lucapette (approved):
:shipit:

PR #1456, opened 2021-04-01
[#1111] Customize Chat Plugin

### LINKED ISSUE #1111: Introduce themeConfig to the chatplugin
Our current api for the widget is somewhat too polarized toward both ends of the spectrum "plug&play" vs "full autonomy for customisation".

The idea is to introduce a _very limited_ set of options you can pass to the widget so that you can start off with a plug&play integration and move to render props only if you have very specialised needs for the widget rendering. 

Here's the proposed list of options:

- `welcomeMessage` - The automatic message the widget sends out once someone opens it
- `primaryColor` - what we use for the background
- `bubbleIcon`
- `exitButtonIcon`
- `sendButtonIcon`

I think we may need a bit more about colors/fonts to make it usable (probably a secondary color at least)

AitorAlgorta:
closes #1111


frontend/chat-plugin/src/App.tsx
```diff
@@ -1,50 +1,78 @@
...
+          <Chat channelId={channelId} config={config} />
         ) : (
           <span style={{color: 'red'}}>Widget authorization failed. Please check your installation.</span>
         )}
       </div>
     );
   }
 }
+
+export type CustomConfig = {
```
lucapette, line 34: I'm not sure Custom adds anything here. I think I'd be ok with `Config` (but maybe I don't have the right context for the naming)

frontend/chat-plugin/src/components/chat/index.tsx
```diff
@@ -120,17 +122,29 @@ const Chat = (props: Props) => {
 
   const headerBar = props.headerBarProp
     ? () => props.headerBarProp(ctrl)
-    : () => <AiryHeaderBar toggleHideChat={ctrl.toggleHideChat} />;
+    : () => <AiryHeaderBar toggleHideChat={ctrl.toggleHideChat} config={props.config} />;
```
lucapette, line 125: config should do right? (also in the rest of the file)

thorstenairy (dismissed):
LGTM

lucapette (dismissed):
Minor comments but looking good!

pascal-airy (approved):

lucapette (approved):

PR #1458, opened 2021-04-01
[#572] Simplify senderType

### LINKED ISSUE #572: Simplify senderType
As it's a binary information at first (either the contact sent a message or they didn't), we should reflect this in the data model.

the nuanced data points (which are mostly source specific) should travel through the system via the headers or the metadata

paulodiniz:
Fixes #572 

Screenshot from the chatplugin

Screenshot 2021-04-07 at 14 39 48

Screenshot from the inbox

Screenshot 2021-04-07 at 14 39 54

### DIFF
--- a/backend/model/message/src/main/java/co/airy/model/message/MessageRepository.java
+++ b/backend/model/message/src/main/java/co/airy/model/message/MessageRepository.java
@@ -2,7 +2,6 @@ package co.airy.model.message;
 
 import co.airy.avro.communication.DeliveryState;
 import co.airy.avro.communication.Message;
-import co.airy.avro.communication.SenderType;
 import co.airy.model.metadata.dto.MetadataMap;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
@@ -23,11 +22,11 @@ public class MessageRepository {
 
     // In preparation for https://github.com/airyhq/airy/issues/572
     public static boolean isFromContact(Message message) {
-        return message.getSenderType().equals(SenderType.SOURCE_CONTACT);
+        return message.getIsFromContact();
     }
 
     public static boolean isFromAiry(Message message) {
-        return message.getSenderType().equals(SenderType.APP_USER);
+        return !message.getIsFromContact();
     }
 
     public static Object resolveContent(Message message) {
--- a/backend/sources/facebook/events-router/src/main/java/co/airy/core/sources/facebook/MessageParser.java
+++ b/backend/sources/facebook/events-router/src/main/java/co/airy/core/sources/facebook/MessageParser.java
@@ -60,7 +60,6 @@ public class MessageParser {
                 headers.put("postback.payload", "__empty__");
             }
         }
-        boolean isFromContact = !isEcho;
 
         Optional.ofNullable(postbackNode)
                 .map(node -> node.get("referral"))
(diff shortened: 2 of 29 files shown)


backend/model/message/src/main/java/co/airy/model/message/MessageRepository.java
```diff
@@ -23,11 +23,12 @@ public static boolean isNewMessage(Message message) {
...
     // In preparation for https://github.com/airyhq/airy/issues/572
     public static boolean isFromContact(Message message) {
-        return message.getSenderType().equals(SenderType.SOURCE_CONTACT);
+        return message.getIsFromContact();
     }
 
     public static boolean isFromAiry(Message message) {
-        return message.getSenderType().equals(SenderType.APP_USER);
+        return message.getHeaders().get("message.from").equalsIgnoreCase("airy");
+        //return message.getSenderType().equals(SenderType.APP_USER);
```
paulodiniz, line 31: do we need something like this or can we do 

```java
return !message.getIsFromContact();
```

backend/sources/facebook/events-router/src/main/java/co/airy/core/sources/facebook/MessageParser.java
```diff
@@ -44,42 +43,35 @@ String getSourceConversationId(final JsonNode webhookMessaging) throws NullPoint
...
 
         Optional.ofNullable(postbackNode)
                 .map(node -> node.get("referral"))
                 .ifPresent(referralNode -> headers.put("postback.referral", referralNode.toString()));
 
         return Message.newBuilder()
                 .setContent(payload)
-                .setSenderType(senderType)
                 .setSenderId(senderId)
+                .setIsFromContact(isFromContact)
```
lucapette, line 72: maybe just taste but if the variable is only used here I'd inline the condition

lib/typescript/httpclient/model/Content.ts
```diff
@@ -8,6 +8,10 @@ export interface Content {
...
 export type RenderedContentUnion = Message | Template | Content;
 
-export function isFromContact(message: RenderedContentUnion) {
-  if (message && 'senderType' in message) return message?.senderType === SenderType.sourceContact;
+export function isFromContact(content: RenderedContentUnion) {
+  if (content && 'fromContact' in content) {
+    return content?.fromContact;
+  } else {
+    return false;
+  }
```
lucapette, line 16: can do early return here.

Even though I think this function is odd (as it takes a renderedcontentunion but seems message specific anyway. I think @AitorAlgorta and I will address it in #1466)

lib/typescript/httpclient/model/Message.ts
```diff
@@ -16,15 +16,9 @@ export enum MessageState {
...
-export enum SenderType {
-  sourceContact = 'source_contact',
-  sourceUser = 'source_user',
-  appUser = 'app_user',
-}
-
 export interface Message extends Content {
   deliveryState: MessageState;
-  senderType: SenderType;
+  fromContact: boolean;
```
lucapette, line 21: 😍

backend/sources/twilio/events-router/src/main/java/co/airy/core/sources/twilio/EventsRouter.java
```diff
@@ -68,7 +67,7 @@ public void startStream() {
                                 messageId,
                                 Message.newBuilder()
                                         .setSenderId(sourceConversationId)
-                                        .setSenderType(SenderType.SOURCE_CONTACT)
+                                        .setIsFromContact(true) //is this true
```
lucapette, line 70: yes! We should remove the comment though

lucapette (dismissed):
Minor comments

Good stuff 👏

lucapette (approved):
:shipit:

PR #1475, opened 2021-04-07
added showmode flag that blocks functionality in chat plugin

### LINKED ISSUE #1474: ShowMode in Chat Plugin
We want to have a `showMode` in the chat plugin so we can use it in the customization section in the `UI` and in the `docs` without creating conversations. 

The idea is that this `showMode` displays all the functionalities of the chat plugin but doesn't trigger any action. No authentication, no send messages, no websocket connection.

AudreyKj:
closes #1474 

- showMode flag added in Chat Plugin's config 
- when set to true, showMode disables the chat's functionality (default is false)


thorstenairy (approved):
LGTM

PR #1478, opened 2021-04-07
[#1004] Enable quickreplies for chatplugin

### LINKED ISSUE #1004: Quick Replies for Chat Plugin
Text + Suggestions (buttons) currently don’t create buttons in neither the chat plugin nor messaging UI.

As an example, it would be a good idea render a button below “some text” here:
screenshot_2021-02-16_at_18 14 54

### Suggestion Requirements

- A row of suggested replies that appear as a row of suggestion chips following the associated message. Maximum 13 suggestions, with a maximum of 25 characters each.

- The chips only display when the associated message is the most recent message within the conversation (including both system user and contact messages). The contact can tap a suggested reply to send the text reply to the system user.

- The chips will disappear when one has been clicked, so it is an ephemeral element of a conversation.

- These suggestions (buttons) should be displayed horizontally below the plain text that prompts them.

**Note**: Chips are compact elements that represent an input, attribute, or action. Chips allow users to enter information, make selections, filter content, or trigger actions.

### Inspiration:
image

### Payload
```
{
   "conversation_id": "<CONVERSATION_ID>",
   "message": {
       "text": "Pick a color:",
       "quick_replies":[
          {
           "content_type":"text",
           "title":"Red",
           "payload":"<POSTBACK_PAYLOAD>"
          },{
           "content_type":"text",
           "title":"Green",
           "payload":"<POSTBACK_PAYLOAD>"
          }    
       ]
   }
}
```

### LINKED ISSUE #1477: Gap between message and border at the last message of app_user
Screenshot 2021-04-07 at 16 19 40

thorstenairy:
closes #1004
closes #1477


AitorAlgorta (dismissed):
🚀

AitorAlgorta (approved):

Issue #1290, opened 2021-03-18
Prometheus Metrics about Spring apps 


pascal-airy:
The metrics provided by [Spring Metrics](https://docs.spring.io/spring-metrics/docs/current/public/prometheus) looks to be a very good start

pascal-airy:
We also want to time how long requests take which Spring Metrics can't do. @lucapette mentioned an HTTP controller which could also be used to expose those metrics if I got that right

PR #1479, opened 2021-04-07
[#1290] Prometheus Metrics about Spring apps

pascal-airy:
- `/actuator/prometheus` endpoint to all Spring apps
- Prometheus Operator Service Monitor configuration to auto-discover all spring apps targets
- Label for spring apps services to make them discoverable by the Service Monitor


infrastructure/helm-chart/charts/apps/charts/api/charts/api-admin/templates/service.yaml
```diff
@@ -3,9 +3,12 @@ kind: Service
 metadata:
   name: api-admin
   namespace: {{ .Values.global.namespace }}
+  labels:
+    type: spring
```
ljupcovangelski, line 7: Perhaps we should standardize these under the `core.airy.co` domain...

Maybe
```
core.airy.co/type: spring
```
is a better option?

Also we have already `    core.airy.co/managed: "true"`

ljupcovangelski (approved):
Looks good :+1:

PR #1488, opened 2021-04-08
[#1466] Extract model lib from httpclient

lucapette:


AitorAlgorta (dismissed):
🚀

thorstenairy (approved):

lib/typescript/render/BUILD
```diff
@@ -7,9 +7,9 @@ package(default_visibility = ["//visibility:public"])
 ts_web_library(
     name = "render",
     deps = [
-        "//frontend/assets",
+        "//frontend/assets", #TODO FIX ME
```
chrismatix, line 10: It's not immediately obvious what needs to be fixed here (I get that this lib should not depend on assets, but I'd still spell it out)

lucapette: true that. I left it to myself as a reminder so that I can look into it (as of now I don't actually know why it is like that and what's the best action to take yet)

AitorAlgorta (approved):

PR #1506, opened 2021-04-09
Move back components to the mono repo

lucapette:
In the context of this pull requests, we did a few things:

- `lib/typescript/components` contains the _used_ parts of airyhq/components
- `frontend/assets` was moved `lib/typescript/assets` (the previous dep graph was strange. One could argue most of these libraries should instead be in `frontend/lib`. It's a simple change so we can do it if we like that more)
- all svgs have been cleaned up with https://github.com/svg/svgo with the command `svgo -f images -r --multipass` Done in [one commit](https://github.com/airyhq/airy/commit/e945981fe7917ec1795406dcb439a9176f07821b) so you can check the result of that in isolation

While doing this, we noticed more things we can move around/clean up/delete. But this pr is already very noisy so, for the sake of visibility, we'll do follow up pull requests


lib/typescript/components/cta/Button/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+  onClick?: (event?) => void;
+  type?: 'submit' | 'button' | 'reset';
+  disabled?: boolean;
+  styleVariant?: styleVariantType;
+  tabIndex?: any;
+  dataCy?: string;
+};
+
+export const Button = ({children, onClick, type, styleVariant, disabled, tabIndex, dataCy}: ButtonProps) => {
+  const styleFor = (variant: styleVariantType) => {
```
chrismatix, line 17: This variable can be moved outside of the method

lib/typescript/components/inputs/Dropdown/index.tsx
```diff
@@ -0,0 +1,81 @@
+import React, {useRef, useState, useEffect, useCallback} from 'react';
+import styles from './style.module.scss';
+import {ReactComponent as ChevronDown} from 'assets/images/icons/chevron-down.svg';
+
+export const Dropdown = ({text, options, variant, onClick}) => {
+  const [dropdownVisible, setDropdownVisible] = useState(false);
+  const componentEl = useRef(null);
+  const buttonEl = useRef(null);
+
+  const styleFor = variant => {
```
chrismatix, line 10: nitpick: This can also be moved outside.

lib/typescript/components/inputs/Input/style.module.scss
```diff
@@ -0,0 +1,182 @@
...
+
+  &:focus-within {
+    background-color: var(--color-background-blue);
+    border: 1px solid var(--color-airy-blue);
+    outline: none;
+  }
+}
+
+.inputInvalid {
+  .inputTitle {
```
chrismatix, line 29: nitpick: Since there is only one input title I think this nesting in `inputValid` is unnecessary.

lib/typescript/components/inputs/SearchField/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+  value: string;
+  setValue: (value: string) => void;
+  resetClicked?: () => void;
+  autoFocus?: boolean;
+  dataCy?: string;
+};
+
+export const SearchField = ({id, placeholder, value, setValue, resetClicked, autoFocus, dataCy}: Props) => {
+  const inputRef = createRef<HTMLInputElement>();
+  const resetButton = useCallback(() => {
```
chrismatix, line 19: I don't think `useCallback` is necessary here. If it is, then `resetClicked` is missing from the deps list.

lib/typescript/components/inputs/TextArea/index.tsx
```diff
@@ -0,0 +1,75 @@
+import React, {Component} from 'react';
+import TextareaAutosize from 'react-autosize-textarea';
+
+import {Input, InputProps} from '../Input';
+import styles from './style.module.scss';
+
+class TextAreaComponent extends Component<IProps & InputProps> {
+  classForState = currentValidationState => {
```
chrismatix, line 8: nitpick: Can be moved outside

chrismatix (commented):
First pass 👍

AitorAlgorta (approved):
```
🚀 🚀       🚀     🚀    🚀 🚀 🚀 🚀     🚀 🚀 🚀 🚀  
🚀    🚀    🚀     🚀    🚀                    🚀 
🚀      🚀 🚀     🚀     🚀                    🚀🚀🚀 
🚀       🚀 🚀     🚀     🚀                   🚀 
🚀           🚀     🚀     🚀 🚀 🚀 🚀     🚀 🚀 🚀 🚀 
```

lucapette:
As discussed with @chrismatix the comments will serve as TODO list for the follw up prs so we can do further improvements with higher visibility than this giant pr

PR #1508, opened 2021-04-09
[#929]Implement the option to end chat

### LINKED ISSUE #929: Airy Live Chat: End Chat option

ogbeche77:
Screen Shot 2021-04-12 at 15 30 03


frontend/chat-plugin/src/components/modal/index.module.scss
```diff
@@ -6,18 +6,21 @@
...
   z-index: 10;
   display: flex;
   justify-content: center;
   align-items: center;
 
   &:before {
     content: '';
-    width: 98%;
-    height: 73%;
+    width: 97%;
```
AitorAlgorta, line 17: does this shrink properly?

frontend/chat-plugin/src/components/chat/index.tsx
```diff
@@ -210,6 +235,22 @@ const Chat = (props: Props) => {
         </div>
       )}
       <BubbleProp render={bubble} />
+      {showModal && (
```
AitorAlgorta, line 238: why is the modal inside the bubble? 🤔

AitorAlgorta (changes requested):
Let's test if the shrink works because working with those percentages is difficult to see without testing

AitorAlgorta (approved):

Issue #740, opened 2021-01-26
CLI doesn't remove configMaps


ljupcovangelski:
If configuration has been added to the `airy.yaml` file, then the cli creates the appropriate configMaps when running `airy config apply`. 

However, if we then delete the configuration for a particular source, then the configMap of the particular source will not be removed and the pods for the source will continue to run.

Probably the solution would involve refactoring the logic and the configuration.

ljupcovangelski:
In order to delete a configMap when a configuration is removed, we must have a refference point where all the components of the Airy platform are listed, so that we can iterate over them to sync the configuratio between the config file and the kubernetes cluster.

One way would be to hardcode them in the cli, for a particular version.
Another way would be to hardcode the components inside the ServiceDiscovery app, in the backend (as we have at this moment) for the `airy status` cli.

I think a good approach would be to have a configMap mounted on the ServiceDiscovery app, presenting a json configuraiton file, where all the components and their configurationItems are listed.
The ServiceDiscovery app can read from there and the configu=


```
{
  "components": {
    "core": [
      {
        "name": "facebook",
        "type": "source",
        "configItems": {
          "appId": null,
          "appKey": null
        }
      },
      {
        "name": "twilio",
        "type": "source",
        "configItems": {
          "number": 0,
          "twilioKey": null
        }
      },
      {
        "name": "webhook",
        "type": "integration",
        "configItems": "siteName"
      },
      {
        "name": "media-resolver",
        "type": "storage",
        "configItems": {
          "s3Bucket": null,
          "accessKey": null
        }
      }
    ],
    "enterprise": [
      {
        "name": "search",
        "type": "ui",
        "configItems": "key"
      }
    ]
  }
}
```

This way, we would also be able to generate the `airy.yaml` file, as the cli will have from where to fetch all the components and their configuration items.

If we change the configmap on the fly, we need to reload the serviceDiscovery app.
Also we can have a convention for the components, for the configMaps and names of the apps - {type}-{name} (source-facebook)

This concept can work at the moment and if we switch to CustomResourcesDefinitions in the future.

lucapette:
I think I like the idea the "only central" place is a config map (something called `components-config`?). That way, as you say, we can refactor the service discovery class in the `client.config` endpoint

ljupcovangelski:
[...]

PR #1510, opened 2021-04-09
[#740] Label and introspect components

ljupcovangelski:
At this moment this is mostly a design concept on how do we structure the data in our cluster, for refactoring `airy config apply` and  `airy status`. I've made the following assumptions:
- Every components is autonomous in the way it defines its own properties (mandatory, configitems, ...).
- Every component can have multiple deployments/statefulsets.
- Every deployment/statefulset defines its own mandatory or optional configItems.
- Properties for the component on which we need to filter are put in labels (ex. mandatory).
- Properties for the component which are strict to the compoent and its deployments/statefulsets are put in annotations (ex. configItems).
Example:
```
  labels:
    core.airy.co/mandatory: "{{ .Values.mandatory }}"
    core.airy.co/component: "{{ .Values.component }}"
  annotations:
    core.airy.co/config-items-mandatory: "FACEBOOK_WEBHOOK_SECRET FACEBOOK_APP_ID FACEBOOK_APP_SECRET"
    core.airy.co/config-items-optional: "FACEBOOK_PAGE_NAME
```

- The sum of all configItems for all the deployments/statefulsets in an a component = the configItems for the component.
- All of the configItems for a component will be put in a single configMap for that component. That is also inline with how the controller works at the moment.
- Once helm instals the `Airy Core` chart, there is one job which gets all the components and their config and puts them in a single configmap.
- We also leave the possibility for non-mandatory components to be deployed by default, for some provider (with the enabled helm parameter).
- Config blocks in the airy.yaml file are related to components, not to the apps inside the components. That is why I think it is more clear having a central reference point what does each component contain.
- Currently the configMap is created with a job running a bash script, but I think in the future we will find much better ways of doing this.


ljupcovangelski:
After discussing about the PR, we opted for another approach and the PR was modified accordingly.
Explanation about the discussion - https://github.com/airyhq/airy/issues/740#issuecomment-817859584

paulodiniz (approved):

Issue #1505, opened 2021-04-09
Rewrite Webhook queue with Beanstalkd


pascal-airy:
As Redis PubSub feature is not suitable for a reliable queue and the current RPlopLPush method is not suited for high throughput I suggest doing a rewrite with https://beanstalkd.github.io/

- [ ] Replace the Redis deployment with a Beanstalk one 
- [ ] Change Webhook Producer to use [Java Client](https://github.com/jpeffer/JavaBeanstalkClient)
- [ ] Change Webhook Consumer to use [Go Client](https://github.com/beanstalkd/go-beanstalk)

PR #1536, opened 2021-04-13
[#1505] Rewrite Webhook queue with Beanstalkd

pascal-airy:


backend/webhook/beanstalk-worker/BUILD
```diff
@@ -1,32 +1,32 @@
-# gazelle:prefix redis-worker
+# gazelle:prefix beanstalk-worker
```
chrismatix, line 1: Good chance to remove the data source from the package name?

lucapette: yes!

pascal-airy: then I think I'd call it `consumer`

backend/webhook/consumer/pkg/worker/worker.go
```diff
@@ -0,0 +1,136 @@
...
+func (t *Task) logError(err error) {
+	log.Println(err)
+	t.errors = append(t.errors, ConsumerError{
+		Err: err,
+		T:   time.Now(),
+	})
+	log.Println(len(t.errors))
+}
+
+func (t *Task) GetErrors() []ConsumerError {
```
pascal-airy, line 134: This function always returns an empty list. What am I missing?

backend/webhook/consumer/pkg/worker/worker.go
```diff
@@ -0,0 +1,135 @@
...
+	T   time.Time
+}
+
+type Task struct {
+	done      chan bool
+	beanstalk *beanstalk.Conn
+	errors    []ConsumerError
+}
+
+type AiryMessage struct {
```
lucapette, line 28: WebhookEvent would reflect the naming we use in other places (even thought the data type is slighly different because of implementation detail. But better name than "message")

backend/webhook/consumer/pkg/worker/worker.go
lucapette: good question :)

backend/webhook/consumer/pkg/worker/worker.go
```diff
@@ -0,0 +1,135 @@
...
+	Headers  map[string]string
+	Body     map[string]interface{}
+}
+
+var backoffSchedule = []time.Duration{
+	0 * time.Second,
+	1 * time.Second,
+	3 * time.Second,
+	10 * time.Second,
+}
```
lucapette, line 39: this works better as a function (I would probably use https://github.com/jpillora/backoff)

backend/webhook/consumer/pkg/worker/worker.go
```diff
@@ -0,0 +1,135 @@
...
+		if err != nil {
+			t.logError(err)
+		} else {
+			break
+		}
+	}
+	t.beanstalk.Delete(id)
+}
+
+func (t *Task) HandleMessage(message string) error {
```
lucapette, line 88: message > event

Also I would change the type back to `[]byte` because we're converting it back and forth but as I understand it there's no reason to do so

backend/webhook/consumer/pkg/worker/worker.go
```diff
@@ -0,0 +1,135 @@
...
+	req, _ := http.NewRequest("POST", data.Endpoint, bytes.NewBuffer(jsonString))
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("User-Agent", "Airy/1.0")
+
+	for k, v := range data.Headers {
+		req.Header.Set(k, v)
+	}
+
+	client := &http.Client{
+		Timeout: 5 * time.Second,
```
lucapette, line 110: 5 may be a bit stringent for some envs. 10 feels like a more reasonable default

backend/webhook/publisher/src/main/java/co/airy/core/webhook/publisher/BeanstalkPublisher.java
```diff
@@ -2,33 +2,34 @@
...
 
-    RedisQueue(RedisTemplate<String, String> redisTemplate) {
-        this.redisTemplate = redisTemplate;
+    BeanstalkPublisher(JobProducer beanstalkdJobProducer) {
+        this.beanstalkdJobProducer = beanstalkdJobProducer;
     }
 
-    void publishMessage(String webhookId, QueueMessage message) {
+
+    void publishMessage(QueueMessage message) {
```
lucapette, line 28: message -> event

since we're touching this :)

backend/webhook/consumer/pkg/worker/worker.go
```diff
@@ -0,0 +1,135 @@
...
+	"net/http"
+	"time"
+)
+
+type ConsumerError struct {
+	Err error
+	T   time.Time
+}
+
+type Task struct {
```
lucapette, line 22: this type doesn't "look like" a task? It\s more the worker it self right?

lucapette (commented):
first pass!

backend/webhook/consumer/pkg/worker/consumer.go
```diff
@@ -0,0 +1,120 @@
...
+type Consumer struct {
+	ready                chan bool
+	worker               *Worker
+	schemaRegistryClient *srclient.SchemaRegistryClient
+}
+
+type Webhook struct {
+	Id       string
+	Endpoint string
+	Headers  map[string]map[string]string
```
pascal-airy, line 28: This is the corresponding Webhook object

backend/webhook/consumer/pkg/worker/worker.go
```diff
@@ -0,0 +1,159 @@
...
+			return
+		default:
+			id, event, err := w.beanstalk.Reserve(1 * time.Minute)
+			if errors.Is(err, beanstalk.ErrTimeout) {
+				continue
+			}
+			if err != nil {
+				log.Println(err)
+				continue
+			}
```
lucapette, line 82: the two if could be squashed together (if not nil { if !timeout {print} continue})

backend/webhook/consumer/pkg/worker/worker.go
```diff
@@ -0,0 +1,159 @@
...
+
+			for {
+				select {
+				case <-ctx.Done():
+					log.Println("terminating worker: context canceled ")
```
lucapette, line 87:
```suggestion
					log.Println("terminating worker: context cancelled")
```

ljupcovangelski (approved):
Looks good :+1:

PR #1546, opened 2021-04-14
[#740] - Adding k8s endpoint to airy controller

paulodiniz:


chrismatix (dismissed):
Very cool!

infrastructure/helm-chart/charts/controller/templates/service.yaml
```diff
@@ -0,0 +1,14 @@
...
+  namespace: {{ .Values.global.namespace }}
+spec:
+  ports:
+    - name: web
+      port: 80
+      targetPort: 8080
+      protocol: TCP
+  type: NodePort
+  selector:
+    app: controller-info
```
ljupcovangelski, line 14: should be `app: airy-controller`

infrastructure/helm-chart/charts/controller/templates/service.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: controller-info
```
ljupcovangelski, line 4: I think that `airy-controller` would be a more appropriate name for the service.

ljupcovangelski (commented):
I can confirm that it works
```
bash-4.4# curl controller-info.default/components
{"components":["api-admin","api-auth","api-communication","api-websocket","media-resolver","source-google","frontend-chat-plugin","frontend-ui","source-chat-plugin","source-facebook","source-twilio","integration-webhook"]}
```
:tada:

ljupcovangelski (approved):
Looks good :+1:

PR #1549, opened 2021-04-14
[#740] Uses components endpoint on service discovery

paulodiniz:


backend/api/admin/src/main/java/co/airy/core/api/config/ServiceDiscovery.java
```diff
@@ -36,13 +27,9 @@ public ServiceDiscovery(@Value("${kubernetes.namespace}") String namespace, Rest
...
     @Scheduled(fixedRate = 1_000)
     private void updateComponentsStatus() {
-        for (String service : services) {
-            try {
-                ResponseEntity<Object> response = restTemplate.exchange(String.format("http://%s.%s/actuator/health", service, namespace), HttpMethod.GET, null, Object.class);
-                components.put(service.replace("-connector", ""), Map.of("enabled", response.getStatusCode().is2xxSuccessful()));
-            } catch (Exception e) {
-                components.put(service.replace("-connector", ""), Map.of("enabled",false));
-            }
+        final ResponseEntity<ComponentsResponsePayload> response = restTemplate.getForEntity("http://airy-controller.default/components", ComponentsResponsePayload.class);
```
chrismatix, line 30: Not saying that this is wrong, but just so I understand it correctly: This map never "forgets" a component, because we only ever append, right? :) (Unless you restart the app of course)

chrismatix (approved):

backend/api/admin/src/main/java/co/airy/core/api/config/ServiceDiscovery.java
lucapette: it is "wrong" :) to be fair a `components.clear()` before the for should do it

lucapette (changes requested):
setting as "request changes" as we have to clear the map before using it

AitorAlgorta (dismissed):
Frontend code looks good :)

chrismatix (approved):

AitorAlgorta (approved):
FE is good

lucapette (approved):

Issue #1517, opened 2021-04-12
Remove email authentication


chrismatix:
First, we want to remove the current authentication system so that we can start the OIDC implementation from a clean slate. 

- [x] Update docs to say that there is no auth by default unless users specify an API key in their `airy.yaml`
- [x] Remove jwt middleware (auth tests should use the API key)
- [x] Remove postgres deployment
- [x] Remove `backend/api/auth` and its deployment/config artifacts
- [x] Set a default for the [senderId](https://github.com/airyhq/airy/blob/84a8366aeb3aa7199be1e458281034cfec4759e1/backend/api/communication/src/main/java/co/airy/core/api/communication/SendMessageController.java#L68) in case there is no auth
- [x] Remove auth from the HTTP client 
- [x] Remove the login / signup views 
- [x] CLI uses the Kubernetes API to get the API key to authenticate requests (if the key is present)

_Note_ The chatplugin jwt middleware remains unchanged since it is concerned with authenticating chatplugin visitors and not Airy Core users

PR #1569, opened 2021-04-16
[#1517] Remove email auth

chrismatix:
addresses #1517


backend/api/communication/src/main/java/co/airy/core/api/communication/SendMessageController.java
```diff
@@ -64,7 +64,8 @@
                 .setHeaders(Map.of())
                 .setDeliveryState(DeliveryState.PENDING)
                 .setSource(channel.getSource())
-                .setSenderId(auth.getPrincipal().toString())
+                // TODO add the oidc id here in https://github.com/airyhq/airy/issues/1518
+                .setSenderId("airy-core-anonymous")
```
chrismatix, line 68: @airyhq/codeowners-backend Even with oidc we will have to keep a default sender id for when there is no authentication configured

frontend/ui/BUILD
```diff
@@ -31,7 +31,6 @@ ts_web_library(
         "@npm//@types/react-redux",
```
chrismatix, line 31: Frontend changes begin here @airyhq/codeowners-frontend

frontend/ui/src/cookies/cookie.ts
```diff
@@ -1,34 +0,0 @@
-export function setCookie(name: string, value: string, domain: string, days = 3650) {
```
chrismatix, line 1: The authentication cookie will be an httpOnly cookie, which means that the frontend does not have direct access and will call `/users.getPrincipal` instead to get the users' information

infrastructure/helm-chart/charts/apps/charts/api/charts/api-admin/templates/deployment.yaml
```diff
@@ -29,8 +29,6 @@ spec:
         image: "{{ .Values.global.containerRegistry}}/{{ .Values.image }}:{{ .Values.global.appImageTag }}"
         imagePullPolicy: Always
         envFrom:
-          - configMapRef:
```
chrismatix, line 32: There is no more `api-config`. All api pods now rely solely on the security config map

frontend/ui/BUILD
AitorAlgorta: 🚀

pascal-airy (approved):
image

lib/typescript/httpclient/README.md
```diff
@@ -4,19 +4,15 @@ The HttpClient Library includes a HTTP client for making requests to Airy's API.
...
-```
-import { HttpClient} from 'httpclient';
-
-const myInstance = new HttpClient(apiUrl, authToken);
-
-myInstance.listChannels()
+```typescript
+import {HttpClient} from "httpclient";
 
+const myInstance = new HttpClient("http://airy.core");
```
lucapette, line 14: since we're at it I'd rename this const to `client`? It reads better (super minor thing I get it:D)

package.json
```diff
@@ -21,7 +21,6 @@
     "react-autosize-textarea": "^7.1.0",
     "react-color": "^2.19.3",
     "react-dom": "16.14.0",
-    "react-facebook-login": "^4.1.1",
```
lucapette, line 24: oh wow :D

lucapette (commented):
Super minor comments. Must have been very satisfying to do this :)

Amazing work 👏👏👏

backend/api/websocket/src/main/java/co/airy/core/api/websocket/WebSocketConfig.java
```diff
@@ -29,10 +29,12 @@
...
     private static final Logger log = AiryLoggerFactory.getLogger(WebSocketConfig.class);
-    private final Jwt jwt;
+    private final String systemToken;
+    private final String systemTokenPrincipal;
 
-    public WebSocketConfig(Jwt jwt) {
-        this.jwt = jwt;
+    public WebSocketConfig(@Value("${systemToken:#{null}}") String systemToken) {
+        this.systemToken = systemToken;
+        this.systemTokenPrincipal = systemToken == null ? null : String.format("system-token-%s", systemToken.substring(0, Math.min(systemToken.length(), 4)));
```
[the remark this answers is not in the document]
chrismatix: Yeah so I was wondering that for a bit, too. If we don't have this here then there is no way to use the websocket programmatically when there is only system token auth.

lucapette (dismissed):
:shipit:

lucapette (dismissed):

paulodiniz (approved):

Issue #1520, opened 2021-04-12
Log user activity


chrismatix:
Prerequisite: #1518

**Motivation**
A basic requirement for security compliance for many companies is the ability to audit user activities on a system. 

GIVEN authentication is turned on for Airy Core
WHEN a user makes a request with a system token or via OIDC authentication
THEN the user information gets logged alongside the request

For this we could add the Kafka appender with `ops.application.logs` back into [RequestsLoggingFilter](https://github.com/airyhq/airy/blob/033981de68f57d3fde9c6c2c11d130553708080c/lib/java/spring/web/src/main/java/co/airy/spring/web/filters/RequestLoggingFilter.java) and attach either
- the last 6 symbols of the system token used 
- or the OIDC profile
Both should be provided by the security context.

PR #1588, opened 2021-04-20
[#1520] Log user activity

paulodiniz:
Fixes https://github.com/airyhq/airy/issues/1520


lib/java/spring/events/src/main/java/co/airy/spring/events/custom/CustomPublisher.java
```diff
@@ -0,0 +1,19 @@
+package co.airy.spring.events.custom;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.ApplicationEventPublisher;
+import org.springframework.scheduling.annotation.Async;
+import org.springframework.stereotype.Component;
+
+@Component
+public class CustomPublisher {
```
paulodiniz, line 9: This will be a kafka publisher

lib/java/spring/events/src/main/java/co/airy/spring/events/custom/CustomListener.java
```diff
@@ -0,0 +1,12 @@
+package co.airy.spring.events.custom;
+
+import org.springframework.context.ApplicationListener;
+import org.springframework.stereotype.Component;
+
+@Component
+public class CustomListener implements ApplicationListener<CustomEvent> {
```
paulodiniz, line 7: This will become a kafka producer

paulodiniz: Or we can not have any listeners and just produce one record in the producer

lib/java/spring/events/src/main/java/co/airy/spring/events/custom/KafkaListener.java
```diff
@@ -0,0 +1,32 @@
...
+    private final OpsApplicationLogs opsApplicationLogs = new OpsApplicationLogs();
+
+    public KafkaListener(KafkaProducer<String, String> producer) {
+        this.producer = producer;
+    }
+
+    @Override
+    public void onApplicationEvent(CustomEvent event) {
+        try {
+            producer.send(new ProducerRecord<>(opsApplicationLogs.name(), "wat", event.getMessage())).get();
```
paulodiniz, line 24: Will change the key here

lib/java/spring/events/src/main/java/co/airy/spring/events/filters/HTTPLogFilter.java
```diff
@@ -0,0 +1,111 @@
...
+
+            final Map toLog = Map.of("body", requestBody, "uri", requestUri, "headers", requestHeaders);
+            try {
+                return objectMapper.writeValueAsString(toLog);
+            } catch (JsonProcessingException e) {
+                e.printStackTrace();
+            }
+        }
+
+        return null;
```
paulodiniz, line 62: This is wrong. It's producing `null` payloads for the healthchecks. Will fix it

lib/java/spring/events/src/main/java/co/airy/spring/events/custom/HttpEventPublisher.java
```diff
@@ -0,0 +1,20 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.ApplicationEventPublisher;
+import org.springframework.scheduling.annotation.Async;
+import org.springframework.stereotype.Component;
+
+import java.util.Map;
+
+@Component
+public class HttpEventPublisher {
+    @Autowired
```
lucapette, line 12: apart from test code we're not autowiring anymore so I'd use a final var with a constructor her as well

lib/java/kafka/schema/src/main/java/co/airy/kafka/schema/ops/OpsApplicationLogs.java
```diff
@@ -0,0 +1,10 @@
+package co.airy.kafka.schema.ops;
+
+import co.airy.kafka.schema.OpsApplication;
+
+public class OpsApplicationLogs extends OpsApplication{
```
lucapette, line 5: checkstyle doesn't complain about the lack of space here? It really should 😂

lib/java/spring/events/src/main/java/co/airy/spring/events/filters/HttpLogFilter.java
```diff
@@ -0,0 +1,76 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+@Component
+public class HttpLogFilter extends OncePerRequestFilter {
+
+    @Autowired
```
lucapette, line 21: same comment here :)

lib/java/spring/events/src/main/java/co/airy/spring/events/filters/HttpLogFilter.java
```diff
@@ -0,0 +1,76 @@
...
+        boolean isFirstRequest = !this.isAsyncDispatch(request);
+        HttpServletRequest requestToUse = request;
+        if (isFirstRequest && !(request instanceof ContentCachingRequestWrapper)) {
+            requestToUse = new ContentCachingRequestWrapper(request, MAX_JSON_LENGTH);
+        }
+
+        try {
+            filterChain.doFilter(requestToUse, response);
+        } finally {
+            publishEvent(requestToUse);
```
paulodiniz, line 36: I've just realized this is the opposite of what we talked about. It will log the request only after processing it

lib/java/spring/events/src/main/java/co/airy/spring/events/filters/HttpLogFilter.java
```diff
@@ -0,0 +1,76 @@
...
+        String payload;
+        try {
+            payload = new String(buf, 0, bufferSize, request.getCharacterEncoding());
+        } catch (Exception e) {
+            return null;
+        }
+
+        return payload;
+    }
+}
```
lucapette, line 76: I would have expected the filter to be in spring web. `lib/spring/events` sounds too generic to have an http log filter to me.

So I'm thinking we can:

- move the filter to web (and let web depend on events and nothing else which is probably less error prone than what we do now as you can't "forget" the dep)
- merge this lib with web
- keep the code as is but specialise the name of the lib to `lib/spring/httplog`

I think my favourite is the first option but two also works. I'm not so fond of option 3 unless we do not add this as a direct dep but package it in the `base_web_app` deps set

lucapette (commented):
First pass. Mostly minor things... but there's s point that I'd surely want to discuss further before we merge it

chrismatix (approved):

PR #1591, opened 2021-04-20
[#1590] Fix api host variable injection in chatplugin

### LINKED ISSUE #1590: Refactor chat plugin into library
**Current state**
The chatplugin is being used as 
- an internal dependency to the inbox
- a library
- a default web app

This causes odd dependency issues, where runtime environments like `window.airy` or `process.env` are not available to the runtime.

**Desired goal**
The entry points here need to be separated so that there is a single interface library interface and the library itself is independent of runtime variables.

Also the HTTP lib needs to be refactored into a singleton module.

chrismatix:
fixes #1590

### DIFF
--- a/frontend/chat-plugin/src/api/index.tsx
+++ b/frontend/chat-plugin/src/api/index.tsx
@@ -1,17 +1,13 @@
 import {QuickReplyCommand, SuggestionResponse, TextContent} from 'render/providers/chatplugin/chatPluginModel';
 import {setResumeTokenInStorage} from '../storage';
 
-declare const window: {
-  airy: {
-    host: string;
-    channelId: string;
-  };
+let host;
+export const setApiHost = apiHost => {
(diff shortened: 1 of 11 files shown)


frontend/chat-plugin/src/api/index.tsx
```diff
@@ -1,17 +1,13 @@
 import {QuickReplyCommand, SuggestionResponse, TextContent} from 'render/providers/chatplugin/chatPluginModel';
 import {setResumeTokenInStorage} from '../storage';
 
-declare const window: {
-  airy: {
-    host: string;
-    channelId: string;
-  };
+let host;
+export const setApiHost = apiHost => {
```
AitorAlgorta, line 5: we should add a type here

AitorAlgorta (approved):
LGTM just a small comment

PR #1596, opened 2021-04-20
[#1557] fixed color picker for chat plugin customization

### LINKED ISSUE #1557: Fix Color picker for Chat Plugin Customization
**Current state** 
- the color picker for Accent Color and Background Color do not pop up at the right place: they pop up way below the input and the user needs to scroll down to see them

**To do**
- fix pop up of color picker

Screenshot 2021-04-15 at 13 29 34

AudreyKj:
closes #1557 

change: 
- the color picker for Accent and Background Color was not appearing at the right place 

before 
Screenshot 2021-04-15 at 13 29 34

after
Screenshot 2021-04-20 at 15 46 32


AitorAlgorta (approved):
🚀 🚀

Issue #1599, opened 2021-04-20
The /client.config endpoint must query the services health endpoint


paulodiniz:
We must query the health endpoint for each service returned on the controller api call to check the status

paulodiniz:
After a call with @lucapette we came up the following proposal

The return `/components` stays as it is now

```json
{
  "components": [
    "api-communication",
    "media-resolver",
    "integration-webhook",
    "api-admin",
    "api-websocket",
    "frontend-chat-plugin",
    "frontend-ui",
    "sources-chat-plugin",
    "sources-facebook",
    "sources-google",
    "sources-twilio"
  ]
}
```

The `/client.config` transforms the output, but it doesn't check the status

```json
{
  "components":  {
     "sources-facebook": {}
     "sources-google": {}
  }
}
```

Benefits: By removing the `enabled` field on the return, we would remove the overhead of having *and* checking a service healthcheck for each component

What do you think?

cc @lucapette @chrismatix @AitorAlgorta

chrismatix:
How would the output of the `status` command work then? Currently, it shows all components and whether or not they are running.

While I think this is a fast way to make this work, I would re-iterate the proposal I made on Slack. Instead of dumbing down the endpoint, we can make a better distinction between *apps* (user runnables in our cluster) and *components* (things you can configure in the `airy.yaml` i.e. a source). 

So to make `/client.config` useful for both use cases we could have `/components` return something like this:
```json5
{
  "sources-facebook-connector": {
    "enabled": true,
    "component": "sources-facebook"
  },
  "sources-facebook-events-router": {
    "enabled": true,
    "component": "sources-facebook"
  },
  "api-communication": {
    "enabled": true,
    "component": "api-communication"
  }
}
```
- The component identifier can be set as a Kubernetes annotation on the deployments/statefulsets
- The frontend can very easily `.reduce` on the component identifier to get the view that it needs

It's also worth considering renaming `enabled` to `running` because it more accurately reflects the check that the endpoint is doing.

Since we are already going for introspection it makes more sense to me to follow through with it.

AitorAlgorta:
it’s easier to find for us the right thing if the component name is they key, like it was in the beginning
```
"source-facebook": {
    "enabled": true,
    "healthy": false
}
```

chrismatix:
@AitorAlgorta that's true but long term it makes more sense if the endpoint reflects the state of the system more precisely. Say for instance you want to display exactly which apps are failing and why in the UI.

[...]

PR #1617, opened 2021-04-22
[#1603] Redirect when trying to add GBM channel

### LINKED ISSUE #1603: Redirect when trying to add GBM Channel
### Current Behaviour

When clicking to add a GBM channel, I get a redirect to my last conversation

### Expected Behaviour

Either dialog to connect the channel, or a message showing the channel is not there

gbm-bug

AudreyKj:
closes #1603

- GBM modal box linked to the Google source from the docs

Screenshot 2021-04-22 at 13 18 00

### DIFF
--- a/frontend/ui/src/pages/Channels/MainPage/index.tsx
+++ b/frontend/ui/src/pages/Channels/MainPage/index.tsx
@@ -3,6 +3,7 @@ import {withRouter, RouteComponentProps} from 'react-router-dom';
 
 import {Source, Channel, Config} from 'model';
 import {FacebookMessengerRequirementsDialog} from '../Providers/Facebook/Messenger/FacebookMessengerRequirementsDialog';
+import {GoogleBusinessMessagesRequirementsDialog} from '../Providers/Google/GoogleBusinessMessagesRequirementsDialog';
 import {TwilioRequirementsDialog} from '../Providers/Twilio/TwilioRequirementsDialog';
 import SourceDescriptionCard from '../SourceDescriptionCard';
 import ConnectedChannelsBySourceCard from '../ConnectedChannelsBySourceCard';
@@ -113,7 +114,7 @@ const SourcesInfo: SourceInfo[] = [
     image: <GoogleAvatarIcon />,
     newChannelRoute: '/',
     channelsListRoute: '/',
-    configKey: 'sources-google',
+    configKey: '',
     channelsToShow: 4,
     itemInfoString: 'channels',
     dataCyAddChannelButton: cyChannelsGoogleAddButton,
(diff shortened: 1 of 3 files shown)


frontend/ui/src/pages/Channels/MainPage/index.tsx
```diff
@@ -113,7 +114,7 @@ const SourcesInfo: SourceInfo[] = [
     image: <GoogleAvatarIcon />,
     newChannelRoute: '/',
     channelsListRoute: '/',
-    configKey: 'sources-google',
+    configKey: '',
```
chrismatix, line 117: Maybe I'm not getting it, but how can this field be empty? 🤔

AudreyKj: yes I know it's counter-intuitive, but I had to remove Google's configKey in order to have the modal pop-up. 

the backend returns `sources-google: {enabled: true}` but we are not yet able to connect Google via the UI, so we want to show the modal box describing the necessary configuration to the user

AitorAlgorta: then you should add the `configKey` and the backend shoud return `enabled: false`. @AudreyKj  @chrismatix  @lucapette

lucapette: yeah @AudreyKj I'm afraid I misunderstod the problem here. We have an issue to fix this behaviour. But it won't work like @AitorAlgorta is mentioning. `enabled: false` will be returned when google is correctly configured _but_ for some reason is not running at the moment.

As of last week, the `/client.config` endpoint has a different contract. We return in the response only components that are present in the airy instance (which means they've been configured already). So the right condition for the flow of the channels new accounts pages is `source component is not null and enabled` to proceed with it.


Having said that: I would merge this anyway. And fix this back in #1599 (which I think is already being worked on by @paulodiniz)

frontend/ui/src/pages/Channels/MainPage/index.tsx
```diff
@@ -160,7 +162,7 @@ const MainPage = (props: MainPageProps & RouteComponentProps) => {
               sourceInfo={infoItem}
               displayButton={!channelsBySource(infoItem.type).length}
               addChannelAction={() => {
-                if (config.components[infoItem.configKey] && config.components[infoItem.configKey].enabled) {
+                if (config.components[infoItem.configKey] && !config.components[infoItem.configKey].enabled) {
                   props.history.push(infoItem.newChannelRoute);
```
AudreyKj, line 166: updated the check to reflect the endpoint change

AitorAlgorta (approved):
🚀

PR #1619, opened 2021-04-22
[#1605] Fix Tags Flow

### LINKED ISSUE #1605: Tag color is always blue
### Current Behaviour

When creating a tag with another color (rather than the default blue), the assigned color is always blue 

### Expected behaviour

The created tag should have the selected color

new-tag

AitorAlgorta:
closes #1605


frontend/ui/src/pages/Tags/TableRow.tsx
```diff
@@ -1,16 +1,20 @@
...
 import {Button, LinkButton} from 'components';
-import {ReactComponent as EditIcon} from 'assets/images/icons/edit.svg';
-import {ReactComponent as TrashIcon} from 'assets/images/icons/trash.svg';
 import ColorSelector from '../../components/ColorSelector';
 import Tag from '../../components/Tag';
 import {Tag as TagModel, TagColor} from 'model';
 import {Settings} from '../../reducers/data/settings';
 import {StateModel} from '../../reducers';
+
+import {ReactComponent as EditIcon} from 'assets/images/icons/edit-pencil.svg';
```
lucapette, line 13: I suppose the "right" name would be `EditPencilIcon` (I know it was already there but since we're touching it...)

lib/typescript/httpclient/src/endpoints/listConversations.ts
```diff
@@ -12,7 +12,7 @@ export const listConversationsDef = {
   },
   mapResponse: response => {
     const conversationData = (response as PaginatedPayload<any>).data.map(messagePayload => ({
-      ...camelcaseKeys(messagePayload, {deep: true, stopPaths: ['metadata.user_data']}),
+      ...camelcaseKeys(messagePayload, {deep: true, stopPaths: ['metadata.user_data', 'metadata.tags']}),
```
lucapette, line 15: :angry-face:

lucapette: is this the only place where we have this problem?

lucapette (approved):
Super minor comments.... looks great!

lib/typescript/httpclient/src/endpoints/listConversations.ts
AitorAlgorta: I tested a bit, and it looks that everything works fine now. But I can't say 100%.

frontend/ui/src/pages/Tags/TableRow.tsx
AitorAlgorta: I will change it

lucapette (approved):

PR #1623, opened 2021-04-23
Add OIDC authentication backend

### LINKED ISSUE #1518: Add OIDC authentication middleware
Prerequisite: #1517 

## Motivation
We don't want to store or maintain users directly on Airy Core. Instead, we want to leverage [Open ID Connect (OIDC)](https://openid.net/connect/) to authenticate users against their custom providers.

### Backend

Users configure authentication mechanisms in the `airy.yaml`. See the [spring docs](https://docs.spring.io/spring-security/site/docs/5.2.x/reference/html/oauth2.html) for required OAuth2.0 provider values.

```yaml
...
    auth:
       systemToken: "token-to-be-used-for-non-web-based-api-interactions"
       oidc:
           allowedEmailPatterns: "*@airy.co,grace@example.org,l.ada@example.org"
           provider: "github"
           clientId: "some value"
           clientSecret: "some super secret value"
           ...
```

Config behavior: 

GIVEN that no auth config is provided 
WHEN a request is made to the backend
THEN there is no authentication check

GIVEN that either system token or oidc config is present
WHEN a request is made to the backend
THEN unauthenticated requests should be rejected

GIVEN that oidc config is present
WHEN a browser user opens the authentication route
THEN the oidc connection authentication flow should be started

GIVEN that a browser user is authenticated with OIDC
WHEN that user makes an authenticated request
THEN the `principal` should be available to endpoints (e.g. using the email for the [senderId](https://github.com/airyhq/airy/blob/51c6b2aad533fe8414b699b5c268ba00983dfee2/backend/api/communication/src/main/java/co/airy/core/api/communication/SendMessageController.java#L67))
THEN the principal email should be checked against `allowedEmails` for authorization

**TODOs**
- [x] docs!
- [x] Add an `allowedEmailPatterns` config that takes a list of emails or email wildcards
- [x] Add a sender id to the `/messages.send` endpoint
- [x] Implement oidc auth. For now, we will only implement and test the **Github provider** (decide on the auth endpoints and document for the frontend)
- [x] Add a `/users.getProfile` endpoint that returns the current OIDC profile.
- [x] Add a logout endpoint so that the frontend can delete the session 

**Resources:**
- 1h in-depth Oauth2.0 and open id connect https://www.youtube.com/watch?v=996OiexHze0
- Using OIDC with Spring security https://www.baeldung.com/spring-security-openid-connect 
- Spring OAuth 2.0 docs https://docs.spring.io/spring-security/site/docs/5.2.x/reference/html/oauth2.html

chrismatix:
resolves #1518 

TODO
- [x] access send message oidc principal
- [x] logout functionality
- [x] Use JWT for stateless spring sessions.
- [x] websocket auth
- [x] ingress routes


backend/api/communication/src/main/resources/application.properties
```diff
@@ -1,4 +1,10 @@
...
+kafka.brokers=localhost:8081
+kafka.schema-registry-url=http://localhost:8080
 kafka.cleanup=${KAFKA_CLEANUP:false}
-kafka.commit-interval-ms=${KAFKA_COMMIT_INTERVAL_MS}
+kafka.commit-interval-ms=1
+systemToken=123456
+oidc.allowedEmailPatterns=*@airy.co
+oidc.provider=github
+oidc.clientId=a4351f06bb9238f2bf18
+oidc.clientSecret=bf99a3c9c7503d3266c62d9ac56d24ed66044d5f
```
chrismatix, line 9: TODO reset this file

backend/api/admin/src/main/java/co/airy/core/api/admin/UsersController.java
```diff
@@ -0,0 +1,22 @@
...
+import org.springframework.http.ResponseEntity;
+import org.springframework.security.core.Authentication;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+public class UsersController {
+
+    @PostMapping("/users.getProfile")
+    ResponseEntity<?> getUserProfile(Authentication auth) {
```
chrismatix, line 14: We could move this to an optional field on the `/client.config` endpoint

lib/java/spring/auth/src/main/java/co/airy/spring/auth/oidc/EmailFilter.java
```diff
@@ -0,0 +1,58 @@
...
+    public EmailFilter(ConfigProvider config) {
+        this.configProvider = config;
+    }
+
+    @Override
+    protected void doFilterInternal(HttpServletRequest req,
+                                    HttpServletResponse res,
+                                    FilterChain chain) throws IOException, ServletException {
+        final Authentication auth = SecurityContextHolder.getContext().getAuthentication();
+        if (auth instanceof OAuth2AuthenticationToken) {
```
chrismatix, line 32: If the oauth flow is complete we still need to check if the authenticated user's email address matches the `allowedEmailPatterns` config

lib/java/spring/auth/src/main/java/co/airy/spring/auth/oidc/UserProperties.java
```diff
@@ -0,0 +1,61 @@
...
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+import java.util.Optional;
+
+@Data
+@Component
+@ConfigurationProperties(prefix = "oidc")
```
chrismatix, line 18: TYL https://docs.spring.io/spring-boot/docs/current/reference/html/spring-boot-features.html#boot-features-external-config-java-bean-binding

lib/java/spring/auth/src/main/java/co/airy/spring/auth/oidc/UserService.java
```diff
@@ -0,0 +1,52 @@
...
+
+    public UserService(GithubApi githubApi) {
+        this.githubApi = githubApi;
+    }
+
+    public OAuth2User loadUser(OAuth2UserRequest userRequest) throws OAuth2AuthenticationException {
+        OAuth2User user = userService.loadUser(userRequest);
+
+        // Github does not return a user's email by default
+        if ("github".equals(userRequest.getClientRegistration().getRegistrationId())) {
```
chrismatix, line 32: GitHub does not actually comply with OIDC. Dex uses the same approach of requesting the email scope and then fetching them separately to fake oidc authentication.

Therefore we create a custom user service that does this extra lookup

lib/java/spring/auth/src/main/java/co/airy/spring/auth/session/CookieSecurityContextRepository.java
```diff
@@ -0,0 +1,117 @@
...
+            super(response, true);
+            this.request = request;
+        }
+
+        @Override
+        protected void saveContext(SecurityContext securityContext) {
+            HttpServletResponse response = (HttpServletResponse) getResponse();
+            Authentication authentication = securityContext.getAuthentication();
+
+            if (authentication instanceof OAuth2AuthenticationToken) {
```
chrismatix, line 99: See the class comment.

If the current security context is set to a successful oauth2 authentication then we set an httpOnly, secure jwt cookie. This was necessary because it's unfeasible to store the entire oauth2 context in a client-side jwt.

Implications:
- All subsequent requests will be authenticated with this token
- Clients can clear this cookie by visiting `/logout`
- Prevents session hijacking through Javascript based attacks
- the original oauth2 context gets lost. So the `ProfileData.java` class has to include all user information needed for logging, workflow etc.

lib/java/spring/auth/src/main/java/co/airy/spring/auth/token/TokenAuth.java
```diff
@@ -0,0 +1,50 @@
...
+
+import lombok.Data;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.GrantedAuthority;
+
+import java.util.Collection;
+import java.util.List;
+
+@Data
+public class TokenAuth implements Authentication {
```
chrismatix, line 11: This authentication represents requests authorized via system token.

backend/api/websocket/src/main/java/co/airy/core/api/websocket/WebSocketConfig.java
```diff
@@ -58,39 +33,4 @@ TaskScheduler heartbeatScheduler() {
 
         return heartbeatScheduler;
     }
-
-    @Override
-    public void configureClientInboundChannel(ChannelRegistration registration) {
```
chrismatix, line 63: To quote [the docs]():

> WebSockets reuse the same authentication information that is found in the HTTP request when the WebSocket connection was made. This means that the Principal on the HttpServletRequest will be handed off to WebSockets. If you are using Spring Security, the Principal on the HttpServletRequest is overridden automatically.

> More concretely, to ensure a user has authenticated to your WebSocket application, all that is necessary is to ensure that you setup Spring Security to authenticate your HTTP based web application.

infrastructure/helm-chart/templates/ingress.yaml
```diff
@@ -92,13 +92,34 @@ spec:
                 name: api-communication
                 port:
                   number: 80
+          - path: /login
+            pathType: Prefix
+            backend:
+              service:
+                name: api-admin
```
chrismatix, line 99: The login flow needs to be completed against a single pod. On success, we exchange the OAuth authentication for an Airy session and it becomes statelessly available for all apps.

paulodiniz (approved):
LGTM

Issue #1615, opened 2021-04-22
Make tags more reliable for the UI


lucapette:
**Current state**

Right now the tags of a conversation may not show up in the conversation view of the UI because a conversation has been tagged with an id the redux store doesn't know about yet.

**Desired state**

Conversations have their tags in all circumstances

**How do we do it**

- [x] Introduce a new `tag` event for the websocket
- [x] Consume this event so the redux store gets "new tags"
- [x] every time we load a new page, we will fetch tags async (like we do with channels already)

**Open questions**

- [x] We reflect the new event in the webhook integration already?

PR #1624, opened 2021-04-23
[#1615] Expose tag events via websocket

lucapette:


AitorAlgorta (dismissed):
Lovely

AitorAlgorta (dismissed):

lib/typescript/websocketclient/payload.ts
```diff
@@ -1,46 +1,47 @@
...
     conversation_id: string;
     channel_id: string;
-    message: MessagePayload;
+    message: {
+      id: string;
+      content: string;
+      delivery_state: DeliveryState;
+      from_contact: boolean;
+      sent_at: Date;
+      metadata: any;
```
chrismatix, line 18: nitpick: Message events don't include metadata.

docs/docs/api/websocket.md
```diff
@@ -81,3 +82,17 @@ field informs the client of the kind of update that is encoded in the payload.
   }
 }
 ```
+
+### Tag
+
+```json5
+{
+  "type": "tag",
```
chrismatix, line 90: So for my understanding: For now we don't handle tag deletions?

frontend/ui/src/actions/metadata/index.ts
```diff
@@ -1,15 +1,9 @@
...
 
-const MERGE_METADATA = '@@metadata/MERGE_METADATA';
 const SET_METADATA = '@@metadata/SET_METADATA';
 
 export const setMetadataAction = createAction(
   SET_METADATA,
   (metadataEvent: MetadataEvent) => metadataEvent
 )<MetadataEvent>();
-
-export const mergeMetadataAction = createAction(
```
chrismatix, line 12: 🎉  So this is also a step on the way to default values on the client.

chrismatix (dismissed):
Very cool refactoring. Left some comments

lib/typescript/websocketclient/payload.ts
lucapette: yeah this was already there bu I agree. I'll change!

docs/docs/api/websocket.md
lucapette: No, the reason why I didn't do this is that right now "deleting tags" is not though off so well as it doesn't cascade to conversations (and maybe we should never do that. I guess we should discuss this in a separate issue though)

AitorAlgorta (approved):

Issue #1650, opened 2021-04-26
Message pagination


paulodiniz:
## Current state

Now we fetch all of the messages from Lucene according to the filter, and paginate them for the response

## Desired State

We want for the returned data from Lucene to be already paginated

AitorAlgorta:
related ticket #1791

PR #1651, opened 2021-04-26
[#1650] Message pagination

paulodiniz:
https://lucene.apache.org/core/5_0_0/core/org/apache/lucene/search/Collector.html


lucapette (approved):
It feels too good to be true you know :)

Issue #1558, opened 2021-04-15
AWS Provider Docs - Add Recommendation to enable HTTPS


steffh:
Since we want users to run and connect to their Airy instance securely in the cloud, we should recommend to enable HTTPS by adding SSL certificates (for example using [Let’s encrypt](https://letsencrypt.org/)).

The recommended way of doing this seems to be to enable SSL on either the ingress controller or the load balancer serving the ingress. 

We should refer to the documentation of AWS or the relevant cloud provider / ingress controller.

To the extent this is necessary for HTTPS to work, we should also show how the configuration needs to be adjusted in the relevant Airy instance.

steffh:
depends on: https://github.com/airyhq/airy/issues/1212

ljupcovangelski:
We had a discussion and we decided to solve https on a `per-provider` basis. Therefore I closed #1212 and for AWS we will write docs on how to create AWS ACM certificate and update the loadbalancer service to use that certificate with the loadbalancer them with the loadbalancer

ljupcovangelski:
We added instructions on howto enable HTTPS for AWS in the docs, however we are using a self-signed certificate and that is not an ideal solution.
We need to instruct on how to create a HTTPS tunnel with a generated `let's encrypt` certificate and also make `Airy core` available on a configurable domain name (example `myairy.customer.com`)

ljupcovangelski:
When trying to create Airy Core on Google Cloud, we realized that the HTTPS approach we have in the docs for AWS works completely different for Google.

So in my opinion we should again explore the option of adding HTTPS capabilities to our Traefik ingress, so that we can unify the process for different types of cloud providers.

ljupcovangelski:
This is done.

PR #1653, opened 2021-04-26
[#1558] Docs for configuring https on AWS

ljupcovangelski:


docs/docs/getting-started/installation/aws.md
```diff
@@ -128,15 +128,68 @@ Error creating vpc:  operation error EC2: CreateVpc, https response error Status
 When encountering this, you can delete some of the resources just as described
 on [here](/getting-started/installation/aws#uninstall-airy-core)
 
+## Enable HTTPS
+
+This section explains all the steps needed to configure HTTPS on your `Airy Core` instance.
```
lucapette, line 133:
```suggestion
This section guides you through the necessary steps to configure HTTPS on your `Airy Core` instance.
```

docs/docs/getting-started/installation/aws.md
```diff
@@ -128,15 +128,68 @@ Error creating vpc:  operation error EC2: CreateVpc, https response error Status
...
+
+Create a self-signing certificate (skip step if you already have your existing https certificate).
```
lucapette, line 135:
```suggestion
Create a self-signing certificate (skip step if you already have an existing HTTPS certificate).
```

docs/docs/getting-started/installation/aws.md
```diff
@@ -128,15 +128,68 @@ Error creating vpc:  operation error EC2: CreateVpc, https response error Status
...
+The certificates will be created in your `~/.airy/certs` directory, using the OpenSSL utility.
```
lucapette, line 136: is it necessary that the certs are in `~/.airy`? I can't say from reading the doc but if it is then we have to be more explicit, if not we should make it clear it's an example

docs/docs/getting-started/installation/aws.md
```diff
@@ -128,15 +128,68 @@ Error creating vpc:  operation error EC2: CreateVpc, https response error Status
...
+Country Name (2 letter code) [DE]:DE
+State or Province Name (full name) [Berlin]:Berlin
+Locality Name (eg, city) []:Berlin
+Organization Name (eg, company) [Internet Widgits Pty Ltd]:Airy GmbH
+Organizational Unit Name (eg, section) []:Development
+Common Name (e.g. server FQDN or YOUR name) []:*.elb.amazonaws.com
+Email Address []:sre@airy.co
+```
+
+Note that when generating the public certificate `public.crt` you must specify a common namd of fqdn, otherwise the certificate will not be used by the AWS LoadBalancer.
```
lucapette, line 153:
```suggestion
Note that when generating the public certificate `public.crt` you must specify a common name of FQDN, otherwise the certificate will not be used by the AWS LoadBalancer.
```

docs/docs/getting-started/installation/aws.md
```diff
@@ -128,15 +128,68 @@ Error creating vpc:  operation error EC2: CreateVpc, https response error Status
...
+
+Next you should upload your certificates to the AWS certificates manager (ACM).
+
+```sh
+aws acm import-certificate --certificate fileb://public.crt --private-key fileb://private.key --region us-east-1
+```
+
+In case you have your own certificates and have one more file with the certificate chain, you can also specify that file as anadditional `--certificate` value.
```
lucapette, line 161:
```suggestion
In case you have your own certificates and have one more file with the certificate chain, you can also specify that file as additional `--certificate` value.
```

It's a little hard to understand what's that "that file" in the second sentence so maybe we should reword a little

docs/docs/getting-started/installation/aws.md
```diff
@@ -128,15 +128,68 @@ Error creating vpc:  operation error EC2: CreateVpc, https response error Status
...
+
+Get the address of your LoadBalancer, assuming that you have your `kube.conf` file in the current working directory:
```
lucapette, line 163: instead of assuming where the file is we can make the example use `KUBE_CONF_DIR` before the command so it's clear

docs/docs/getting-started/installation/aws.md
```diff
@@ -216,7 +269,7 @@ do
 done
 ```
 
-Delete all used network interfaces
+Delete all used network interfaces (iIf the command fails, please check if all the `loadbalancers` are deleted and run the previous command one more time):
```
lucapette, line 272:
```suggestion
Delete all used network interfaces (iIf the command fails, check if all the `loadbalancers` are deleted and run the previous command one more time):
```

docs/docs/getting-started/installation/aws.md
```diff
@@ -128,15 +128,68 @@ Error creating vpc:  operation error EC2: CreateVpc, https response error Status
...
+kubectl --kubeconfig ./kube.conf -n kube-system get service traefik --output jsonpath='https://{.status.loadBalancer.ingress[0].hostname}{"\n"}'
+```
+
 ## Integrate public webhooks
 
 The public webhooks will be accessible on the public LoadBalancer which is
 created by the Ingress loadBalancer Kubernetes service.
 
-To get the public URL of your AWS Airy Core installation you can run:
+To get the public URL of your AWS Airy Core installation, asuming that your `kube.conf` file is in the current working directory, you can run:
```
lucapette, line 189: Using an ENV dir for the kube conf in the snippets, we also avoid adding this sentence before each example

docs/docs/getting-started/installation/aws.md
```diff
@@ -243,14 +296,19 @@ do
...
 
 ```sh
 for gateway in $(aws ec2 describe-internet-gateways --filters Name=attachment.vpc-id,Values=${vpc_id} --query 'InternetGateways[].InternetGatewayId' --output text)
 do
     aws ec2 detach-internet-gateway --internet-gateway-id ${gateway} --vpc-id ${vpc_id}
     aws ec2 delete-internet-gateway --internet-gateway-id ${gateway}
 done
+```
+
+Delete the route tables (expect that the command will fail for the default route table, but still you shouldn't have any problem deleting the VPC in the next step):
```
lucapette, line 309: `expect that the command will fail for the default route table` if it always fail we should reword like "the command will always fail for the default route table but it should still...."

lucapette (commented):
First pass!

As we're providing steps, it's probably helpful to make those part of the structure via paragraphs (with `###` I suppose)

lucapette (approved):

PR #1675, opened 2021-04-29
[#1652] Fix rendering messages with render library

### LINKED ISSUE #1652: Fix rendering messages with render library
**Current state** 
- There are a lot of messages that appear as "cannot render this content", especially for Google 

**To do**
- Investigate why so many messages are invalid 

Screenshot 2021-04-26 at 16 36 15

Screenshot 2021-04-29 at 10 19 07

AudreyKj:
closes #1652 

- the Inbound Suggestion Response messages from Google caused errors so couldn't be rendered because they are not nested in a `message` object

### DIFF
--- a/lib/typescript/render/providers/google/GoogleRender.tsx
+++ b/lib/typescript/render/providers/google/GoogleRender.tsx
@@ -51,7 +51,7 @@ function render(content: ContentUnion, props: RenderPropsUnion) {
 }
 
 function googleInbound(message: Message): ContentUnion {
-  const messageJson = message.content.message;
+  const messageJson = message.content.message ?? message.content;
 
   if (messageJson.richCard?.standaloneCard) {
     const {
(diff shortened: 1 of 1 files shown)


lucapette (approved):

thorstenairy (approved):

AitorAlgorta (approved):

lib/typescript/render/providers/google/GoogleRender.tsx
```diff
@@ -51,7 +51,7 @@ function render(content: ContentUnion, props: RenderPropsUnion) {
 }
 
 function googleInbound(message: Message): ContentUnion {
-  const messageJson = message.content.message;
+  const messageJson = message.content.message ?? message.content;
```
chrismatix, line 54: I'm not sure I understand how we got here? 🤔 If it's inbound the content should always have the same structure, right?

AudreyKj: Yes! Except when the inbound messages are a`SuggestionResponse` (the user clicks on a Suggestion and sends it as a message). In this case, messages have a different structure (`message.content.suggestionResponse.text` instead of `message.content.message.text`)

PR #1682, opened 2021-04-29
[#1665] Fix chatplugin reconnection problem

### LINKED ISSUE #1665: Chatplugin websocket connection crash loop
**Issue**
When there is a connection error with the WebSocket, the app immediately tries to re-authenticate causing an infinite crash loop that puts pressure on the CPU.

**Desired behavior**
When there is a WebSocket connection error, the should be a time out between connection attempts.

AitorAlgorta:
closes #1665 also fixed a small problem with the bubbleIcon

### DIFF
--- a/frontend/chat-plugin/src/App.tsx
+++ b/frontend/chat-plugin/src/App.tsx
@@ -65,5 +65,6 @@ export const config: Config = {
       },
     },
   },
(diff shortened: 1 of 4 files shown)


thorstenairy (approved):
LGTM

Issue #1486, opened 2021-04-08
Upgrade all dependencies


lucapette:
Dependabot is helping us quite a bit on the JS deps but can't do anything for our java things. The idea is to use this issue as a collaborative to do list and keep it open until we complete the list.

- [x] spring boot
- [x] General kafka upgrade
   - [x] kafka libs
   - [x] kafka images
   - [x] karapace
   - [x] confluent libs 
- [x] rules_docker
- [x] rules_pkg
- [x] rules_nodejs
- [x] react

chrismatix:
Closing this, since we have upgraded all the relevant packages

PR #1691, opened 2021-05-03
[#1486] Upgrade kafka images

ljupcovangelski:
required for #1486


lib/java/kafka/streams/src/main/java/co/airy/kafka/streams/KafkaStreamsWrapper.java
```diff
@@ -174,16 +174,19 @@ public KafkaStreamsWrapper withExcludeInternalTopics(boolean excludeInternalTopi
 
     public <T> ReadOnlyKeyValueStore<String, T> acquireLocalStore(String storeName) {
         int retries = 0;
-        do
+        do {
             try {
+                if (!streams.state().equals(KafkaStreams.State.RUNNING)) {
```
chrismatix, line 179: Prior to [KAFKA-10271](https://github.com/apache/kafka/commit/ccff7227ce11747c758c0903b34ee1d84119b9f9#) the underlying `QueryableStoreProvider` would expensively lookup stores from all registered `StreamThreadStateStoreProvider` instances. `StreamThreadStateStoreProvider::getStore` throws an exception if the thread is not `RUNNING` so there was an implicit check for that whenever we requested a store.

[KAFKA-10271](https://github.com/apache/kafka/commit/ccff7227ce11747c758c0903b34ee1d84119b9f9#) removed that extra lookup for performance reasons so that it became possible to access stores that would crash when trying to read from them.

chrismatix: This also explains why the integration tests were failing, while the runtime was working: The state stores do eventually get ready, but for the tests that were failing they were not ready at the expected time.

lucapette (approved):

PR #1700, opened 2021-05-03
[#1693]  Move defaults to airy yaml

### LINKED ISSUE #1693: Move default values from Helm to the airy yaml
## Current state
Values such as the [allowedOrigins](https://github.com/airyhq/airy/blob/8254a5f5ae6aa60ea59e74b88c8582cc1e4d4b9e/infrastructure/helm-chart/charts/apps/charts/api/templates/security.yaml#L10) are assigned a default in the Helm charts. 

Therefore they don't show up in the `airy.yaml` that is generated by `airy create` (or `airy creatе --init-only`). This means that on subsequent runs of `airy config apply` it can happen that those values get overwritten because `config apply` deletes values that are not present in the `airy.yaml`

## Proposal
Instead of defaulting these values in the Helm chart we turn [the embedded airy.yaml scaffolding file](https://github.com/airyhq/airy/blob/9dc02efa832b5902c09aa685c1972db5670a0a68/cli/pkg/workspace/template/airy.yaml) into a go template file and assign defaults there. 

**Benefits** 
- The state of the `airy.yaml` on creation is necessarily in sync with the cluster config maps
- Values can only be removed when you remove them from the airy.yaml itself
- The `GetHelmOverrides` method for the providers can be replaced with a typed dictionary that gets passed to the template

chrismatix:
resolves #1693


cli/pkg/cmd/create/create.go
```diff
@@ -43,7 +42,14 @@ func create(cmd *cobra.Command, args []string) {
 		cfgDir = args[0]
 	}
 
-	dir, err := workspace.Create(cfgDir)
+	w := console.GetMiddleware(func(input string) string {
+		return color.Colorize(color.Cyan, "#\t"+input)
+	})
+	provider := providers.MustGet(providers.ProviderName(providerName), w)
+	overrides := provider.GetOverrides()
```
chrismatix, line 49: Instead of passing the overrides as `--set` flags we merge them into the `airy.yaml` first before using it as a values file in Helm.

cli/pkg/workspace/template/copy.go
```diff
@@ -0,0 +1,81 @@
...
+type Variables struct {
+	NgrokEnabled            bool
+	Version                 string
+	Namespace               string
+	Host                    string
+	LoadBalancerAnnotations string
+}
+
+//go:embed src
+var templateDir embed.FS
```
chrismatix, line 21: Refactored the templating part of the workspace module into its own module

cli/pkg/workspace/template/copy.go
```diff
@@ -0,0 +1,81 @@
...
+func recCopy(writePath string, templatePath string, entry fs.DirEntry, data Variables) error {
+	dstPath := filepath.Join(writePath, entry.Name())
+	templatePath = filepath.Join(templatePath, entry.Name())
+
+	if !entry.IsDir() {
+		content, err := templateDir.ReadFile(templatePath)
+		if err != nil {
+			return err
+		}
+		if _, err := os.Stat(dstPath); err == nil {
```
chrismatix, line 47: This was a bug: The `create` command was previously overwriting config that was already present.

go.mod
```diff
@@ -6,6 +6,9 @@ go 1.16
 // Automatically generated by running //tools/update-deps
 
 require (
+	github.com/Masterminds/goutils v1.1.1 // indirect
+	github.com/Masterminds/semver v1.5.0 // indirect
+	github.com/Masterminds/sprig v2.22.0+incompatible // indirect
```
chrismatix, line 11: This is the same templating library that helm uses

infrastructure/helm-chart/templates/configmap.yaml
```diff
@@ -2,6 +2,18 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: hostnames
-  namespace: {{ .Values.global.namespace }}
+  namespace: {{ .Values.global.kubernetes.namespace }}
 data:
-  HOST: {{ .Values.global.host }}
+  HOST: {{ .Values.global.kubernetes.host }}
+---
```
chrismatix, line 8: Hoisted the `security` configmap to the top level

ljupcovangelski (commented):
I think that the order in which we generate the `airy.yaml` file should be:
- we grab the helm defaults
- we grab the overrides for a specific provider
- we provision `Airy core` in the Kubernetes cluster add grab the generated values (such as `host`)
- we generate the `airy.yaml` file at the end

chrismatix:
In an offline conversation @ljupcovangelski and I decided that the best move forward would be to:
1. default the host value in the helm `values.yaml` with `http://airy.core`
2. leave the host variable empty in the `airy.yaml` for aws
3. use the AWS `PostInstallation` hook to update the `airy.yaml` with the correct host value

To improve `1.` in such a way that we don't even need to default to an imaginary host value we introduce this follow up ticket for improvement #1761

ljupcovangelski (approved):
I tested and it works great :tada:

PR #1704, opened 2021-05-03
[#1694] Adding icons for lastMessage

### LINKED ISSUE #1694: No last message text in conversationList when a "non-text" message was sent last
CURRENT:
Screenshot 2021-04-30 at 10 03 29

SUGGESTION:
Screenshot 2021-05-03 at 11 46 00

I would recommend the solution we have in platform. Using a image that is shown, when a template is the last message.

thorstenairy:
closes #1694


infrastructure/cli.yaml
```diff
@@ -0,0 +1,5 @@
+apihost: http://airy.core
```
AitorAlgorta, line 1: Delete file 💣

AitorAlgorta (changes requested):

AitorAlgorta (approved):

PR #1737, opened 2021-05-05
[#1519] Implement auth UI behavior

### LINKED ISSUE #1519: Implement OIDC authentication UI
Prerequisite: #1518 

With OIDC the authentication state will be managed entirely by Spring as an HTTP session. This means that the frontend no longer has to store/handle a JWT for authentication. However, the inbox UI still needs to be aware of who is currently authenticated or not.

To address this problem #1518 adds a `/users.getPrincipal` endpoint that will return the profile information of the currently logged-in users.

So the flow would be as follows:
1. User loads the react app
2. App requests `/client.config` before rendering anything
3. **If principal present**: Store the user information in redux (not local storage) and proceed
4. **If 403 unauthorized**: Redirect to oidc auth endpoint (see docs provided in #1518)

If the oidc flow is successful the backend will redirect the user back to the app after 4. and the process will repeat.

chrismatix:
resolves #1519

### DIFF
--- a/lib/typescript/httpclient/client.ts
+++ b/lib/typescript/httpclient/client.ts
@@ -59,15 +59,47 @@ interface EndpointDefinition<T, K = void> {
 }
 
 export class HttpClient {
-  public readonly apiUrlConfig?: string;
-  private readonly unauthorizedErrorCallback?: (body: any) => void;
+  public readonly apiUrl?: string;
+  public readonly loginUrl?: string;
+  private readonly unauthorizedErrorCallback?: (body: any, loginUrl: string) => void;
 
-  constructor(apiUrlConfig: string, unauthorizedErrorCallback?: (body: any) => void) {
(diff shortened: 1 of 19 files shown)


lib/typescript/httpclient/client.ts
```diff
@@ -59,15 +59,47 @@ interface EndpointDefinition<T, K = void> {
...
+      mode: 'cors',
+      credentials: 'include',
+      body: body as BodyInit,
+    });
+
+    return this.parseBody(response);
+  }
+
   private async parseBody(response: Response): Promise<any> {
+    if (this.isAuthRedirect(response)) {
```
chrismatix, line 97: This is not an optimal solution. I could not get Spring to return a `403` with the OAuth middleware. So instead we try and detect when we are being redirected and treat that as an authorization error.

lucapette (approved):

PR #1747, opened 2021-05-06
[#1749] Fixed activeFilterCount

thorstenairy:


lucapette (dismissed):
So if I understand the code correctly:

- we do not need the count (it's either "active" or not the filter)
- if that's true, then the variable keeping the count is not needed as you can do one long or expression like `filterA.length > 0 || filterB.length >0` and so on which would declutter the function quite a bit
- if we do need the count, then I'd declare the variable right before using it inside the function and also rename it to `activeFilterCount(): number` so it's much clearer what it does (the little details you know :))
- Whatever we need the count or not, I'm not sure I understand why sometimes we write `currentFilter?.` and others we do `currentFilter.`. As I see it, either it's always needed or never needed (maybe I'm not getting it though). If it's always needed I'd early return `currrentFilter === undefined then false` (or 0 if we need the count)

thorstenairy:
Unfortunately we need to count as we have 3 states for isStateOpen (true, false and undefined) so currentFilter.length is 1 although we didn't set any filter when we select "ALL". DisplayName has kinda the same behaviour, if you never search it never appears in currentFilter, but after you searched and you removed it it appears as null or undefined in currentFilters.

AitorAlgorta (dismissed):

AitorAlgorta (approved):

PR #1775, opened 2021-05-11
[#1768] bad calls to conversations info when opening inbox

### LINKED ISSUE #1768: Bad calls to conversations.info when opening inbox
**Problem**

When opening the inbox, the current conversation is being requested with `/conversations.info` but the request payload is empty which results in a loop of `400 Bad Request` calls.

This does not impact the app behavior.

image

**Expected behavior**
The endpoint is called with a conversation id when needed.

AudreyKj:
closes #1768

### DIFF
--- a/frontend/ui/src/actions/conversations/index.ts
+++ b/frontend/ui/src/actions/conversations/index.ts
@@ -5,6 +5,7 @@ import {PaginatedResponse} from 'httpclient';
 import {HttpClientInstance} from '../../InitializeAiryApi';
 import {StateModel} from '../../reducers';
 import {setMetadataAction} from '../metadata';
+import {ConversationInfoError} from '../../types';
 
 const CONVERSATION_LOADING = '@@conversation/LOADING';
 const CONVERSATIONS_LOADING = '@@conversations/LOADING';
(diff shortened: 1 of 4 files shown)


frontend/ui/src/actions/conversations/index.ts
```diff
@@ -87,8 +88,8 @@ export const getConversationInfo = (conversationId: string, retries?: number) =>
       dispatch(mergeConversationsAction([response]));
       return Promise.resolve(true);
     })
-    .catch(async (error: Error) => {
-      if (retries > 5) {
+    .catch(async (error: ConversationInfoError) => {
+      if (retries > 5 || error.status === 404) {
         return Promise.reject(error);
       } else {
```
AudreyKj, line 94: if the conversation Id is not found (invalid match.params), we return and do not fetch again to avoid loops

frontend/ui/src/pages/Inbox/Messenger/MessengerContainer/index.tsx
```diff
@@ -35,10 +35,10 @@ const MessengerContainer = ({
...
 
   useEffect(() => {
-    if (!currentConversation) {
+    if (!currentConversation && match.params.conversationId) {
       getConversationInfo(match.params.conversationId);
     }
-  }, [conversations]);
+  }, [currentConversation, match.params.conversationId]);
 
   const hideSuggestedReplies = () => {
```
AudreyKj, line 43: there was a 404 loop when opening inbox because the endpoint was called  even if there was no conversation ID

frontend/ui/src/actions/conversations/index.ts
AitorAlgorta: The loop is necessary to find the conversation when we get a websocket notification. If we remove this loop we won't be able to display conversations that come from ws because sometimes some metadata updates arrive even before the actual conversation.

lucapette (approved):

PR #1799, opened 2021-05-17
Add health status to components endpoint

### LINKED ISSUE #1599: The /client.config endpoint must query the services health endpoint

chrismatix:
fixes #1599


lib/java/spring/kafka/healthcheck/src/main/java/co/airy/spring/kafka/healthcheck/ProducerHealthCheck.java
```diff
@@ -1,37 +0,0 @@
...
-import org.apache.kafka.clients.producer.ProducerRecord;
-import org.springframework.stereotype.Service;
-
-import java.time.Instant;
-
-@Service
-public class ProducerHealthCheck {
-    private final KafkaProducer<String, HealthCheck> producer;
-
-    ProducerHealthCheck(KafkaProducer<String, HealthCheck> producer) {
```
chrismatix, line 15: This wasn't actually used so I removed it

infrastructure/helm-chart/charts/apps/charts/api/charts/api-admin/templates/deployment.yaml
```diff
@@ -51,8 +51,6 @@ spec:
             configMapKeyRef:
               name: kafka-config
               key: KAFKA_COMMIT_INTERVAL_MS
-        - name: SERVICE_NAME
```
pascal-airy, line 54: Why did you remove the service names?

chrismatix: @pascal-airy Because I don't think that they are used. There used to be an implicit usage in the `ProducerHealthCheck`, but that was also unused so I removed both.

pascal-airy (approved):
lgtm!

PR #1821, opened 2021-05-21
[#1808] Improved lastMessageSent

### LINKED ISSUE #1808: Wrong lastMessageIcon for Twilio SMS messages
If you receive a text message from twilio SMS the lastMessageIcon displays the template icon although we expect a string.
On top of that we don't display the avatar and name (in this case phone number) in the conversationHeader.

Screenshot 2021-05-18 at 14 30 59

thorstenairy:
closes #1808

### DIFF
--- a/frontend/ui/src/pages/Inbox/ConversationListItem/index.tsx
+++ b/frontend/ui/src/pages/Inbox/ConversationListItem/index.tsx
@@ -88,8 +88,38 @@ const ConversationListItem = (props: ConversationListItemProps) => {
     }
   }, [active, conversation, currentConversationState]);
 
-  const lastMessageIcon = (conversation: Conversation) => {
+  const lastMessageIsText = (conversation: Conversation) => {
     const lastMessageContent = conversation.lastMessage.content;
+
+    if (typeof lastMessageContent === 'string') {
+      if (lastMessageContent.includes('&Body=' && '&FromCountry=')) {
+        const startText = lastMessageContent.search('&Body=');
+        const endText = lastMessageContent.search('&FromCountry=');
+        const textLength = endText - startText;
+        const enCodedText = lastMessageContent.substring(startText + 6, startText + textLength);
+        const replaced = enCodedText.split('+').join(' ');
+        const text = decodeURIComponent(replaced);
+        return text;
+      } else if (lastMessageContent.includes('&Body=' && '&To=whatsapp')) {
+        const startText = lastMessageContent.search('&Body=');
+        const endText = lastMessageContent.search('&To=whatsapp');
+        const textLength = endText - startText;
+        const enCodedText = lastMessageContent.substring(startText + 6, startText + textLength);
+        const replaced = enCodedText.split('+').join(' ');
+        const text = decodeURIComponent(replaced);
+        return text;
+      }
+    }
+    if (lastMessageContent.text || lastMessageContent.message?.text) {
+      return <FormattedMessage message={conversation.lastMessage} />;
+    } else if (lastMessageContent.suggestionResponse) {
+      return <>{conversation.lastMessage.content.suggestionResponse.text}</>;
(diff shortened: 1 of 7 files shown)


frontend/ui/src/pages/Inbox/ConversationListItem/index.tsx
```diff
@@ -88,8 +88,38 @@ const ConversationListItem = (props: ConversationListItemProps) => {
     }
   }, [active, conversation, currentConversationState]);
 
-  const lastMessageIcon = (conversation: Conversation) => {
+  const lastMessageIsText = (conversation: Conversation) => {
     const lastMessageContent = conversation.lastMessage.content;
+
+    if (typeof lastMessageContent === 'string') {
```
chrismatix, line 94: This code should be in the rendering library, right?

AudreyKj (approved):
looks good to me!

frontend/ui/src/pages/Inbox/ConversationListItem/index.tsx
```diff
@@ -93,21 +93,23 @@ const ConversationListItem = (props: ConversationListItemProps) => {
...
-        const enCodedText = lastMessageContent.substring(startText + 6, startText + textLength);
-        const replaced = enCodedText.split('+').join(' ');
-        const text = decodeURIComponent(replaced);
-        return text;
+        return (
+          <SourceMessage
+            source={conversation.channel.source}
+            contentType="message"
+            content={conversation.lastMessage}
+            isTwilioConversationListItem={true}
```
chrismatix, line 110: The API of the render library needs to be source agnostic. What does this attribute change?

thorstenairy: the styling of the message itself, so it not a bubble as in messageList @chrismatix

frontend/ui/src/pages/Inbox/ConversationListItem/index.tsx
thorstenairy: it is in the rendering lib, we just didn't use it. We should refactor the conversationListItem so it uses the render lib for every source, don't we @AitorAlgorta @chrismatix

AitorAlgorta: Do we really want to use the render lib also fro the preview of the message? I am not sure what is best

frontend/ui/src/pages/Inbox/ConversationListItem/index.tsx
AitorAlgorta: I agree with @chrismatix the render lib should be agnostic. That's what I meant in my previous comment. Either we don't use the render lib for this and we do it outside checking all the types, or have a different component in the render lib for the preview of the message. But we shouldn't change the `SourceMessage` for this. @airyhq/codeowners-frontend

frontend/ui/src/pages/Inbox/ConversationListItem/index.tsx
lizhutter: I think we should use the render lib to show the preview of the message, that way it shows more of a description than just an icon

chrismatix: @AitorAlgorta Picking up on your other comment: Doing it in the render library is simpler, because each renderer is split into a `mapper` and a `renderer`. The `SourceMessagePreview` can re-use the `mapper` part. This is reason why I did the split in the first place :) 

Furthermore, we can simplify the preview into "text" vs "something else" (images, data, audio, etc.) to make the components simpler. So we add two components and then a `SourceMessagePreview`, where every source uses its `mapper` and then maps text to the text component and anything else to the "something else" (data) component.

AitorAlgorta (approved):

Issue #1616, opened 2021-04-22
Airy on AWS: Customize create parameters and provide cost estimations


ljupcovangelski:
We need to allow for customizing some parameters when doing `airy create` on AWS:
- Instance type: The type of the instances in the EKS cluster (currently it is `c5.xlarge` ~ $122 per month)
- Scaling config: Number of instances in the nodeGroup (currently it is the AWS default - 2)

Also we should update the docs with information for calculating how much will it roughly cost to run `Airy Core` on AWS.

ljupcovangelski:
The instance type is done, for the number of nodes - we stay with two nodes by default for now.

PR #1833, opened 2021-05-25
[#1616] Customize instance type for AWS

ljupcovangelski:
Works like this:
```
airy create --provider aws --provider-config instanceType=c5.large
```
Default instance is `c5.xlarge`.


pascal-airy (approved):
good stuff!

PR #1864, opened 2021-05-28
[#1773] Filter conversations by source

### LINKED ISSUE #1773: Inbox: Filter bySource option
Currently, system users are able to filter the Inbox by tags, open/closed state, and read/unread.

We should offer system users the ability to filter bySource. This way, users can filter by Facebook, Google, Chat Plugin, SMS, and WhatsApp. This filtering will be done via filter modal.

chrismatix:
resolves #1773


lib/typescript/model/ConversationFilter.ts
```diff
@@ -1,10 +0,0 @@
-export interface ConversationFilter {
```
chrismatix, line 1: This does not belong in the model library since it's a UI model

pascal-airy (approved):
lgtm!

thorstenairy (approved):
LGTM! but we should think about the positioning of the pop up module of the filtering as it has now 1 more option. Might look a bit odd on a 13' display

PR #1879, opened 2021-05-31
[#1716] Edit contact information

### LINKED ISSUE #1716: Editing Contact Information
For the “contact information” that we have in core currently: we want to allow system users to edit the information or rename the contact.

We still do not want to save contacts in core, but we should enable this at least for system users' viewing within the conversation. Otherwise, every conversation via chat plugin will appear as though it is with `chatplugin##` rather than a contact's real name.

From my understanding: if this information is just metadata that system users could edit manually, it wouldn’t require that we come up with the context of contacts at all.

Also, contact information already appears when conversations are started via other channels (like Facebook or WhatsApp), so if it makes it easier: chat plugin conversations could be the only ones affected by this.

Ideally, it would be great if we could make it editable via UI, but I'm open to alternatives.

chrismatix:
required for #1716


pascal-airy (approved):
lgtm!

PR #1935, opened 2021-06-07
[#1933] Suggested Replies does not fit in messageInput

### LINKED ISSUE #1933: Suggested Replies does not fit in messageInput
Screenshot 2021-06-07 at 10 37 56

AudreyKj:
closes #1933

### DIFF
--- a/frontend/ui/src/pages/Inbox/MessageInput/index.tsx
+++ b/frontend/ui/src/pages/Inbox/MessageInput/index.tsx
@@ -72,7 +72,9 @@ const MessageInput = (props: MessageInputProps & ConnectedProps<typeof connector
   const sendButtonRef = useRef(null);
   const emojiDiv = useRef<HTMLDivElement>(null);
   const templateSelectorDiv = useRef<HTMLDivElement>(null);
+  const selectedSuggestedReplyDiv = useRef<HTMLDivElement>(null);
   const removeTemplateButton = useRef(null);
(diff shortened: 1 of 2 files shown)


frontend/ui/src/pages/Inbox/MessageInput/index.tsx
```diff
@@ -292,6 +338,12 @@ const MessageInput = (props: MessageInputProps & ConnectedProps<typeof connector
   };
 
   const selectSuggestedReply = (reply: SuggestedReply) => {
+    if (selectedSuggestedReply) setSelectedSuggestedReply(null);
+
+    if (input) setInput('');
+
+    if (selectedTemplate) setSelectedTemplate(null);
+
```
AudreyKj, line 346: here improved Ux: the newest selected templates or suggested reply is the one that gets selected in the input, overwriting any previously selected templates or suggested reply

lib/typescript/render/SourceMessagePreview.tsx
```diff
@@ -18,7 +18,7 @@ const FormattedMessage = ({message}: FormattedMessageProps) => {
   return <div />;
 };
 
-const isImageFromGoogleSource = (messageText: string) => {
+const isImageFromGoogleSource = (messageText: string | undefined) => {
   if (!messageText) return false;
```
AudreyKj, line 22: fix the type here, as messageText can be either a string or undefined

PR #2068, opened 2021-06-29
[#2060] edit text templates 

### LINKED ISSUE #2060: Edit Text Templates
Currently, when a system user views their current available text templates and they select one to send, they are unable to edit the text before sending it.

screenshot_2021-06-29_at_11 23 20

The image below shows the current view when a user selects a template:
screenshot_2021-06-29_at_11 23 32

Ideally: when a text template is selected, the text should be copied to the input field instead of rendering the whole template within the input bar.

AudreyKj:
closes #2060


thorstenairy (approved):
LGTM

PR #2092, opened 2021-07-01
[#661] Added possibility to connect Google via UI

### LINKED ISSUE #661: Enable users to connect via Google Business Messages
### Summary
To enhance overall experience, we should offer the ability to connect Google Business Messages for Google business pages.

### Requirements
On Airy Core, it should be an option to connect GBM for Google business pages. We need to clearly state that Google only allows a few partners to enable messaging. Along with this notice, we need to deliver a message including the Airy GBM documentation.

When connecting Google Business Messages, we need to request the following information:
- Agent ID
- Avatar
- Display Name

### Design Note:
Please note that although the following items are included in the above image, we will not include these:
- Welcome Message
- Add a name (optional)

For the Avatar, we need to clarify that this will be used for internal purposes only as a means of differentiating between channels. We also need to state that without a set image, it will be assigned a standard icon within Core.
3 3 0 Connect Google @2x

Once the Agent ID has been connected, the Google business page should be visible on the Admin page, along with the avatar and Agent ID. It should be possible to see whether the page is properly connected. If it is not, we should display an error message next to the affected page.
3 4 0 Connect GBM@2x

thorstenairy:
closes #661

### DIFF
--- a/frontend/ui/src/actions/channel/index.ts
+++ b/frontend/ui/src/actions/channel/index.ts
@@ -11,6 +11,7 @@ import {
   ConnectTwilioSmsRequestPayload,
   ConnectTwilioWhatsappRequestPayload,
   UpdateChannelRequestPayload,
+  ConnectChannelGoogleRequestPayload,
 } from 'httpclient';
 
 import {HttpClientInstance} from '../../InitializeAiryApi';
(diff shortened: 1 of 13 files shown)


AitorAlgorta (approved):
🚀

AudreyKj (approved):
👏

PR #2120, opened 2021-07-06
[#1590] Refactor: Split chatplugin into library and app targets

### LINKED ISSUE #1590: Refactor chat plugin into library

chrismatix:
resolves #1590 

- Got rid of `AiryWidget`
- [Split code](https://github.com/airyhq/airy/blob/ca076057ae9173152388ed96ef5b7cdc60c9d907/frontend/chat-plugin/README.md)
- `AiryChatPlugin` as the single interface for libraries and apps 
- Some code refactoring


frontend/chat-plugin/lib/src/components/chat/index.tsx
```diff
@@ -48,18 +48,6 @@ const Chat = (props: Props) => {
     defaultWelcomeMessage.content = config.welcomeMessage;
   }
 
-  const customStyle = {
```
chrismatix, line 51: This is now handled in the `AiryChatPlugin` component, which acts as a wrapper

AudreyKj (approved):

PR #2123, opened 2021-07-07
Change endpoint for webhook to /twilio

armanjindal:
In twilio-sources.mdx the Twilio Webhook URL endpoint was`/facebook` instead of `/twilio` which resulted in a 404. Just changed the endpoint to be the right one.


AitorAlgorta (approved):
Thanks! <3

PR #2129, opened 2021-07-07
[#2115] Conversation Status Changes occasionally not reflected in Inbox UI

### LINKED ISSUE #2115: Conversation Status Changes occasionally not reflected in Inbox UI
**Describe the bug**
In the Inbox UI, sometimes status changes for conversations seem to be not properly transmitted to the API, as users reported they clicked e.g. on CLOSED, the status was updated in the frontend, but when they came back to the conversation it was marked OPEN again (without any new events happening in the meantime).

**To Reproduce**
1. In the Inbox UI of any Airy instance with a lot of conversations
2. when you go manually through them and change the status of multiple conversations (e.g. from OPEN to CLOSED) 
3. you will see inconsistencies happening from time to time, i.e. button not clickable or button clickable, but status change is not reflected properly in the conversation list / conversation view or when reloading the view for the relevant conversation

**Expected behavior**
Each interaction with the buttons to change the state of the conversation should be reflected (i) in a change of the conversation list, (ii) the relevant buttons on top of the conversation view and (iii) successfully transmitted to the backend via the API / websocket.

**Screenshots**

The agent reported to have clicked on CLOSED in this case after the question had been answered:

image001

Another example for a status inconsistency between conversation list on the left and conversation view on the right:

Bildschirmfoto 2021-07-05 um 14 10 38

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser: Chrome
 - Version: 91

**Additional context**
The issue seems to happen occasionally but is reproducible when working the Inbox UI over a longer period of time.

AudreyKj:
closes #2115 

There is a delay between the time the state conversation is changed (through the action conversation state) and the websocket update (through the action setMetadata): this delay causes a buggy behaviour when the user toggles the state.

i'm suggesting to keep in Redux only the state changed through the action conversation state.


frontend/ui/src/reducers/data/conversations/index.ts
```diff
@@ -216,7 +216,10 @@ function allReducer(
           [action.payload.identifier]: {
             id: action.payload.identifier,
             ...state.items[action.payload.identifier],
-            metadata: merge({}, state.items[action.payload.identifier]?.metadata, action.payload.metadata),
+            metadata: {
+              ...merge({}, state.items[action.payload.identifier]?.metadata, action.payload.metadata),
+              state: state.items[action.payload.identifier].metadata.state,
```
chrismatix, line 221: I'm not sure I understand this: When a metadata update comes in we ignore it and keep the present value even when it's `undefined`?

AudreyKj: the conversation state is currently being changed twice in redux each time (first via the conversation state action and then via the websocket's metadata update). There is a delay between both updates, which creates a buggy behaviour. Maybe i'm missing something but I don't think this value needs to be updated twice by two different actions.

my idea is to ignore the websocket's metadata update for conversation state because it comes after the first update

AitorAlgorta: but then we might be ignoring some other relevant metadata updates right? like the tags from example. Do tags work with this changes?

AudreyKj: I'm only ignoring the metadata update for state (the changes are made for the others metadata updates like tags or display name ect.)

`metadata: {
              ...merge({}, state.items[action.payload.identifier]?.metadata, action.payload.metadata),
              state: state.items[action.payload.identifier].metadata.state,`

chrismatix: @AudreyKj yes the bug is likely due to the delay. As far as I understand the first update is optimistic, but just like with the contact display name we rely on the final state to arrive via the WebSocket. 

Another thing to consider with this solution is that other users of the same inbox won't be able to see that a conversation's state has changed. 

>**To Reproduce**
Steps to reproduce the behavior:
> 1. In the Inbox UI of any Airy instance with a lot of conversations
> 2. when you go manually through them and change the status of multiple conversations (e.g. from OPEN to CLOSED) 
> 3. you will see inconsistencies happening from time to time, i.e. button not clickable or button clickable, but status change is not reflected properly in the conversation list/conversation view or when reloading the view for the relevant conversation

Were you able to reproduce the behavior? It could also be that the API calls to update the state are failing for this user somehow.

AudreyKj: @chrismatix yeah I was able to reproduce locally and see that inconsistencies are happening from time to time (it happens occasionally and when the user quickly toggles the state because of the delay). The API calls are working locally but it could also be that the API calls are failing for a particular user.

yeah this solution is not ideal because it will prevent state conversation change in real time - I'm researching others alternatives

AitorAlgorta (approved):

Issue #1212, opened 2021-03-11
Https setup for traefik


ljupcovangelski:
For Airy core to be safely usable in cloud environments, we need to configure the ingress controller to use https and some default certificates.
As this connection will be insecure, we should also add documentation for how to modify the default certificates.

ljupcovangelski:
We discussed few options on how to enable `https` for Airy Core, among which:
1) Put https in core, in traefik with some default certs and instructions on how to change them
2) Put https in core, in traefik with letsencrypt
3) Use per-provider solution. For example, for AWS - write docs on how to create AWS ACM certificate and update the service to use that certificate with the loadbalancer them with the loadbalancer
4) Instruct people how to deploy ngrok and have the access to the ngrok public webhook

We decided not to put https in core and to put https on a `per-provider` basis (option 3).

Therefore this ticket is not needed anymore.

steffh:
As talked with @ljupcovangelski, we should have another look at this topic as HTTPS is crucial due to the requirements of conversational sources we integrate with.

I would suggest to reassess the position to have another ingress pod inside the Kubernetes cluster e.g. that could run certbot (https://certbot.eff.org/) to renew SSL certificates automatically via Let's Encrypt.

ljupcovangelski:
After some research and discussions, best approach would be to put the ingress controller in a separate helm chart, so that:
- HTTPS is not part of the core chart.
- When deploying `Airy Core` one can choose if he wants to use https and what is the hostname.
- We can put a per-provider defaults (don't use for Minikube, use specific annotations for AWS or Google).

For choosing the certificates, there are two options:
- Use letsencrypt (will require the hostname to be defined upfront).
- Use existing certificates that the user has and put them in Kubernetes secrets.

AitorAlgorta:
@steffh @lizhutter

steffh:
@ljupcovangelski that sounds like a good idea! please proceed with this plan! thank you!

ljupcovangelski:
I created an initial deployment with letsencrypt. After creation, the ingress controller reads the `host` entries in the `ingress` resources and creates the letsencrypt certificates. Renewals and redirect from HTTP->HTTPS is configured. Letsencrypt doesn't work with the loadbalancer url (because the url is very long) so a custom domain must be used.

The following variables determine the behavior https://github.com/airyhq/airy/pull/2133/files#diff-ffe3c279bc151d521a2e415d826a6ed84d085f428990445dca6b08ff9478e7ad
[...]

PR #2133, opened 2021-07-09
[#1212] HTTPS inside Kubernetes

ljupcovangelski:
- Added option for helm to deploy with letsencrypt.
- After creation, the ingress controller reads the `host` entries in the `ingress` resources and creates the letsencrypt certificates.
- Letsencrypt doesn't work with the loadbalancer url, because the url is very long.

Potential usage:
- Run `airy create`
- Add DNS entry for `myairy.mydomain.com` -> created loadbalancer
- Patch the ingress resources
```
for ingress in $(kubectl get ingress --no-headers -o custom-columns=":metadata.name"); do kubectl get ingress ${ingress} -o json |  jq '(.spec.rules[0].host="myairy.mydomain.com")' | kubectl apply -f -; done
```

And the letsencrypt certificates are created. Also redirect from https to http is configured.

### DIFF
--- a/cli/pkg/providers/aws/aws.go
+++ b/cli/pkg/providers/aws/aws.go
@@ -49,42 +49,45 @@ func (p *provider) GetOverrides() tmpl.Variables {
 	}
 }
 
-func (p *provider) PostInstallation(dir workspace.ConfigDir) error {
-	conf, err := dir.LoadAiryYaml()
-	if err != nil {
-		return err
-	}
+func (p *provider) PostInstallation(providerConfig map[string]string, dir workspace.ConfigDir) error {
+	if providerConfig["hostUpdate"] != "false" {
+		conf, err := dir.LoadAiryYaml()
+		if err != nil {
+			return err
+		}
 
-	clientset, err := p.context.GetClientSet()
-	if err != nil {
-		return err
-	}
+		clientset, err := p.context.GetClientSet()
+		if err != nil {
+			return err
+		}
 
-	ingressService, err := clientset.CoreV1().Services("kube-system").Get(context.TODO(), "traefik", metav1.GetOptions{})
-	if err != nil {
-		return err
-	}
+		ingressService, err := clientset.CoreV1().Services("kube-system").Get(context.TODO(), "traefik", metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
 
-	loadBalancerUrl := ingressService.Status.LoadBalancer.Ingress[0].Hostname
+		loadBalancerUrl := ingressService.Status.LoadBalancer.Ingress[0].Hostname
 
-	if err = p.updateIngress("airy-core", loadBalancerUrl, conf.Kubernetes.Namespace); err != nil {
-		return err
-	}
-	if err = p.updateIngress("airy-core-ui", loadBalancerUrl, conf.Kubernetes.Namespace); err != nil {
-		return err
-	}
-	if err = p.updateIngress("airy-core-redirect", loadBalancerUrl, conf.Kubernetes.Namespace); err != nil {
-		return err
-	}
+		if err = p.updateIngress("airy-core", loadBalancerUrl, conf.Kubernetes.Namespace); err != nil {
+			return err
+		}
+		if err = p.updateIngress("airy-core-ui", loadBalancerUrl, conf.Kubernetes.Namespace); err != nil {
+			return err
+		}
+		if err = p.updateIngress("airy-core-redirect", loadBalancerUrl, conf.Kubernetes.Namespace); err != nil {
+			return err
+		}
 
-	if err = p.updateHostsConfigMap(loadBalancerUrl, conf.Kubernetes.Namespace); err != nil {
-		return err
-	}
+		if err = p.updateHostsConfigMap(loadBalancerUrl, conf.Kubernetes.Namespace); err != nil {
+			return err
+		}
 
-	return dir.UpdateAiryYaml(func(conf workspace.AiryConf) workspace.AiryConf {
-		conf.Kubernetes.Host = loadBalancerUrl
-		return conf
-	})
+		return dir.UpdateAiryYaml(func(conf workspace.AiryConf) workspace.AiryConf {
+			conf.Kubernetes.Host = loadBalancerUrl
+			return conf
+		})
+	}
+	return nil
 }
 
 type KubeConfig struct {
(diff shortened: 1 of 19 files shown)


cli/pkg/providers/aws/aws.go
```diff
@@ -49,42 +49,45 @@ func (p *provider) GetOverrides() tmpl.Variables {
 	}
 }
 
-func (p *provider) PostInstallation(dir workspace.ConfigDir) error {
-	conf, err := dir.LoadAiryYaml()
-	if err != nil {
-		return err
-	}
+func (p *provider) PostInstallation(providerConfig map[string]string, dir workspace.ConfigDir) error {
+	if providerConfig["hostUpdate"] != "false" {
```
ljupcovangelski, line 53: We control with a flag whether to update the hosts after the installation or `global.kubernetes.host` is used in the ingress definitions and in the `hostnames` configMap.

cli/pkg/workspace/template/src/airy.yaml
```diff
@@ -4,14 +4,20 @@ kubernetes:
...
     {{ $k }}: {{ $v }}
   {{- end }}
-{{- end }}
+  {{- end }}
 security:
   allowedOrigins: "*"
   jwtSecret: {{ randAlphaNum 128 }}
+components:
+  sources:
+  integration:
```
ljupcovangelski, line 23: I thought it also makes sense to add these items when creating an skel `airy.yaml` with `--init-only`

docs/docs/getting-started/installation/aws.md
```diff
@@ -225,6 +229,56 @@ At this point, the frontend and the API services of `Airy Core` should be access
 airy api endpoint
 ```
 
+### HTTPS using Let's Encrypt
```
ljupcovangelski, line 232: @steffh please have a look at this section.

docs/docs/getting-started/installation/aws.md
```diff
@@ -225,6 +229,56 @@ At this point, the frontend and the API services of `Airy Core` should be access
...
+  letsencryptEmail: "mymail@myhostname.com"
+
+```
+
+The `kubernets.host` value should be set to your desired hostname. Configure the e-mail address you want to use for your Let's Encrypt registration under `ingress.letsencryptEmail`.
+
+After setting these parameters, create your `Airy Core` instance with the following option:
+
+```sh
+airy create --provider aws --provider-config hostUpdate=false
```
ljupcovangelski, line 260: @steffh if the `--provider-config` is problematic (we use it also to overwrite vpcId and instanceType for AWS), I will try to remove it.

infrastructure/helm-chart/charts/ingres/templates/ingress-controller-https-letsencrypt.yaml
```diff
@@ -0,0 +1,162 @@
...
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: traefik-ingress-controller
+  namespace: kube-system
+  labels:
+    k8s-app: traefik-ingress-lb
+spec:
+  serviceName: traefik-ingress-lb
+  replicas: 0
```
ljupcovangelski, line 62: Started with 0 replicas and instructions how to scale up after the DNS is configured, because Let's Encrypt will return error before that and might block/throttle requests in the next hour.

infrastructure/helm-chart/templates/ingress-ui.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
     traefik.frontend.rule.type: PathPrefixStrip
 spec:
   rules:
-    - host: {{ get (urlParse .Values.global.kubernetes.host) "host" }}
+    - host: {{ .Values.global.kubernetes.host }}
```
ljupcovangelski, line 11: In the `airy.yaml` file now we specify only the hostname and the protocol (http or https) is specified in `ingress.https` (true or false).

infrastructure/helm-chart/charts/ingres/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+appVersion: "1.0"
+description: A Helm chart for Airy Platform networking resources
+name: ingress
+version: 1.0
```
ljupcovangelski, line 5: Eventually the `Ingress controller` chart should be separate from `Airy Core`, so that people can customize how and if they want to deploy an ingress controller.

chrismatix (dismissed):

docs/docs/getting-started/installation/aws.md
```diff
@@ -225,6 +229,56 @@ At this point, the frontend and the API services of `Airy Core` should be access
...
+You should create a CNAME DNS record for the hostname that you set under `kubernetes.host` in the previous step to point to the hostname of the LoadBalancer, created by AWS for the ingress service:
+
+```sh
+export KUBECONFIG="PATH/TO/DIR/kube.conf"
+kubectl get --namespace kube-system service traefik --output jsonpath='{.status.loadBalancer.ingress[0].hostname}{"\n"}'
+```
+
+#### Start the ingress controller
+
+If the ingress controller is started before the DNS record is added, the Let's Encrypt servers will block an throttle the registration attempts. That is why we recommend starting the ingress controller after the DNS record is added.
```
pascal-airy, line 274:
```suggestion
If the ingress controller is started before the DNS record is added, the Let's Encrypt servers will block and throttle the registration attempts. That is why we recommend starting the ingress controller after the DNS record is added.
```

infrastructure/helm-chart/templates/configmap.yaml
```diff
@@ -4,7 +4,15 @@ metadata:
...
 data:
-  HOST: {{ .Values.global.kubernetes.host }}
+  {{ if .Values.global.ingress }}
+   {{ if .Values.global.ingress.https }}
+  HOST: https://{{ .Values.global.kubernetes.host }}
+   {{ else }}
+  HOST: http://{{ .Values.global.kubernetes.host }}
+   {{ end }}
+  {{ else }}
+  HOST: http://{{ .Values.global.kubernetes.host }}
```
pascal-airy, line 14: why is this line here twice?

ljupcovangelski: The logic is:
```
if the ingress section exists
  if ingress.https == true
    host: https...
  else (ingress.https != true)
    host: http...
else (ingress section doesn't exist)
  host: http...
```

pascal-airy: ah because the evaluation of `if .Values.global.ingress.https` would fail if `if .Values.global.ingress` isn't there

pascal-airy (approved):
👍

PR #2159, opened 2021-07-14
[#2150] Messages with same sent at time don't disappear in the message list

### LINKED ISSUE #2150: Same timestamp messages get merged in the message list
**Describe the bug**
When there are messages in the `application.communication.messages` topic that have the same `sentAt` timestamp they don't show up in the `/messages.list` endpoint

**To Reproduce**

1. On a running Airy installation
2. Produce two messages with the same timestamp

**Expected behavior**
Every message shows up in the `/messages.list` endpoint

**Additional context**
The `TreeSet` data structure we use, compares values by time and is backed by a map. Therefore values with the same timestamp simply get overwritten.

chrismatix:
resolves #2150


.github/workflows/main.yml
```diff
@@ -86,7 +86,7 @@ jobs:
         if: ${{ startsWith(github.ref, 'refs/heads/main') }}
         run: |
           sudo apt-get install -y expect
-          bazel run //frontend/chat-plugin:publish-npm release
+          bazel run //frontend/chat-plugin/lib:publish-npm release
```
chrismatix, line 89: @AitorAlgorta fixed it here

pascal-airy (approved):
good stuff 🍪

PR #2167, opened 2021-07-16
[#1211] Rename conversations.read endpoint

### LINKED ISSUE #1211: Rename /conversations.read endpoint
In PR [1199](https://github.com/airyhq/airy/pull/1199#discussion_r591512797) we discussed that the name `/conversations.read` leads to an api method called `readConversations` which is a bit confusing. We need to discuss if we want to rename this method and if, what the new name should be.

chrismatix:
resolve #1211


AitorAlgorta (approved):
🚀

PR #2169, opened 2021-07-16
[#2148] enable users to connect via instagram

### LINKED ISSUE #2148: Enable users to connect via Instagram
### Summary
To enhance overall experience, we should offer the ability for users to connect Instagram pages. The backend work for this is already completed, so we need to focus on the frontend aspect for the UI.

We should follow the same setup that we used to connect Facebook Messenger #659, however the Facebook Page ID would be Instagram Business Account ID.

If no data is entered into the optional fields, then the standard name and picture should be set to be the same as the name and picture displayed on the corresponding Instagram page.

Once the page(s) are connected, they should be displayed and we should offer the option to edit them, just like connected Facebook pages.

We should also include the Instagram [documentation](https://airy.co/docs/core/sources/instagram) underneath the Instagram header on the connection page, within the `more information about this source` button.

AudreyKj:
closes #2148 

Screenshot 2021-07-19 at 15 03 20


frontend/ui/src/components/ChannelAvatar/index.tsx
```diff
@@ -4,6 +4,7 @@ import {ReactComponent as WhatsappAvatar} from 'assets/images/icons/whatsapp_ava
 import {ReactComponent as SmsAvatar} from 'assets/images/icons/sms_avatar.svg';
 import {ReactComponent as FacebookAvatar} from 'assets/images/icons/messenger_avatar.svg';
 import {ReactComponent as AiryAvatar} from 'assets/images/icons/airy_avatar.svg';
+import {ReactComponent as InstagramIcon} from 'assets/images/icons/instagram_avatar.svg';
```
thorstenairy, line 7: would follow the current naming as InstragramAvatar instead AvatarIcon

frontend/ui/src/pages/Channels/Providers/Instagram/InstagramConnect.tsx
```diff
@@ -0,0 +1,152 @@
...
+          value={name}
+          onChange={(event: React.ChangeEvent<HTMLInputElement>) => setName(event.target.value)}
+          height={32}
+          fontClass="font-base"
+        />
+        <Input
+          id="image"
+          label="Image URL (optional)"
+          placeholder="Add an URL"
+          hint="The standard picture is the same as the Facebook Page"
```
thorstenairy, line 138: Even Instagram is Facebook, but wouldn't call it Facebook Page here.

thorstenairy (approved):
LGTM 🙏

PR #2170, opened 2021-07-19
[#1805] Fix Chatplugin reconnection

### LINKED ISSUE #1805: Bug: Chat Plugin: web socket doesn't recover
When an Airy instance is not reachable anymore (example: when it gets restarted and there is an active session via the web socket in the chat plugin) the web socket goes down and the chat plugin fails to authenticate and displays “reconnecting”, but this state never recovers. Even when the instance is up again ~30s later: the state stays like this forever, so the web socket doesn’t recover and it does not reconnect.

screenshot_2021-05-18_at_13 43 52
screenshot_2021-05-17_at_20 18 50

We can't let users wait for something that is never reconnecting.

From my understanding: either something is not working and it should actually be able to reconnect and then that should work normally, or we need to tell users something along the lines of “connection failed” - “please refresh this page in your browser”

chrismatix:
resolves #1805

### DIFF
--- a/frontend/chat-plugin/lib/src/api/index.tsx
+++ b/frontend/chat-plugin/lib/src/api/index.tsx
@@ -47,23 +47,27 @@ export const getResumeToken = async (channelId: string, authToken: string) => {
   setResumeTokenInStorage(channelId, jsonResumeToken.resume_token);
 };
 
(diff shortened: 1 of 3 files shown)


thorstenairy (approved):
LGTM

AitorAlgorta (approved):
🚀

PR #2196, opened 2021-07-22
Fix chatplugin code generator

### LINKED ISSUE #2002: Improving Chat Plugin Code Generator
**Currently:** the code generator for the Airy Live Chat Plugin has both singular and double quotation marks.

Screenshot 2021-06-16 at 07 30 54

We should change this so that all quotation marks are double.

AitorAlgorta:
closes #2002


thorstenairy (approved):
LGTM

Issue #1911, opened 2021-06-03
Reorganization of the helm charts


ljupcovangelski:
We need to re-organize the helm charts, so that core can be deployed as a separate chart, without the prerequisites. This will open the way to:
- helm upgrade.
- airy install (or `airy create --some-flag`) - resulting in creating core without creating kafka brokers and other prerequisites.
- Namespaced installation of core instances (in both Kubernetes and Kafka).
- Creating chart repositories with a specific version.

ljupcovangelski:
Done.

PR #2241, opened 2021-08-04
[#1911] Reorganize the helm charts

ljupcovangelski:
- Pushing helm charts to S3
- Introducing `helm install` installation alternative
- Enabling multiple Airy Core instances in a single Kubernets cluster
- Enabling Airy Core upgrade


.github/workflows/main.yml
```diff
@@ -71,7 +71,15 @@ jobs:
...
-
+      - name: Publish helm charts to S3
+        if: startsWith(github.ref, 'refs/heads/release') || startsWith(github.ref, 'refs/heads/main') || startsWith(github.ref, 'refs/heads/develop')
+        run: |
+          ./scripts/upload-helm-charts.sh
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          AWS_REGION : ${{ secrets.AWS_REGION }}
+          GITHUB_BRANCH: ${{ github.ref }}
```
ljupcovangelski, line 82: When merging to develop or main, we will push the packaged helm charts to s3.

cli/pkg/cmd/create/helm.go
```diff
@@ -78,10 +78,11 @@ func (h *Helm) Setup() error {
 }
 
 func (h *Helm) InstallCharts() error {
+	chartURL := "https://airy-core-helm-charts.s3.amazonaws.com/stable/airy-" + h.version + ".tgz"
```
ljupcovangelski, line 81: Instead of building our own image, a generic `alpine/helm:3.6.3` image is used with an URL with the appropriate version of the charts.

scripts/upload-helm-charts.sh
```diff
@@ -0,0 +1,36 @@
...
+helm repo add airy ${s3_basepath}
+mkdir helm-repo
+cd helm-repo
+helm package ../infrastructure/helm-chart/ --version "${version}"
+helm package ../infrastructure/helm-chart/charts/prerequisites/charts/kafka/
+helm package ../infrastructure/helm-chart/charts/prerequisites/charts/beanstalkd/
+helm package ../infrastructure/helm-chart/charts/core/ --version "${version}"
+helm package ../infrastructure/helm-chart/charts/ingress-controller/ --version "${version}"
+helm package ../infrastructure/helm-chart/charts/tools/charts/akhq/
+helm package ../infrastructure/helm-chart/charts/tools/charts/kafka-connect/
```
ljupcovangelski, line 35: We create a single package with the prerequisites, but also we will publish them separately, so that people can install only `Airy Core` without the `Prerequisites`.
Also AKHQ and KafkaConnect from now on can be installed directly via `helm install`.

infrastructure/helm-chart/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
 appVersion: "1.0"
 description: A Helm chart for Airy Core
-name: helm-charts
-version: 0.5.0
+name: airy
+version: 0-develop
```
ljupcovangelski, line 5: The version of the helm-charts is stamped at CI time. However, a version for the chart must be specified and it must start with a digit. So, I'm not sure what is the best value here.

infrastructure/helm-chart/charts/prerequisites/charts/kafka/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
 appVersion: "1.0"
 description: A kafka helm chart for Airy Platform
-name: airy
-version: 0.5.0
+name: kafka
+version: 2.7.0
```
ljupcovangelski, line 5: The version of the Kafka chart, should be in accordance with the version of Kafka.

tools/build/bazel_status.sh
```diff
@@ -12,10 +12,6 @@ else
   BRANCH=$(git branch --show-current)
 fi
 
-if [[ $BRANCH == "develop" ]]; then
-  echo STABLE_VERSION "develop"
-else
-  echo STABLE_VERSION $(cat ./VERSION)
-fi
+echo STABLE_VERSION $(cat ./VERSION)
```
ljupcovangelski, line 15: I removed the version `develop` for the images and the cli, because the `helm charts` cannot have a version `develop`. The version must start with a digit, so we will use the actual version (example: 0.28.0-alpha).

scripts/dev_cli.sh
```diff
@@ -4,8 +4,9 @@ IFS=$'\n\t'
 
 brew install -q coreutils wget
 
+VERSION=$(cat ./VERSION)
 AIRY_BIN=/usr/local/bin/airy_dev
-BUCKET=https://airy-core-binaries.s3.amazonaws.com/develop/darwin/amd64
+BUCKET=https://airy-core-binaries.s3.amazonaws.com/${VERSION}/darwin/amd64
```
pascal-airy, line 9: for that to work, I think you also need to change the upload-cli-binary.sh to upload an alpha cli binary on merge to develop

scripts/upload-helm-charts.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+set -eo pipefail
+IFS=$'\n\t'
+
+bucket_name="airy-core-helm-charts"
+version=$(cat ./VERSION)
+
+case ${GITHUB_BRANCH} in
+refs/heads/develop)
+    s3_basepath=s3://${bucket_name}/stable
```
pascal-airy, line 10: develop is saved to `stable`?

scripts/dev_cli.sh
ljupcovangelski: Correct. I think this is it
https://github.com/airyhq/airy/pull/2241/files#diff-10eb59b0cdafa7a3fe250423fffb995b31a284266a6b04191eb090e92f9b4276

scripts/upload-helm-charts.sh
ljupcovangelski: Correct. It will not cause a problem for the moment, as VERSION is always used when fetching the helm chart version. I opened an issue however to fix this in the future.
https://github.com/airyhq/airy/issues/2256

scripts/dev_cli.sh
pascal-airy: this only changes which Version is backed into the binary but the file itself is not created by the CI I think

ljupcovangelski: You are rignt, this is for the bucket where the binary is actually uploaded to and it differs from that is set in the `upload-cli-binary.sh` script :+1: . I will fix it.

ljupcovangelski: Ok, I think this solves it https://github.com/airyhq/airy/pull/2241/commits/8e5a0498dcb0e81fd236db527338ddeda8e472f4

pascal-airy (approved):
LGTM

PR #2277, opened 2021-08-11
More configs for chatplugin

### LINKED ISSUE #2272: Chat Plugin: Config Options
In terms of customizing the Live Chat for a website(s), there should be more config options readily available to improve the overall user experience.

### Relating to both desktop and web browser:
- [x] The option to have the Live Chat always expanded, or have it minimized unless a website visitor clicks on it
- [x] The option to customize the size of the Live Chat, height and width. For this option, we should offer the system user the generic dimensions and allow them to edit if they want to
- [x] The option to disable the Live Chat on mobile browsers

### For discussion:
- [x] Set the widget to not open on first page load on mobile version -- also have a height setting in the css of "-webkit-fill-available" with a fallback of "100vh" for browsers that don't support that

AitorAlgorta:
closes #2272


chrismatix (approved):
Good change 👍

Issue #2243, opened 2021-08-05
Improve the Airy upgrade process


steffh:
**Is your feature request related to a problem? Please describe.**

As an Airy Open Source user I want to be able to upgrade my Airy instance to a new version.

At the moment this is accomplished by running several commands sequentially and amending config files manually.

**Describe the solution you'd like**

A CLI command such as `airy upgrade` would be nice to run instead of individual commands.

The upgrade process also should be more visibly documented in the Airy docs.

**Describe alternatives you've considered**
-

**Additional context**
-

ljupcovangelski:
So if I understand correctly, if somebody wants to upgrade `Airy Core` to a new version, the procedure is the following:
- You need to download new version of the CLI
- Run `airy upgrade`. You will get information if upgrade is possible for your existing `Airy Core` version or you first need to upgrade to an intermediate version.
- The command will do the upgrade and will update your `airy.yaml` file with the new version, after the upgrade completes successfully. 
- If the upgrade is not successful - you will be reverted to your previous version.

chrismatix:
Agree with the sequence of steps.

> Do we tie the upgrade to the version which is bundled in the CLI? Let's say your CLI version is 0.30.0 and your Airy Core cluster is 0.25.0. What if you want to upgrade to 0.27.0. Should we allow airy upgrade --version 0.27.0? If the --version flag is not set - we default to the version that is in the CLI.

I would definitely tie it to the latest version of the CLI, so no `--version` flag. Anything else would be much more expensive.

> Will upgrade only concern the version change? 

Upgrades necessarily also entail changes to Kubernetes and Kafka infrastructure such as adding routes, deployments, resetting and creating topics etc. 

To do this I would propose the following: Running `airy upgrade` does not execute the upgrade immediately. Rather it returns a command that can be run with `$(airy upgrade)`. Pseudo behavior:

```bash
$ airy upgrade
$ /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/airyhq/airy/scripts/upgrades/0.27.0/run.sh)"
```

Benefits:
- We can fix migration scripts after releases by pushing to the cdn URL 
- We can leverage bash scripts for our upgrades the way we have been doing it in cloud

ljupcovangelski:
> Upgrades necessarily also entail changes to Kubernetes and Kafka infrastructure such as adding routes, deployments, resetting and creating topics etc.
> 
[...]

PR #2292, opened 2021-08-16
[#2243] Introduce airy upgrade

ljupcovangelski:
The `airy upgrade` command currently produces the following output
```
$ airy upgrade
CLI version:  0.29.0-alpha
Current Airy Core version:  0.28.0
New Airy Core version:  0.29.0-alpha

Are you sure that you want to proceed? [Y/y]
y

Upgrading the helm charts of Airy Core...
Applying config from the configuration file.
applied configuration for "security"
Writing the new version into the configuration file.
Copying the configuration file in the Airy Core K8s cluster.

✅ Aity Core upgraded
```
- [x] Perform helm upgrade
- [x] Roll-back if the upgrade wasn't successful
- [x] Example definition of the migration jobs
- [x] Docs

### DIFF
--- a/cli/pkg/cmd/upgrade/upgrade.go
+++ b/cli/pkg/cmd/upgrade/upgrade.go
@@ -1,0 +1,101 @@
+package upgrade
+
+import (
+	"cli/pkg/cmd/config"
+	"cli/pkg/console"
+	"cli/pkg/helm"
+	"cli/pkg/kube"
+	"cli/pkg/workspace"
+	"fmt"
+
+	"github.com/airyhq/airy/infrastructure/lib/go/k8s/util"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+var (
+	approve    bool
+	version    string
+	Version    string
+	UpgradeCmd = &cobra.Command{
+		Use:   "upgrade",
+		Short: "Upgrades an instance of Airy Core",
+		Long:  `Upgrades an existing Airy Core instance, read from a workspace directory (default .)`,
+		Args:  cobra.MaximumNArgs(1),
+		Run:   upgrade,
+	}
+)
+
+func init() {
+	UpgradeCmd.Flags().StringVar(&version, "version", "", "Specify a version to upgrade to (optional).")
+	UpgradeCmd.Flags().BoolVar(&approve, "approve", false, "Upgrade automatically without asking for an approval (optional).")
+}
+
+func upgrade(cmd *cobra.Command, args []string) {
+	autoApprove, _ := cmd.Flags().GetBool("approve")
+	workspacePath, err := cmd.Flags().GetString("workspace")
+	if err != nil {
+		console.Exit("Unable to find suitable workspace :", err)
+	}
+	dir := workspace.Init(workspacePath)
+	kubeCtx := kube.Load()
+	clientset, err := kubeCtx.GetClientSet()
+	namespace := viper.GetString("namespace")
+	if err != nil {
+		console.Exit("Could not find an installation of Airy Core. Get started here https://airy.co/docs/core/getting-started/installation/introduction")
+	}
+
+	if version == "" {
+		version = Version
+	}
+	clusterAiryConf, err := kube.GetDeployedAiryYaml(namespace, clientset)
+	if err != nil {
+		console.Exit("Unable to retrieve existing version of Airy Core: ", err.Error())
+	}
+	oldVersion := clusterAiryConf.Kubernetes.AppImageTag
+
+	fmt.Println("CLI version: ", Version)
+	fmt.Println("Current Airy Core version: ", oldVersion)
+	fmt.Println("New Airy Core version: ", version)
+	fmt.Println()
+
+	if autoApprove != true {
+		if util.ConfirmToProceed() == false {
+			console.Exit("Upgrade aborted.")
+		}
+	}
+
+	fmt.Println("Upgrading the helm charts of Airy Core...")
+	helm := helm.New(clientset, version, namespace, dir.GetAiryYaml())
+	if cmErr := helm.UpsertAiryConfigMap(); cmErr != nil {
+		console.Exit("Unable to copy config file into Airy Core :", cmErr)
+	}
+	if upgradeErr := helm.UpgradeCharts(); upgradeErr != nil {
+		fmt.Println("Upgrading the Helm Charts failed with err: ", upgradeErr)
+		fmt.Println("Attempting to roll back the upgrade...")
+		if rollBackErr := helm.RollBackUpgrade(oldVersion); rollBackErr != nil {
+			console.Exit("Roll Back of the upgrade failed with err: ", rollBackErr)
+		}
+		console.Exit("The roll back of the upgrade was successful.")
+	}
+
+	fmt.Println("Applying config from the configuration file.")
+	config.ApplyConfig(workspacePath)
+
+	fmt.Println(("Writing the new version into the configuration file."))
+	if writeVersionErr := dir.UpdateAiryYaml(func(conf workspace.AiryConf) workspace.AiryConf {
+		conf.Kubernetes.AppImageTag = version
+		return conf
+	}); writeVersionErr != nil {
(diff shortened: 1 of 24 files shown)


cli/pkg/cmd/upgrade/upgrade.go
```diff
@@ -0,0 +1,110 @@
...
+	clientset, err := kubeCtx.GetClientSet()
+	namespace := viper.GetString("namespace")
+	if err != nil {
+		console.Exit("could not find an installation of Airy Core. Get started here https://airy.co/docs/core/getting-started/installation/introduction")
+	}
+
+	if version == "" {
+		version = Version
+	}
+	oldVersion := conf.Kubernetes.AppImageTag
```
ljupcovangelski, line 56: Not sure if we should read the current version from the `airy.yaml` file or from the `airy-config-map` configMap, inside the Kubernetes cluster.

cli/pkg/cmd/upgrade/upgrade.go
```diff
@@ -0,0 +1,110 @@
...
+
+	fmt.Println("Upgrading the helm charts of Airy Core...")
+	helm := helm.New(clientset, version, namespace, dir.GetAiryYaml())
+	if cmErr := helm.UpsertAiryConfigMap(); cmErr != nil {
+		console.Exit("Unable to copy config file into Airy Core.")
+	}
+	if upgradeErr := helm.UpgradeCharts(); upgradeErr != nil {
+		fmt.Println("Upgrading the Helm Charts failed with err: ", upgradeErr)
+		fmt.Println("Attempting to roll back the upgrade...")
+		if rollBackErr := helm.RollBackUpgrade(oldVersion); rollBackErr != nil {
```
ljupcovangelski, line 75: The roll-back is initiated only if `helm upgrade` fails. For the other steps, I'm not sure if it is needed or wise to do that.

cli/pkg/helm/helm.go
```diff
@@ -1,4 +1,4 @@
-package create
+package helm
```
ljupcovangelski, line 1: Moved this code in a separate package outside of `cli/pkg/cmd/create`, so that we can re-use it.

infrastructure/helm-chart/charts/core/charts/provisioning/templates/job-wait.yaml
```diff
@@ -3,7 +3,7 @@ kind: Job
 metadata:
   name: wait-for-api-communication
   annotations:
-    "helm.sh/hook": "post-install"
+    "helm.sh/hook": "post-install,post-upgrade"
```
ljupcovangelski, line 6: Wait also after the upgrade, not only after the installation.

cli/pkg/cmd/upgrade/upgrade.go
chrismatix: The one from the cluster makes more sense to me, since it should reflect the current state. The config file always represents the "should" state.

cli/pkg/cmd/upgrade/upgrade.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+var (
+	approve    bool
+	version    string
+	Version    string
+	UpgradeCmd = &cobra.Command{
+		Use:   "upgrade [workspace directory]",
```
chrismatix, line 22: Let's keep the workspace directory as a flag `--workspace` to keep it consistent with other commands that operate on an existing workspace.

cli/pkg/cmd/upgrade/upgrade.go
```diff
@@ -0,0 +1,110 @@
...
+	fmt.Println("Copying the configuration file in the Airy Core K8s cluster.")
+	if cmErr := helm.UpsertAiryConfigMap(); cmErr != nil {
+		console.Exit("Unable to copy config file into Airy Core.")
+	}
+
+	fmt.Println("✅ Airy Core upgraded")
+	fmt.Println()
+}
+
+func explainHost(hostName string) string {
```
chrismatix, line 101: This seems unused

infrastructure/helm-chart/charts/core/charts/upgrading/templates/scripts.yaml
```diff
@@ -0,0 +1,38 @@
...
+    done
+
+  reset-streaming-app.sh: |
+    #!/bin/sh
+    kafka_brokers=${KAFKA_BROKERS:-kafka:9092}
+    input_topics=${INPUT_TOPICS}
+    consumer_group=${CONSUMER_GROUP:-1}
+    delay=${1:-10}
+
+    kafka-streams-application-reset.sh --bootstrap-servers ${kafka_brokers} --to-earliest --application-id ${consumer_group} --input-topics ${input_topics}
```
ljupcovangelski, line 38: When we want to add a migration script, we need to create a file with the version (example 0.29.0.yaml) in the templates directory and add a block which creates the resources conditionally:
```
{{ if eq .Values.global.kubernetes.appImageTag "0.28.0" }}
  (bash scripts run from inside the cluster, using the upgrading-scripts as helper scrips)
{{ end }}
```

chrismatix: That makes sense. How would one clean up the resources post migration?

ljupcovangelski: I guess we need to also do that inside the migration jobs (have a post-upgrade job that waits until stuff is done and cleans up). In case there needs to be a migration app that must run longer then the actual upgrade, then we can also deploy a `CronJob` that checks periodically and after the migration app needs to be deleted it also clears itself.

docs/docs/getting-started/installation/aws.md
```diff
@@ -260,6 +260,16 @@ After setting these parameters, create your `Airy Core` instance with the follow
...
+:::note
+In case you have created your `Airy Core` instance without Let's Encrypt and want to add it later, you must use the `airy upgrade` command.
+
+Even if you don't upgrade to a new version, just modify your `airy.yaml` file as explained in this section and run
+
+`airy upgrade`
+
+After the upgrade is done, continue with setting up your DNS and starting the ingress controller.
+:::
+
```
ljupcovangelski, line 272: @steffh this PR also makes it possible for Let's Encrypt to be enabled after installation. Please have a look at this doc.
https://deploy-preview-2292--docs-airy.netlify.app/getting-started/installation/aws#https-using-lets-encrypt

infrastructure/helm-chart/charts/core/charts/upgrading/templates/scripts.yaml
chrismatix: Understood. For now, that seems like lots of added complexity only to, say, remove a Java pod. Perhaps it's ok to just include that part in the release notes for now. What do you think?

docs/docs/getting-started/installation/aws.md
```diff
@@ -260,6 +260,16 @@ After setting these parameters, create your `Airy Core` instance with the follow
...
 ```
 
+:::note
+In case you have created your `Airy Core` instance without Let's Encrypt and want to add it later, you must use the `airy upgrade` command.
+
+Even if you don't upgrade to a new version, just modify your `airy.yaml` file as explained in this section and run
+
+`airy upgrade`
+
+After the upgrade is done, continue with setting up your DNS and starting the ingress controller.
```
ljupcovangelski, line 270:
```suggestion
After the upgrade is done, continue with setting up your DNS and starting the ingress controller something else.
```

infrastructure/helm-chart/charts/core/charts/upgrading/templates/scripts.yaml
ljupcovangelski: I would leave the part for the migration/cleanup jobs until we actually have an upgrade with a migration. Then we will see with an actual example how this will work.
I think for all the migration resources we can add some specific annotations and at the end we can use a generic cleanup job.

ellaknox (changes requested):

pascal-airy (approved):
lgtm. Good job 👍

Issue #2255, opened 2021-08-09
Docs for deploying Airy Core with helm


ljupcovangelski:
Add docs for how to deploy `Airy Core` with helm:
- Deploy `Airy Core` on an existing Kubernetes cluster. 
- Run with your own Kafka/Prerequisites.
- Use your own ingress controller.
- Deploy tools such as AKHQ and KafkaConnect.

chrismatix:
🌈 [Inspiration](https://rasa.com/docs/rasa-x/installation-and-setup/install/helm-chart/) 🌈

ljupcovangelski:
At the moment `airy.yaml` isn't suitable for using it as a `values` file for `helm`, because we don't have a `global:` section.

There are few options on addressing this issue:
1. We add a flag to airy create that will output also a helm version of the file. For example `airy create --init-only --geenrate-helm` will create both `airy.yaml` and `helm.yaml` that can later be used for doing `helm install`. `airy config apply` will still use the `airy.yaml` file, so it is a bit strange.
2. We write in the docs that in order for users to be able to install airy with helm, they need to transform the file on their own. So the process is: 1) run `airy create --init-only` 2) run something like `echo "global:" > helm.yaml; sed 's/^/  /g' airy.yaml >> helm.yaml` 3) do `helm install ... --values helm.yaml`
3. We can put the `global` (now kubernetes.global) values into a separate file. So that we have one file for global stuff, that is not used for airy config apply . I think we discussed also to have even separate files for `prerequisites`, etc, so I'm not sure. When we use `terraform` for creating `Airy Core` these global settings will be probably put in variables, so also from that perspective it is reasonable to have them separated from what is in `airy.yaml`. 
4. We revert `kubernetes` back to `global` in the `airy.yaml` file and then the same file can be used for helm install/upgrade.
5. We introduce `airy install` that only takes care of the `helm install` part, after there is an existing cluster created. That would be transforming the `airy.yaml` file and running `helm install`.

What do we have in the kubernetes/global section:
```
appImageTag: 0.29.0-alpha
containerRegistry: ghcr.io/airyhq
namespace: default
host: airy.core
ngrokEnabled: false
```
So I think that only the first three should stay and ngrok and host should be moved into an ingress section which shouldn't be under global. We can get rid of the host, as it should only be used for `allowedOrigins` and `when you have a certificate`.

ljupcovangelski:
[...]

PR #2341, opened 2021-08-27
[#2255] Modular deployment of Airy with helm

ljupcovangelski:
Deploy Airy with helm. Use the `airy.yaml` file as `--values` file for helm.

- [x] Add dependencies between the man chart and the subcharts
- [x] Remove appImageTag from global
- [x] Remove namespace from global
- [x] Remove containerRegistry from global
- [x] Remove host from global and move it to ingress chart
- [x] Remove namespace from all resources (namespacing is done by Helm)
- [x] Move ngrokEnabled and ngrok resources to ingress chart
- [x] Add docs

We should be able to deploy Airy modularly, to a different namespace with helm
```
helm install airy --values airy.yaml --set appImageTag=0.29.0 --set global.security.systemToken=test \
--set prerequisites.kafka.enabled=false \
--set tools.akhq.enabled=true \
--set tools.kafka-connect.enabled=false
```


infrastructure/helm-chart/values.yaml
```diff
@@ -1 +1,20 @@
...
+    commitInterval: 1000
+provisioning:
+  kafka:
+    image: ghcr.io/airyhq/infrastructure/kafka
+    imageTag: 2.7.0
+tools:
+  segment:
+    enabled: true
+serviceAccount: airy-admin
+apiHost: http://airy.core
```
ljupcovangelski, line 20: We are generating some of the defaults with the `cli` now, but if we want for people to be able to install Airy with helm, without using the cli, I think the defaults should be in the chart.
Otherwise we should instruct people to do `helm create --init-only` (with some provider) and then run `helm install --values ./airy.yaml`.

infrastructure/helm-chart/templates/config/configmap.yaml
```diff
@@ -0,0 +1,21 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: core-config
+data:
+  APP_IMAGE_TAG: {{ default .Chart.Version .Values.global.appImageTag }}
+  CORE_ID: {{ randAlphaNum 10 | lower }}
+  CHATPLUGIN_JWT_SECRET: {{ randAlphaNum 128 | quote }}
+  API_HOST: {{ .Values.apiHost }}
+  SEGMENT_ANALYTICS_ENABLED: {{ .Values.tools.segment.enabled }}
```
pascal-airy, line 21: thanks 🙌

cli/pkg/helm/helm.go
```diff
@@ -219,18 +206,6 @@ func (h *Helm) UpsertAiryConfigMap() error {
...
-func airyYamlToHelmValues(content string) string {
-	scanner := bufio.NewScanner(strings.NewReader(content))
-	var builder strings.Builder
-	builder.WriteString("global:\n")
-	for scanner.Scan() {
-		builder.WriteString("  " + scanner.Text() + "\n")
-	}
-	return builder.String()
-}
-
```
ljupcovangelski, line 233: No need to create a new file.

PR #2376, opened 2021-09-06
[#2358] preventing second enter when sending messages

### LINKED ISSUE #2358: Prevent second 'Enter' or double click when sending messages
Currently in the Inbox, a system user is able to send a message twice (or even three times) if they accidentally click the `Send` button more than once or press `Enter` more than once.

We should disable this so that users are prevented from accidentally resending the same message within a few seconds. We could possibly run a script in the background to block the button from being clicked multiple times, possibly within 3 or 5 seconds - happy to discuss this.

AudreyKj:
closes #2358

### DIFF
--- a/frontend/ui/src/pages/Inbox/MessageInput/index.tsx
+++ b/frontend/ui/src/pages/Inbox/MessageInput/index.tsx
@@ -139,9 +139,10 @@ const MessageInput = (props: Props) => {
   }, [selectedTemplate]);
 
   const sendMessage = () => {
-    if (!channelConnected) {
+    if ((!selectedTemplate && !selectedSuggestedReply && !input) || !channelConnected) {
       return;
     }
+
     setSelectedSuggestedReply(null);
     setSelectedTemplate(null);
     sendMessages(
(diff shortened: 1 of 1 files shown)


frontend/ui/src/pages/Inbox/MessageInput/index.tsx
```diff
@@ -139,9 +139,10 @@ const MessageInput = (props: Props) => {
   }, [selectedTemplate]);
 
   const sendMessage = () => {
-    if (!channelConnected) {
+    if ((!selectedTemplate && !selectedSuggestedReply && !input) || !channelConnected) {
       return;
```
AudreyKj, line 143: we return when the user clicks on the send button and there isn't anything to be send (i.e. when the input text, the template selector and suggestedReply are empty)

chrismatix: There is also already this [line](https://github.com/airyhq/airy/blob/b00bf27e959d4b6304f2c9e787e9ee09a1de4966/frontend/ui/src/pages/Inbox/MessageInput/index.tsx#L354). Perhaps we could turn it into a function `canSendMessage` and then re-use it in both places? 🤔

AudreyKj: good catch 👍 fixing now!

chrismatix (approved):

Issue #2390, opened 2021-09-08
Upgrade ingress controller


ljupcovangelski:
The ingress controller which we are using by default - Traefik v1.7 is not compatible with the latest Kubernets versions 1.22+.
We should upgrade Traefik, however v2 needs the resource `IngressRoute` instead of `Ingress` because of the `PathPrefixStrip` that we are using.

If we want to stick with the standard `Ingress` definition, then we should go for another ingress controller, probably Nginx.

PR #2392, opened 2021-09-09
[#2390] Upgrade traefik ingress controller

ljupcovangelski:


ljupcovangelski:
We changed the ingress controller to nginx-ingress.

PR #2409, opened 2021-09-17
Improve AWS docs

ljupcovangelski:
- [x] Add information about S3 buckets that should be publicly accessible (media-resolver component)
- [x] Add information about the roles and policy attachments we are creating
- [x] Add information about where all the data (sensitive data) is kept
- [x] Add exact information about which AWS services are used and reference to their pricing models.
- [x] Add a section for backup and recovery
- [x] Add guidance on managing AWS service limits


pascal-airy (approved):
lgtm! nice job 👍

PR #2426, opened 2021-09-20
[#628] make library compatible with node.js

### LINKED ISSUE #626: refactor library's module with a class

### LINKED ISSUE #628: make library compatible with node.js and the browser
**Current state:** 
- the library is only compatible with the browser

**To do:** 
- make library compatible with both node + browser 
- change api request config to make it compatible (fetch vs axios vs superagent to be discussed)
- add appropriate bundling (Webpack)

AudreyKj:
closes #628

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     "emoji-mart": "3.0.1",
     "linkifyjs": "^2.1.9",
     "lodash-es": "^4.17.21",
+    "node-fetch": "^2.0.0",
     "preact": "^10.5.14",
     "react": "^17.0.2",
     "react-autosize-textarea": "^7.1.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -16,6 +16,7 @@
     "emoji-mart": "3.0.1",
     "linkifyjs": "^2.1.9",
     "lodash-es": "^4.17.21",
+    "node-fetch": "^2.0.0",
     "preact": "^10.5.14",
```
AudreyKj, line 20: I've installed v2 because v3 is buggy (webpack pbls)
https://github.com/node-fetch/node-fetch/issues/1265

AitorAlgorta (approved):
did you test it? works?

PR #2430, opened 2021-09-22
[#2423] track core installation

### LINKED ISSUE #2423: Track Airy Core Installation
- Start Installation
- Finish Installation

some information about the kind of installation (minikube/aws)
Airy Core Version

pascal-airy:
Resolves #2423

### DIFF
--- a/cli/pkg/cmd/create/create.go
+++ b/cli/pkg/cmd/create/create.go
@@ -60,8 +60,8 @@ func create(cmd *cobra.Command, args []string) {
 	airyAnalytics := console.NewAiryAnalytics(disableTracking)
 	airyAnalytics.Track(analytics.Track{
 		AnonymousId: "AiryUser",
-		Event:  "installation_started",
-	  })
+		Event:       "installation_started",
+	})
 	provider := providers.MustGet(providers.ProviderName(providerName), w, airyAnalytics)
 	overrides := provider.GetOverrides()
 	overrides.Version = version
@@ -136,10 +136,8 @@ func create(cmd *cobra.Command, args []string) {
 	}
 
 	airyAnalytics.Track(analytics.Track{
-		Event:  "installation_succesful",
-		Properties: analytics.NewProperties().
-		  Set("plan", "Enterprise"),
-	  })
+		AnonymousId: "AiryUser",
+		Event: "installation_succesful"})
 	fmt.Printf("📚 For more information about the %s provider visit https://airy.co/docs/core/getting-started/installation/%s", providerName, providerName)
 	fmt.Println()
 }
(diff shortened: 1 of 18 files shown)


ljupcovangelski (commented):
Should we append a certain unique-id for each installation, so that we can connect it to other events in the future? Note that that is not traceable to the user or any user data, just so that we can put together various events by the same installation. Currently we generate a `unique` string with helm, so perhaps we can move that to the CLI and overwrite the helm variable. If the overwrite doesn't come - then helm should also generate this value (in case Airy is created directly with helm or Terraform and without the CLI).

ljupcovangelski (commented):
What about events in case the installation is not successful?

cli/pkg/cmd/create/create.go
```diff
@@ -49,6 +53,20 @@ func create(cmd *cobra.Command, args []string) {
...
 		return color.Colorize(color.Cyan, "#\t"+input)
 	})
+
+	client := analytics.New("7OOYJk8lQ2jfmZXKz5jJuvcK50BNDHit")
+	var rtm runtime.MemStats
+	runtime.ReadMemStats(&rtm)
+	track(client, analytics.Identify{
+		Traits: analytics.NewTraits().
+			Set("provider", providerName).
+			Set("numCpu", runtime.NumCPU()),
```
ljupcovangelski, line 63: This is about the CPU on the machine that runs the CLI right? So it is only meaningful for Minikube.

cli/pkg/cmd/create/create.go
```diff
@@ -37,6 +40,7 @@ func init() {
 	CreateCmd.Flags().StringVar(&namespace, "namespace", "default", "(optional) Kubernetes namespace that Airy should be installed to.")
 	CreateCmd.Flags().BoolVar(&initOnly, "init-only", false, "Only create the airy workspace directory and exit.")
 	CreateCmd.Flags().BoolVar(&noApply, "no-apply", false, "Don't apply any component configuration found in an existing airy.yaml file after creation.")
+	CreateCmd.Flags().BoolVar(&disableTracking, "disable-tracking", false, "Disables sending any events to Segment.")
```
chrismatix, line 43:
```suggestion
	CreateCmd.Flags().BoolVar(&disableTracking, "disable-tracking", false, "Disables sending events to Segment.")
```

infrastructure/terraform/demos/analytics/main.tf
```diff
@@ -75,7 +75,7 @@ data "template_file" "values_yaml" {
 
 resource "helm_release" "airy_core" {
   name  = "airy-release"
-  chart = "https://airy-core-helm-charts.s3.amazonaws.com/stable/airy-${replace(file("${path.module}/../../../../VERSION"), "\n", "")}.tgz"
```
chrismatix, line 78: I suppose that this should be reverted before merging

pascal-airy: got it 👍

docs/docs/cli/usage.md
```diff
@@ -74,6 +74,7 @@ airy create [workspace directory] [flags]
 #### Options
 
 ```
+      --disable-tracking                 Disables sending events to Segment.
```
ljupcovangelski, line 77: Perhaps we can add sending `anonymized` events to Segment.

ljupcovangelski (approved):
Looks good :+1:

PR #2431, opened 2021-09-22
Add more chatplugin options 

### LINKED ISSUE #2393: Live Chat Plugin Config Option: Remove Input Bar
As an Airy system user, I should have the option to configure my Live Chat Plugin in a way that the input bar can be switched on or off in the Chat Plugin customization page. The default mode would be that the input bar is visible and interactive.

This could be useful for users who only want their contacts to follow a chain of templates or bots rather than have a conversation.

AitorAlgorta:
closes #2393


AudreyKj (dismissed):

thorstenairy (approved):
LGTM

PR #2447, opened 2021-09-27
[#2424] track channel connected

### LINKED ISSUE #2424: Track Channel Connected
Segment Java library add to ChannelController in ApiAdmin

- Connect/Disconnect
- Channel type

pascal-airy:
Resolves #2424


backend/api/admin/src/main/resources/application.properties
```diff
@@ -3,3 +3,6 @@ kafka.schema-registry-url=${KAFKA_SCHEMA_REGISTRY_URL}
 kafka.cleanup=${KAFKA_CLEANUP:false}
 kafka.commit-interval-ms=${KAFKA_COMMIT_INTERVAL_MS}
 kubernetes.namespace=${KUBERNETES_NAMESPACE}
+segment.analytics.enabled=${TRACKING_ENABLED}
```
chrismatix, line 6: No need to put this in the `application.properties` explicitly we have found. You can simply directly reference the variable in the `@Value` annotation.

infrastructure/helm-chart/charts/core/charts/components/charts/api/charts/api-admin/templates/deployment.yaml
```diff
@@ -61,6 +61,16 @@ spec:
             configMapKeyRef:
               name: beanstalk-config
               key: BEANSTALK_PORT
+        - name: CORE_ID
```
chrismatix, line 64: I'd propose to reference this like the security config map:
```yaml
envFrom:
  - configMapRef:
     name: security
```
to avoid redundantly spelling out variables that every app may depend on.

lib/java/spring/web/src/main/java/co/airy/spring/web/config/SegmentAnalyticsAutoConfiguration.java
```diff
@@ -0,0 +1,32 @@
...
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+
+@Configuration
+@EnableConfigurationProperties(SegmentProperties.class)
+@ConditionalOnProperty("segment.analytics.enabled")
```
chrismatix, line 15: This annotation belongs on the `@Bean` constructor `L22`

lib/java/spring/web/src/main/java/co/airy/spring/web/filters/RequestLoggingFilter.java
```diff
@@ -31,6 +35,12 @@
     private final ObjectMapper objectMapper;
     private final RequestLoggingIgnorePatterns requestLoggingIgnorePatterns;
 
+    @Value("${core.id")
+    private String coreId;
+
+    @Autowired(required = false)
```
chrismatix, line 41: As a principle we only use the `@Autowired` annotation in tests and the [constructor injection](https://reflectoring.io/constructor-injection/) everywhere else.

lib/java/spring/web/src/main/java/co/airy/spring/web/filters/RequestLoggingFilter.java
```diff
@@ -48,6 +58,14 @@ protected void doFilterInternal(HttpServletRequest request, HttpServletResponse
         } finally {
             if ("POST".equals(requestToUse.getMethod())) {
                 log.info(createMessage(requestToUse));
+                if (analytics != null && requestToUse.getRequestURI().endsWith(".connect")) {
```
chrismatix, line 61: Not a big fan of having a piece of generic logging middleware in a library depend on business logic. `airy-core/lib/` should be decoupled from everything `airy-core/backend`.

Here's an alternative proposal for still keeping this generic:
1. Create a separate tracking library in `lib/java/spring`
2. Have a component `TrackingRequestFilter` that works like this request logging filter
3. Library users can supply a (list of) bean(s) `RouteTracking` that a) match the request url/method and b) return a `TrackMessage`
4. Inject the `RouteTracking` beans supplied by the library users into the `TrackingRequestFilter`

Thereby you have a generic library that apps can use to track route calls, while keeping the tracking (business) logic separate. `RouteTracking` would be the API contract.

- Names are of course open to change
- There are more ways of designing this than the way I described above. The important part is to separate the "Whenever there is call that matches predicate X do Y" logic from the implementation of X and Y

chrismatix (changes requested):

ljupcovangelski (approved):
Great work :+1:

chrismatix (approved):
Very cool 👏 

You could also replace the `ConditionalOnProperty` on the `RouteTracking` beans with `Lazy` since it's up to the library to load them. Then you don't have to check for `segment.analytics.enabled` everywhere :)

PR #2449, opened 2021-09-28
[#2445] fix http client library compatibility with node.js

### LINKED ISSUE #2445: httpclient library node.js compatibility
**To do**
--> fix httpclient library compatibility with node.js

AudreyKj:
closes #2445 

**Changes** 
- added necessary packages used in the library 
- [ Regenerator](https://github.com/facebook/regenerator) added because using the library in Node.js produced the [error
`ReferenceError: regeneratorRuntime is not defined`](https://github.com/babel/babel/issues/9849)

### DIFF
--- a/lib/typescript/httpclient/BUILD
+++ b/lib/typescript/httpclient/BUILD
@@ -18,7 +18,6 @@ ts_web_library(
         "@npm//camelcase-keys",
         "@npm//node-fetch",
         "@npm//regenerator-runtime",
-        "@npm//core-js",
     ],
 )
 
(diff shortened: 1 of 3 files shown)


lib/typescript/httpclient/BUILD
```diff
@@ -17,6 +17,7 @@ ts_web_library(
         "@npm//@types/node",
         "@npm//camelcase-keys",
         "@npm//node-fetch",
+        "@npm//regenerator-runtime",
```
chrismatix, line 20: @AudreyKj What do we need this for? 🤔 

Node environments should support `async` out of the box.

PR #2451, opened 2021-09-29
[#2434] Fix instagram echo ingestion

### LINKED ISSUE #2434: Instagram: Echo Webhook is not properly ingested
## Describe the bug

As an Airy user having `instagram` sources connected,
when I receive an echo webhook at my Airy instance (e.g. because an agent replied in the Facebook Page Inbox to an Instagram message),
then such echo message seems to be not properly ingested as it doesn't show up as a message in the relevant conversation in Airy

## To Reproduce

1. To an Instagram Message reply via Facebook Page Inbox (https://business.facebook.com/BUSINESSNAME/inbox) 
2. Wait for the echo to appear in the relevant conversation in Airy
3. Notice it doesn't get properly appended to the relevant conversation in Airy

## Expected behavior

- An echo webhook (https://developers.facebook.com/docs/messenger-platform/instagram/features/webhook/#echo-webhook) should be properly ingested and appear in the relevant conversation (comparable with the behavior of Facebook Messenger) 
- In case an outbound message has been sent by the Airy Inbox or API and appended to the relevant conversation already, an echo should not be appended for the same message

## Environment

- Airy version 0.30.0

chrismatix:
resolves #2434

### DIFF
--- a/backend/sources/facebook/events-router/src/main/java/co/airy/core/sources/facebook/MessageMapper.java
+++ b/backend/sources/facebook/events-router/src/main/java/co/airy/core/sources/facebook/MessageMapper.java
@@ -73,7 +73,7 @@ public class MessageMapper {
         } else if (appId != null && !appId.equals(this.facebookAppId)) {
             // Third party app
             senderId = appId;
-        } else if (appId == null && !"instagram".equals(channel.getSource())) {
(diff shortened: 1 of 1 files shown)


thorstenairy (approved):
LGTM

PR #2467, opened 2021-10-06
[#2424] Fix defaulting of the segment variable

chrismatix:
follow up to #2424

### DIFF
--- a/backend/sources/facebook/connector/src/main/java/co/airy/core/sources/facebook/Connector.java
+++ b/backend/sources/facebook/connector/src/main/java/co/airy/core/sources/facebook/Connector.java
@@ -20,15 +20,14 @@ import org.apache.avro.specific.SpecificRecordBase;
 import org.apache.kafka.streams.KeyValue;
 import org.slf4j.Logger;
 import org.springframework.beans.factory.annotation.Value;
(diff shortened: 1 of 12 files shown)


AitorAlgorta (dismissed):

backend/sources/facebook/connector/src/main/java/co/airy/core/sources/facebook/Connector.java
```diff
@@ -169,7 +168,6 @@ public RequestLoggingIgnorePatterns requestLoggingIgnorePatterns() {
     }
 
     @Bean
-    @ConditionalOnProperty("segment.analytics.enabled")
```
pascal-airy, line 172: I think without this conditional there is no way to disable tracking

chrismatix: @pascal-airy [This](https://github.com/airyhq/airy/blob/64c6a59314f0a6496e0804b2c73b743f688893f4/lib/java/tracking/src/main/java/co/airy/tracking/TrackingRequestFilter.java#L18) is the place that ultimately controls whether or not data is sent to Segment. If it's not initialized then the `RouteTracking` beans won't do anything

PR #2537, opened 2021-10-29
Print airy core version with version command

### LINKED ISSUE #2530: Amend the airy version command to print the version of Airy core
When invoked inside a workspace, the `airy version` sub-command should print the version of the `Airy Core instance`. 

```
$ airy version
CLI version: 0.33.0, GitCommit: 9e9ffa393f7b1d4e92464dba2f97d4ae0e5fef1d
Airy instance version: 0.32.0 
```

If it is invoked outside of a workspace, then it prints only the CLI version, like it is now.

If it is invoked inside a workspace, but it cannot retrieve the version from the Kubernetes cluster (`core-config` configmap), then it should return a warning

```
$ airy version
CLI version: 0.33.0, GitCommit: 9e9ffa393f7b1d4e92464dba2f97d4ae0e5fef1d
Warning: Unable to retrieve the version of the Airy Core instance.
```

lucile-airy:
Resolves #2530 exactly as described.

### DIFF
--- a/cli/pkg/cmd/root.go
+++ b/cli/pkg/cmd/root.go
@@ -7,6 +7,7 @@ import (
 	"cli/pkg/cmd/status"
 	"cli/pkg/cmd/ui"
 	"cli/pkg/cmd/upgrade"
+	"cli/pkg/kube"
 	"cli/pkg/workspace"
 	"fmt"
 	"os"
@@ -29,7 +30,7 @@ var RootCmd = &cobra.Command{
 	TraverseChildren: true,
 	PersistentPreRun: func(cmd *cobra.Command, args []string) {
 		if cmd.Name() != "create" && cmd.Name() != "version" {
-			workspace.Init(cliConfigDir)
+			workspace.Init(cliConfigDir, true)
 		}
 	},
 }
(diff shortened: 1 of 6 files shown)


cli/pkg/workspace/init.go
```diff
@@ -3,22 +3,25 @@ package workspace
...
 	"fmt"
-	"github.com/spf13/viper"
 	"os"
 	"path/filepath"
+
+	"github.com/spf13/viper"
 )
 
-func Init(path string) ConfigDir {
+func Init(path string, log bool) ConfigDir {
```
chrismatix, line 12: It's a bit of a bad practice to pass boolean flags to methods, especially if the flag only has one use case. Here: "Don't log a failure for the `airy version` command".

The idiomatic way would be to return an error `func Init(path string, log bool) (ConfigDir, error)`. Then the function users can decide what to do with that error (log or ignore)

cli/pkg/cmd/root.go
```diff
@@ -39,7 +40,23 @@ var versionCmd = &cobra.Command{
...
 	Run: func(cmd *cobra.Command, args []string) {
-		fmt.Printf("Version: %s, GitCommit: %s\n", Version, CommitSHA1)
+		fmt.Printf("CLI version: %s, GitCommit: %s\n", Version, CommitSHA1)
+
+		wsPath, _ := cmd.Flags().GetString("workspace")
+		dir := workspace.Init(wsPath, false) // Will exit if command is invoked outside of a workspace
+		dir.LoadAiryYaml()
+		kubeCtx := kube.Load()
+		set, err := kubeCtx.GetClientSet()
+		if err != nil {
```
chrismatix, line 50: We should log this error.

cli/pkg/cmd/root.go
```diff
@@ -39,7 +40,23 @@ var versionCmd = &cobra.Command{
...
+			return
+		}
+
+		coreConfig, err := kube.GetCmData("core-config", viper.GetString("namespace"), set)
+		if err != nil {
+			fmt.Println("Warning: Unable to retrieve the version of the Airy Core instance.")
```
chrismatix, line 56: Can't hurt to include the original error in this message, so that we can debug this in case a user has issues :)

cli/pkg/workspace/init.go
lucile-airy: The update now works with the error return. Let me know if this is the correct way!

chrismatix (approved):
Good work 🙌

PR #2538, opened 2021-11-01
[2457] Fix upgrade to same version

ljupcovangelski:
Making sure that the controller restarts on upgrade, so that it starts all the components, for which there are already config maps.

### DIFF
--- a/infrastructure/helm-chart/templates/controller/deployment.yaml
+++ b/infrastructure/helm-chart/templates/controller/deployment.yaml
@@ -18,6 +18,8 @@ spec:
     metadata:
       labels:
         app: airy-controller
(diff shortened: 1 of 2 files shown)


infrastructure/helm-chart/templates/controller/deployment.yaml
```diff
@@ -18,6 +18,8 @@ spec:
     metadata:
       labels:
         app: airy-controller
+      annotations:
+        releaseTime: {{ (now) | toString | trunc 28 | quote }}
```
pascal-airy, line 22: that's nice!

pascal-airy (approved):

PR #2609, opened 2021-11-22
[#2604] Added attachment toggle for chatplugin

### LINKED ISSUE #2604: Chat Plugin Configs: Attachments
As we start to send/render attachments via the different channels, we should offer Airy users the ability to configure their attachment preferences for the Live Chat Plugin alongside the rest of the Chat Plugin configurations.

We should offer Airy users checkboxes for the following:

- [ ] File
- [ ] Image
- [ ] Video

These checkboxes can be shown at the bottom of the configurations list. Users can decide which types of attachments they want to receive via their Chat Plugin, and this decision should be reflected in the code on the lefthand side.

Screenshot 2021-11-18 at 17 10 46

thorstenairy:
closes #2604


AudreyKj (approved):
LGTM

PR #2628, opened 2021-11-30
Add contacts service

### LINKED ISSUE #2293: Introduce Contacts as a separate entity
**Is your feature request related to a problem? Please describe.**

So far Airy handles contact profile information as metadata of conversations due to path dependency from sources such as `facebook` or `google` where conversations can only be initiated by the contact. 

The reasons to include the concept of `Contacts` as a separate entity from `Conversations` are: 

- A contact could be imported or created before the first message is sent (e.g. imported from lists or CRM systems), so that for sources that don't require to initiate the conversation by the contact (e.g. SMS or WhatsApp), the first message in a conversation could be sent from the Airy instance to the contact   
- There could be several conversations that all relate to the same physical person i.e. contact (e.g. the same person on WhatsApp, Facebook, Google or talking to different Facebook pages)
- It would generally be the better fitting place to store and modify profile information (names, avatar images, etc.) instead of the metadata store of conversations syncing information in the way they are stored in existing CRM systems (Salesforce, etc.)

**Describe the solution you'd like**

1. Introduce Contact Objects as a separate Topic:

```
{
  "id": "UUID",
  "first_name": "Mike",
  "last_name": "Smith",
  "display_name": "Mike Smith",
  "avatar_url": "SOMEURL",
  "title": "Chief Developer Evangelist",
  "gender": "somevalue",
  "timezone": 2.00, // relative to GMT
  "locale": "en_US",
  "organization_name": "Airy GmbH", 
  "contact_details": {
    "phone": "+49123456789",
    "email": "liz@airy.co",
    "key": "value" // would be freely extendible
  },
  "address": {
    "organization_name": "Airy GmbH"
    "address_line1": "Some Street"
    "addess_line2": "",
    "postcode": "12346",
    "city": "Berlin",
    "state": null,
    "country": "DE"
  },
  "conversations": [
    {
      "conversation_id": "UUID",
      "source":  "chat_plugin"
    },
    {
      "conversation_id": "UUID",
      "source":  "facebook"
    }  
  ],
  "metadata": {
    "facebook": { 
      "psid": "1234567890",
      "page_id": "9876543210"
    },
    "salesforce_id": "ABC123",
    "zendesk_id": "XYZ789",
    "key": "value" // would be freely extendible
  },
  "created_at": "TIMESTAMP",
  "updated_at": "TIMESTAMP"
}
```

2. Introduce HTTP API Endpoints:

- POST `/contacts.create`: Creates a single contact object

- POST `/contacts.import`: Imports an Array of contact objects, e.g. names and phone numbers

- POST `/contacts.list`: Array of all contacts, should be queryable

- POST `/contacts.info`: Return contact information by contact id

- POST `/contacts.update`: Updates the contact object with additional information; empty values having `{ "key": ""}` should clear/remove the relevant property

- POST `/contacts.refetch`: source-specific, for a list of contact ids, refetch the contact information from the original source, e.g. Facebook  

- POST `/contacts.merge`: target contact_id and array of contact ids to be "merged" into target contact_id, by adding the relevant conversation in the conversations property to target and hiding the formerly separate contact_ids; when a formerly separate contact_id is queried, e.g. from a conversation, the information   

- POST `/contacts.delete`: contacts without a conversation should be deletable, otherwise only if the relevant conversation was archived or "removed" first

3. Consume Events for Contacts topic

Depending on the relevant source of the events the Contacts topic should be able to consume events from external systems, such as `contact.created`, `contact.updated`, `customer_created`, `customer_updated` to the extent the relevant events and their properties have been properly mapped to the contacts topic.

**Describe alternatives you've considered**
-

**Additional context**
-

chrismatix:
required for #2293 

Features:
- Create contact manually
- Create contact automatically from conversations
- Update contact information from conversation metadata
- Update contacts
- Get and list contacts


backend/api/contacts/src/main/java/co/airy/core/contacts/dto/Contact.java
```diff
@@ -0,0 +1,225 @@
...
+            public static String POSTAL_CODE = "address.postal_code";
+            public static String CITY = "address.city";
+            public static String STATE = "address.state";
+            public static String COUNTRY = "address.country";
+        }
+    }
+
+
+    @JsonIgnore
+    public List<Metadata> toMetadata() {
```
chrismatix, line 140: The contact object is stored entirely in metadata. Therefore we need a full mapping from one representation to the other.

lib/java/kafka/schema/src/main/java/co/airy/kafka/schema/application/ApplicationCommunicationContacts.java
```diff
@@ -0,0 +1,17 @@
+package co.airy.kafka.schema.application;
+
+import co.airy.kafka.schema.ApplicationCommunication;
+
+import java.util.Map;
+
+public class ApplicationCommunicationContacts extends ApplicationCommunication {
```
chrismatix, line 7: In my opinion, it makes sense to store all contact-related metadata in a separate topic. Even though the data structure is the same the data has a different meaning and we should be able to separate it on a storage level.

backend/api/contacts/src/main/java/co/airy/core/contacts/ContactsController.java
```diff
@@ -0,0 +1,176 @@
...
+                .conversations(Optional.ofNullable(payload.getConversations()).orElse(contact.getConversations()))
+                .displayName(Optional.ofNullable(payload.getDisplayName()).orElse(contact.getDisplayName()))
+                .avatarUrl(Optional.ofNullable(payload.getAvatarUrl()).orElse(contact.getAvatarUrl()))
+                .gender(Optional.ofNullable(payload.getGender()).orElse(contact.getGender()))
+                .locale(Optional.ofNullable(payload.getLocale()).orElse(contact.getLocale()))
+                .organizationName(Optional.ofNullable(payload.getOrganizationName()).orElse(contact.getOrganizationName()))
+                .timezone(Optional.ofNullable(payload.getTimezone()).orElse(contact.getTimezone()))
+                .title(Optional.ofNullable(payload.getTitle()).orElse(contact.getTitle()))
+                .via(Optional.ofNullable(payload.getVia()).orElse(contact.getVia()))
+                .build();
```
juan-sebastian, line 141: It is a bit difficult to read what is going on here. maybe it can improve readability by moving the map part to a function like so
```java
final addressBuilder = (address, payloadAddress) -> {...}
```
And/Or have a helper function for: `Optional.ofNullable(payload.getConversations()).orElse(contact.getConversations())` 
So we cold do

```java
.conversations(get(payload.getConversations(), contact.getConversations()))
```

I don't know if this is even possible in java. 🙈

juan-sebastian (commented):
Nice work 💪. I left 2 small comments

backend/api/contacts/src/main/java/co/airy/core/contacts/Stores.java
```diff
@@ -0,0 +1,198 @@
...
+        }
+    }
+
+    @Override
+    public Health health() {
+        getContactStore();
+        getConversationsStore();
+        getConversationToContactStore();
+        return Health.up().build();
+    }
```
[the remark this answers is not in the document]
chrismatix: @juan-sebastian It's true that it could be more explicit. We do it like that everywhere though and the "advantage" of this approach is that Spring automatically includes the thrown error in the status response and there's a log of course.

juan-sebastian (approved):

backend/api/contacts/src/main/java/co/airy/core/contacts/ContactsController.java
chrismatix: I cleaned it up, by moving the address related merging code into the address class: https://github.com/airyhq/airy/pull/2628/files#diff-8aba54f3e719ce007647fefe1fdaddcf2df144d8fbe338a2a25e92ceab82a611R122

juan-sebastian (approved):

PR #2630, opened 2021-12-01
[#2611] Session timeout 

### LINKED ISSUE #2611: Chat Plugin: session times out
Currently, if you interact with the chat plugin and leave it open for 1+ hours, you get a "Reconnecting..." message that does not go away.

Realistically, the session shouldn't "time out" and users should be able to go back to the chat plugin and write after 1+ hours.

Able to duplicate the issue with other chat plugins as well.

AitorAlgorta:
closes #2611


frontend/chat-plugin/lib/src/websocket/index.ts
```diff
@@ -70,7 +70,14 @@ class WebSocket {
     this.client.onStompError = (frame: IFrame) => {
       console.error('Broker reported error: ' + frame.headers['message']);
       console.error('Additional details: ' + frame.body);
-      authenticate(this.channelId);
+      console.error('errorrr: ' + frame.headers.keys);
```
thorstenairy, line 73: remove console.error

thorstenairy (approved):
LGTM

lib/typescript/assets/scss/fonts.scss
```diff
@@ -2,56 +2,56 @@
...
-  line-height: 2rem;
+  font-size: 20px;
+  line-height: 32px;
 }
 
 @mixin font-l {
   font-family: 'Lato', sans-serif;
-  font-size: 1.563rem; // 25px
-  line-height: 2rem;
+  font-size: 25px;
```
chrismatix, line 23: @AitorAlgorta what's the reason for switching this to pixel's? If there is a good reason we should document it, because as far as I can understand `rem` scales while pixels are fixed size.

AitorAlgorta: That's why. `rem` not being a fix size brought some problems in different customer webpages. We want the chat plugin to look always the same no matter what and with `rem` it was looking really weird. Messages had una size. the header bar hada different one, the input a different one. I changed it to `px` so it always looks the same. @chrismatix

AitorAlgorta: Screenshot 2021-12-08 at 10 22 46

AitorAlgorta: Look at this ridiculous difference in sizes. We cannot control how `rem` behaves in different customer webpages, with `px` we have more control over it.

chrismatix: Makes sense. Was not thinking about it in the context of the chatplugin 👍

PR #2638, opened 2021-12-07
[#1246] Add documentation on how to run images locally

### LINKED ISSUE #1246: Use Minikube registry to test images locally
**Current State**

A common issue we face when developing features is testing images in integration with the core installation. In #1041 we introduce `minikube` as a new local provider. 

**Idea** We could use the local image registry minikube provides to test and reload changes without the need of pushing image tags to GitHub.

In the context of this issue, we also want to document the dev workflow. Idea: add a new document in the `concepts` section of the docs and link it back from the contributing.md. The doc should mention:

- how to run airy create for minikube
- set up the local registry
- create a cluster from a branch
- run all the cypress tests as well

lucile-airy:
Resolves #1246.

Notes:
- It is probably a good idea to add a new bazel rule that will publish local images with the tag `:latest` (instead of `:image`), otherwise minikube will not refresh the running image when you overwrite it again.
- @ljupcovangelski said: We should add a change to the helm charts, because this patch can be overwritten if somebody does `airy upgrade`. (Perhaps also add a flag to `airy upgrade` in case this is desirable behaviour?)

### DIFF
--- a/docs/docs/guides/contributing.md
+++ b/docs/docs/guides/contributing.md
@@ -111,6 +111,52 @@ Bazel also offers a friendly and powerful autocompletion, please refer to [this
 document](https://github.com/bazelbuild/bazel/blob/master/site/docs/completion.md)
 to install it locally.
 
+### Publishing and running images on a local instance
+
+You can test your images at runtime if you have a local instance of Airy Core
+running on Minikube.
+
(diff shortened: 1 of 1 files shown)


pascal-airy (approved):
awesome stuff!

PR #2657, opened 2021-12-14
[#2623] Return message delivery error as metadata

### LINKED ISSUE #2623: Return delivery error metadata

In case delivery for a Facebook message fails, I want the reason of failure to be returned async via metadata webhook/websocket.

When sending a message to Facebook using `/messages.send` fails I want to receive the error in the metadata.

chrismatix:
resolves #2623


backend/sources/facebook/connector/src/main/java/co/airy/core/sources/facebook/api/Api.java
```diff
@@ -51,16 +57,15 @@
     private final HttpHeaders httpHeaders = new HttpHeaders();
     private final String appId;
     private final String apiSecret;
-    private static final String errorMessageTemplate =
-            "Exception while sending a message to Facebook: \n" +
-                    "Http Status Code: %s \n" +
-                    "Error Message: %s \n";
 
-    public Api(ObjectMapper objectMapper, RestTemplateBuilder restTemplateBuilder,
```
chrismatix, line 59: Here we use inject the default airy object mapper, but it's better to explicitly define one for Facebook here.

backend/sources/viber/connector/src/main/java/co/airy/core/sources/viber/services/Api.java
```diff
@@ -81,4 +92,33 @@ public void setWebhook(String webhookUrl) throws Exception {
     public void removeWebhook() throws Exception {
         getApiResponse("/set_webhook", objectMapper.writeValueAsString(Map.of("url", "")));
     }
+
+
+    @Override
+    public void onApplicationEvent(ApplicationReadyEvent event) {
+        restTemplate = restTemplateBuilder
+                .errorHandler(new ResponseErrorHandler() {
```
juan-sebastian, line 100: Interesting anonymous classes 🤔

backend/sources/facebook/connector/src/main/java/co/airy/core/sources/facebook/Connector.java
```diff
@@ -78,13 +78,22 @@
 
             return List.of(KeyValue.pair(message.getId(), message), KeyValue.pair(getId(metadata).toString(), metadata));
         } catch (ApiException e) {
-            log.error(String.format("Failed to send a message to Facebook \n SendMessageRequest: %s \n Error Message: %s \n", sendMessageRequest, e.getMessage()), e);
```
juan-sebastian, line 81: Even if we are returning the error now. It does not hurt to log it anyway

juan-sebastian (dismissed):
Looks good to me.  I would say that it will be great to have a bit more logs. not only when there is an error/warning but also on the happy path info/debug

backend/sources/viber/connector/src/main/java/co/airy/core/sources/viber/services/Api.java
chrismatix: Yeah. By the way, there is yet another library which we used in the past and that could allow us to write declarative, simpler and thus more maintainable REST clients: https://github.com/OpenFeign/feign 

The best case though is that we rely on the new [source api](https://airy.co/docs/core/api/source) and write those source clients in a different language (go perhaps? ;) )

backend/sources/facebook/connector/src/main/java/co/airy/core/sources/facebook/Connector.java
chrismatix: Agreed, I will add them back in

backend/sources/viber/connector/src/main/java/co/airy/core/sources/viber/services/Api.java
juan-sebastian: > Yeah. By the way, there is yet another library which we used in the past and that could allow us to write declarative, simpler and thus more maintainable REST clients: https://github.com/OpenFeign/feign

I will have a look thanks

> The best case though is that we rely on the new source api and write those source clients in a different language (go perhaps? ;) )

That will be a great Idea. We can discuss that further

juan-sebastian (approved):
Looks good to me. Great work 💪

PR #2658, opened 2021-12-15
[#2622] api expose facebook page tokens via channels api

### LINKED ISSUE #2622: API: Expose Facebook Page Tokens via Channels API

As an Airy developer,
I want to see channel specific credentials (such as Facebook Page Tokens for sources of type `facebook` and `instagram`), 
so that I can monitor the health of the token and send direct API calls, e.g. to edit the messenger profile.

At the same time I would wish for more consistency regarding API calls and responses, e.g. the Channel Connect Request for source type `facebook` returns the `page_id` as `source_channel_id` while the relevant request for source type `instagram` returns the `account_id` as `source_channel_id` and doesn't return the `page_id` at all while the parameter is still required for `POST /channels.instagram.connect` (https://airy.co/docs/core/api/endpoints/channels#instagram).

Screenshot 2021-11-29 at 11 33 30

Screenshot 2021-11-29 at 11 43 13

To an authenticated request to `POST /channels.list` return channel-specific credentials and allow for consistency at least in metadata with the corresponding connect-request for the relevant source type. Repeat key-value-pairs if the naming is not consistent, so the developer doesn't have to guess what means what for which source.

```
 {
      "id": "daf6df04-7883-5309-9a33-1fc880ed1567",
      "source": "instagram",
      "source_channel_id": "17841400425075798",
      "connected": true,
      "metadata": {
        "name": "SOMENAME",
        "page_id": "FB-PAGE-ID",
        "page_token": "FB-PAGE-TOKEN",
        "account_id": "17841400425075798"
      }
    }
```

Feel free to rename properties as it makes sense.

We could provide an API Wrapper that sends authenticated requests to the source API, e.g. to access Messenger Profile, etc. 

## Additional context

-

juan-sebastian:
closes #2622


juan-sebastian:
For now I just added the integration tests.

backend/sources/facebook/connector/src/test/java/co/airy/core/sources/facebook/ChannelsControllerTest.java
```diff
@@ -0,0 +1,230 @@
...
+
+        final List<Metadata> metadataList = kafkaTestHelper.consumeValues(1, applicationCommunicationMetadata.name());
+        final Metadata metadata = metadataList.get(0);
+
+        assertThat(MetadataKeys.ChannelKeys.NAME, equalTo(metadata.getKey()));
+        assertThat(connectPayload.getName(), equalTo(metadata.getValue()));
+
+
+        // Disconnect from instagram channel
+        final DisconnectChannelRequestPayload disconnectRequestPayload = new DisconnectChannelRequestPayload(
```
chrismatix, line 194: One more thing I forgot to mention about our integration tests: We generally avoid using the request payloads from that app packages. Because if we want to test the real thing then the payload definition may also contain an error :) 

So there are two way's we usually construct these payloads:
1. Build an inline string, i.e. `String.format("{\"channel_id\":\"%s\"}", channelId)`
2. [Using jackson](https://github.com/airyhq/airy/blob/280bf7ec36094a0c1935ac5a8a0d3d8ab3bf2087/backend/sources/viber/connector/src/main/java/co/airy/core/sources/viber/services/Api.java#L69)

chrismatix (commented):
Looks good sofar. Small comment :)

chrismatix (dismissed):
Great work 🙌

juan-sebastian:
Add page-id page-token and account-id to metadata for FB and instagram sources and improve tests

backend/sources/facebook/connector/src/test/java/co/airy/core/sources/facebook/ChannelsControllerTest.java
```diff
@@ -0,0 +1,255 @@
...
+
+        assertThat("facebook", equalTo(channel.getSource()));
+        assertThat(pageWithConnectInfo.getId(), equalTo(channel.getSourceChannelId()));
+        assertThat(ChannelConnectionState.CONNECTED, equalTo(channel.getConnectionState()));
+
+        final List<Metadata> metadataList = kafkaTestHelper.consumeValues(4, applicationCommunicationMetadata.name());
+        metadataList.stream().forEach((metadata) -> {
+            final String key = metadata.getKey();
+            switch (key) {
+                case MetadataKeys.ChannelKeys.NAME:
```
chrismatix, line 134: I like this way of testing metadata lists! :)

backend/sources/facebook/connector/src/main/java/co/airy/core/sources/facebook/api/model/MetadataKeys.java
```diff
@@ -0,0 +1,10 @@
+package co.airy.core.sources.facebook.api.model;
+
+
+public class MetadataKeys extends co.airy.model.metadata.MetadataKeys {
```
chrismatix, line 4: No need to do subclassing here since these are basically constants.

chrismatix (approved):

PR #2662, opened 2021-12-16
[#2626] Chat Plugin full screen for mobile

### LINKED ISSUE #2626: Chat plugin mobile version view
For the chat plugin mobile version, the site visitor should be able to scroll normally on the website and the chat plugin bubble should be visible when it's closed.

If a visitor clicks to open the chat plugin, the chat will become full screen and the bubble will disappear -- so that they can only minimize the chat plugin with the X option on the righthand side of the header.

If they minimize the chat with the X option, the experience will return to the original -- the bubble will be visible on the page and the visitor will be able to scroll.

AitorAlgorta:
closes #2626


AudreyKj (approved):
LGTM

PR #2664, opened 2021-12-17
[#2663] Pull busybox from Github packages

### LINKED ISSUE #2663: Busybox image too many requests error from docker.com
We use `busybox` in our `initContainers` to check if Kafka/Zookeeper are running before starting the apps.

When restarting all the pods at once, sometimes we endup with a throttling error from DockerHub

```
429 Too Many Requests - Server message: toomanyrequests: You have reached your pull rate limit.
```
We should move this image to Github Packages or find another solution.

ljupcovangelski:
- Push our own busybox image.
- Set it as default in the helm charts.
- Make it optional so that people can overwrite with `--set global.busyboxImage`.

Fixes #2663.


pascal-airy (approved):
lgtm

PR #2671, opened 2021-12-20
[#2670] Prevent messages spam

### LINKED ISSUE #2670: Prevent messages from sending more than once from Inbox
Related to #2358

Currently, it is possible for a system user to accidentally send the same message more than once if they click the `Send` button more than once or if they press `Enter` more than once.

screenshot_2021-12-10_at_15 28 27

The image above is an example (responding to a message from the chat plugin) of what happens if a system user accidentally presses `Enter` more than once.

Ideally, we should set some kind of timeout that disables users from sending the same message within a couple of seconds.

One idea is to use a fix like [this](https://stackoverflow.com/questions/38456244/prevent-multiple-time-form-submit-with-enter-key/38456995#38456995) where we could set a timeout of 1s for hitting enter.

### LINKED ISSUE #2673: UI crashes onClick Install & Customize Chatplugin
## Describe the bug

The UI crashes when you a User tries to Install & Customize any Chatplugin Channel

## To Reproduce

1. I installed Airy using UI
2. Connect a new Chatplugin channel
3. Edit this Chatplugin channel to Install & Customize it
4. Click on Install & Customize
5. See Error

## Expected behavior

Click on Install & Customize and see your custom Chatplugin

thorstenairy:
closes #2670 
closes #2673

### DIFF
--- a/frontend/chat-plugin/lib/src/airyRenderProps/AiryBubble/index.tsx
+++ b/frontend/chat-plugin/lib/src/airyRenderProps/AiryBubble/index.tsx
@@ -40,7 +40,7 @@ const AiryBubble = (props: Props) => {
   return (
     <div
       className={style.hideBubble}
-      style={props.isChatHidden ? {height: '60px', display: 'block'} : {}}
+      style={props.isChatHidden && {height: '60px', display: 'block'}}
       onClick={() => props.toggleHideChat()}
       data-cy={props.dataCyId}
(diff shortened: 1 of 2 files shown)


AudreyKj (approved):

thorstenairy:
was fixed in the release/0.37.0 to be in the release as it was a blocker

PR #2672, opened 2021-12-20
[#1497] media resolver crashes on bad configuration

### LINKED ISSUE #1497: Media Resolver crashes on bad configuration
### Current Behaviour
On startup, the media resolver app tries to build an AWS client based on the config values and that fails if they are invalid which results in the app crashing

### Expected Behaviour
The app should log that the configuration is invalid and not crash

### Example configuration which produced the error
```yaml
media-resolver:
    storage:
      s3:
        key: "changeme"
        secret: "changeme"
        bucket: "changeme"
        region: "changeme"
        path: "path"
```

juan-sebastian:
closes #1497 

Test if the connection is active and health on startup. if not sets the microservice as unhealthy (down)

### DIFF
--- a/backend/media/src/main/java/co/airy/core/media/services/MediaUpload.java
+++ b/backend/media/src/main/java/co/airy/core/media/services/MediaUpload.java
@@ -43,7 +43,7 @@ public class MediaUpload implements HealthIndicator {
         this.path = appendSlash(path);
         this.host = new URL(String.format("https://%s.s3.amazonaws.com/", bucket));
 
-        // Check if the connection is healthy
+        // Check if the connection is healthy. This is only done during the construction face
         checkConnectionStatus();
     }
 
@@ -84,10 +84,10 @@ public class MediaUpload implements HealthIndicator {
                 connectionStatus = true;
             }
         } catch (SdkClientException e) {
-            log.error("AWS SDK extension", e);
+            log.error("AWS SDK exception", e);
 
         } catch (Exception e) {
-            log.error("unexpected extension", e);
+            log.error("unexpected exception", e);
         }
     }
 
--- a/backend/media/src/main/java/co/airy/core/media/services/MediaUpload.java
+++ b/backend/media/src/main/java/co/airy/core/media/services/MediaUpload.java
@@ -10,6 +10,8 @@ import com.amazonaws.services.s3.model.PutObjectRequest;
 import lombok.Getter;
 import org.slf4j.Logger;
 import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.actuate.health.Health;
+import org.springframework.boot.actuate.health.HealthIndicator;
 import org.springframework.retry.annotation.EnableRetry;
 import org.springframework.retry.annotation.Retryable;
 import org.springframework.stereotype.Service;
@@ -23,7 +25,7 @@ import java.util.List;
 
 @Service
 @EnableRetry
-public class MediaUpload {
+public class MediaUpload implements HealthIndicator {
     private static final Logger log = AiryLoggerFactory.getLogger(MediaUpload.class);
     private final AmazonS3 amazonS3Client;
     private final String bucket;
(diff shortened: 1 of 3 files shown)


backend/media/src/main/java/co/airy/core/media/services/MediaUpload.java
```diff
@@ -14,22 +19,30 @@
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLConnection;
+import java.util.List;
 
 @Service
 @EnableRetry
 public class MediaUpload {
```
chrismatix, line 26: So the cool thing about healthchecks in Spring is that _any_ component can signal its healthy state. For instance `MediaUpload` can also implement the `HealthIndicator` interface. This way we don't need to query the `connectionStatus` variable from the `MediaController`

Sorry, I forgot to mention this. The rest of the code looks great 🙏

chrismatix (commented):
Good implementation, minor nitpick about Spring 🙏

backend/media/src/main/java/co/airy/core/media/services/MediaUpload.java
juan-sebastian: Ok no worries, I moved the Heath check from `MediaController` to `MediaUpload`. Because there is no logic on the `MediaController` Heath implementation. I don't think it is necessary to have one on there

backend/media/src/main/java/co/airy/core/media/services/MediaUpload.java
```diff
@@ -14,22 +21,30 @@
...
+
     public MediaUpload(AmazonS3 amazonS3Client,
                        @Value("${s3.bucket}") String bucket,
                        @Value("${s3.path}") String path) throws MalformedURLException {
         this.amazonS3Client = amazonS3Client;
         this.bucket = bucket;
         this.path = appendSlash(path);
         this.host = new URL(String.format("https://%s.s3.amazonaws.com/", bucket));
+
+        // Check if the connection is healthy
```
pascal-airy, line 46: I'd try to limit the use of comments to when the logic is not deducible from the code

backend/media/src/main/java/co/airy/core/media/services/MediaUpload.java
```diff
@@ -59,4 +74,29 @@ private String resolveContentType(final String fileName, final InputStream is) t
...
+            List<Bucket> buckets = amazonS3Client.listBuckets();
+            if (buckets != null) {
+                log.info("check connection to S3 successful");
+                connectionStatus = true;
+            }
+        } catch (SdkClientException e) {
+            log.error("AWS SDK extension", e);
+
+        } catch (Exception e) {
+            log.error("unexpected extension", e);
```
pascal-airy, line 90: `extension` is that a typo?

pascal-airy (commented):
Thanks for picking this one up!

backend/media/src/main/java/co/airy/core/media/services/MediaUpload.java
juan-sebastian: I think even if it is a bit redundant. That make it obvious and easier to the eye to see. In this case it easy to know that the check it is only done during the construction face.

backend/media/src/main/java/co/airy/core/media/services/MediaUpload.java
juan-sebastian: I wanted to write `exception` 🙈. I am bit dyslexic it won't be the last time this will happen 😅

backend/media/src/main/java/co/airy/core/media/services/MediaUpload.java
pascal-airy: Alright, I don't have a strong opinion so yours that you did the feature matter more :)

pascal-airy (approved):
lgtm!

Issue #2675, opened 2021-12-22
NPM package release rule needs username password


pascal-airy:
As we had problems with 2FA we should change this rule to use an automation token instead of username password

PR #2679, opened 2021-12-23
[#2675] NPM package release rule needs username…

pascal-airy:
…password

https://github.com/airyhq/airy/issues/2675

it should be a token that was created as a:

Screen Shot 2021-12-22 at 09 53 47


juan-sebastian (approved):
Looks good to me 👍

chrismatix (approved):
Finally they added tokens 🙏

Issue #2353, opened 2021-08-30
Change API urls to hyphenation


chrismatix:

steffh:
Some thoughts:

- We should put 301 redirects in place for the existing routes to not break existing clients / implementations.

- Some routes we could simplify instead of hyphenate, e.g. `POST /media.uploadFile` -> `POST /media.upload`

PR #2680, opened 2021-12-23
Add option to add notes to a conversation

### LINKED ISSUE #1998: Optional: Notes About Conversations in UI
**Currently:** the only contact information available in the Inbox is the the name of the contact and any tags related to the conversation.

Screenshot 2021-06-15 at 13 39 37

**Ideally:** we should offer Airy system users to write and view notes below the Tags section within the contact information.

**Requirements:**

- If the conversation is displayed and I click on the notes field, I will be able to manually type in and edit notes
- If I am editing the notes field and I hit enter, I will be able to save the written note
- If I have a long note above 500 characters and I want to see the note, I will only be able to see part of the note (maximum 3 lines of text/special characters/numbers visible at a time)
- If I have multiple notes and I want to see an older specific note, I can scroll down to access older notes

**Notes:**

- Allowed input: text, special characters, numbers
- No character limit

lucile-airy:
Resolves #1998


backend/api/communication/src/main/java/co/airy/core/api/communication/ConversationsController.java
```diff
@@ -168,6 +160,83 @@
         return ResponseEntity.noContent().build();
     }
 
+    @PostMapping("/conversations.addNote")
+    ResponseEntity<?> conversationAddNote(@RequestBody @Valid ConversationAddNoteRequestPayload payload) {
+        final String conversationId = payload.getConversationId().toString();
+        final String text = payload.getText();
+        if (text != null && (text.length() == 0 || text.length() >50000)) {
```
chrismatix, line 167: You can also use annotations on the request payload class instead of writing this code. `@NotNull` and `@Size` should do the trick.

backend/api/communication/src/main/java/co/airy/core/api/communication/ConversationsController.java
```diff
@@ -168,6 +160,83 @@
...
+        try {
+            stores.storeMetadata(metadata);
+        } catch (Exception e) {
+            return ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).body(new RequestErrorResponsePayload(e.getMessage()));
+        }
+
+        return ResponseEntity.noContent().build();
+    }
+
+    @PostMapping("/conversations.deleteNote")
```
chrismatix, line 190: Per user request (#2353 ) we now use hyphenation for multi word path parts. So these routes should be `/conversations.add-note` and `/conversations.delete-note`

backend/api/communication/src/main/java/co/airy/core/api/communication/lucene/DocumentMapper.java
```diff
@@ -40,8 +42,11 @@ public Document fromConversationIndex(ConversationIndex conversation) {
 
         for (MetadataNode node : conversation.getMetadata()) {
             final String key = String.format("metadata.%s", node.getKey());
-            // Index but don't store metadata
-            document.add(new TextField(key, node.getValue(), Field.Store.NO));
+            if (node.getKey().startsWith(MetadataKeys.ConversationKeys.NOTES)) {
```
chrismatix, line 45: Why do we need to store notes in Lucene?

backend/api/communication/src/test/java/co/airy/core/api/communication/ConversationsNoteTest.java
```diff
@@ -0,0 +1,240 @@
...
+        String finalNoteKey = noteKey;
+
+        webTestHelper.post("/conversations.info",
+                        "{\"conversation_id\":\"" + conversationId + "\"}")
+                .andExpect(status().isOk())
+                .andExpect(jsonPath("$.id", is(conversationId)))
+                .andExpect(jsonPath(String.format("$.metadata.notes['%s']", finalNoteKey), is(text)));
+
+        webTestHelper.post("/conversations.deleteNote",
+                        "{\"conversation_id\":\"" + UUID.fromString(conversationId) + "\",\"note_id\":\"" + UUID.fromString(finalNoteKey) + "\"}")
```
chrismatix, line 121: What's the use of `UUID.fromString` here if it gets casted back into a string by the concatenation?

frontend/ui/src/pages/Inbox/Messenger/ConversationMetadata/index.tsx
```diff
@@ -81,13 +112,32 @@ const ConversationMetadata = (props: ConnectedProps<typeof connector>) => {
...
-  const conversationTags = () =>
-    Object.keys(conversation.metadata.tags || {})
+  const conversationTags = () => {
+    return Object.keys(conversation.metadata.tags || {})
       .map(tagId => tags[tagId])
       .filter(tag => tag !== undefined)
       .sort(tagSorter);
+  };
+
+  const conversationNotes = () => {
```
chrismatix, line 134: Refactoring idea: Let's move both `conversationTags` and `conversationNotes` to `lib/typescript/model/Conversation.ts` so that we keep data logic close to the types.

chrismatix (commented):
First up: **Very, very** impressive first feature PR, both on the frontend and on the backend. I left some small comments here and there.

Two things that need changes:
1. There is one mismatch with the [feature description](https://github.com/airyhq/airy/issues/1998):
> If I have multiple notes and I want to see an older specific note, I can scroll down to access older notes

To me this implies that we also need to return the time information. So instead of  
```json5
{
  // conversation ...
  "notes": {
    "note-id-1": "Note 1",
    "note-id-2": "Note 2"
  }
}
```
we can use the metadata timestamp to return this:
```json5
{
  // conversation ...
  "notes": {
    "note-id-1": {
      "text": "This is a note",
      "timestamp": "2017-01-01T00:00:00.000Z"
    }
  }
}
```

**How to do this?**
This is of course a bit tricky since it changes the default metadat mapping behavior of the [MetadataObjectMapper](https://github.com/airyhq/airy/blob/c06f95b67de4b3f77a7b04f150823f498c23977b/backend/model/metadata/src/main/java/co/airy/model/metadata/MetadataObjectMapper.java#L23) 

One way of solving this would be to
1. Turn the mapper into a class without static functions so that it can use state
2. Allow for registering mapping plugins that can be added for a specific prefix
3. Register such a prefix for mapping the notes in the conversation controller



2. The docs are missing. Writing the docs first helps us catch issues like (1.) before sitting down and doing and doing all the hard coding work :)

frontend/ui/src/components/Note/index.tsx
```diff
@@ -14,29 +14,33 @@ type NoteProps = {
...
 const Note = ({note, onClick, removeNote, updateNote}: NoteProps): JSX.Element => {
-  return (
-    <div className={styles.note} onClick={onClick}>
-      <div className={`${styles.noteInner} ${onClick ? styles.clickable : ''} ${removeNote ? styles.isRemovable : ''}`}>
-        <div>
-          <span className={`${styles.noteNameExpanded}`}>{note.text}</span>
-          {removeNote && (
-            <span className={styles.removeNote} onClick={removeNote}>
-              <Close className={styles.closeButton} title="Delete" />
+  if (note.timestamp && note.text) {
```
chrismatix, line 17: nitpick: Checking for the inverse and returning early makes for less indentation and better readability.

frontend/ui/src/components/Note/index.tsx
```diff
@@ -14,29 +14,33 @@ type NoteProps = {
...
-        </div>
-        <div>
-          <p className={`${styles.noteDate}`}>
-            {note.timestamp.toLocaleDateString()}{' '}
-            {note.timestamp.toLocaleTimeString([], {hour: '2-digit', minute: '2-digit'})}
-          </p>
+          </div>
+          <div>
+            <p className={`${styles.noteDate}`}>
+              {note.timestamp.toLocaleDateString()}{' '}
```
chrismatix, line 36: We already have a generic [date formatter](https://github.com/airyhq/airy/blob/1ed5f94a556da6dae8b6811a7190df126f4ac548/frontend/ui/src/services/format/date.ts#L3). Let's use that instead :)

frontend/ui/src/pages/Inbox/Messenger/ConversationMetadata/index.tsx
```diff
@@ -216,7 +193,7 @@ const ConversationMetadata = (props: ConnectedProps<typeof connector>) => {
   };
 
   const renderTagsDialog = () => {
-    const filteredTags = getFilteredTags();
+    const filteredTags = getFilteredTags(conversation, tags, tagName);
```
chrismatix, line 196: Good refactoring 👍

lib/typescript/model/Conversation.ts
```diff
@@ -22,3 +27,34 @@ export interface Conversation {
...
 }
+
+export function conversationNotes(conversation: Conversation) {
+  return Object.keys(conversation.metadata.notes || {})
+    .map(noteId => {
+      return {
+        id: noteId,
+        text: conversation.metadata.notes[noteId]['text'],
+        timestamp: conversation.metadata.notes[noteId]['timestamp']
+          ? new Date(parseInt(conversation.metadata.notes[noteId]['timestamp']))
```
chrismatix, line 38: So here the timestamp can either be a string or a `Date`, but in the typescript interface it says `string`. We should decide for one and stick with it 🙏

frontend/ui/src/components/Note/index.module.scss
```diff
@@ -0,0 +1,79 @@
...
+
+.note {
+  display: inline-block;
+  padding: 4px;
+  max-width: 100%;
+  min-width: 100%;
+}
+
+.noteInner {
+  //display: inline-flex;
```
chrismatix, line 11: nitpick: Leftover comment (same below)

lib/typescript/model/Conversation.ts
lucile-airy: But the date comes from the backend as a string to prevent metadata discrepancies, so it has to be defined as such in the metadata interface as well. However, we want to work with the note's date as an object in the functional code, so in the `Note` interface (`lib/typescript/model/Note`) it is defined as a Date. Does that make sense or should we move the date conversion to all places where it's needed? This seems counterproductive given the sort operations. I hope this makes sense 🤞

lib/typescript/model/Conversation.ts
```diff
@@ -2,6 +2,8 @@ import {Contact} from './Contact';
 import {Message} from './Message';
 import {Metadata} from './Metadata';
 import {Channel} from './Channel';
+import {Tag as TagModel, Note as NoteModel} from 'model';
```
chrismatix, line 5: These imports should be:
```
import {Tag as TagModel} from './Tag';
import {Note as NoteModel} from './Tag';
```

lib/typescript/model/Conversation.ts
chrismatix: No, it makes perfect sense. I only wanted to point out that in the ternary you return a `: ''` as the default when the field does not exist. Thereby you break the `NoteModel` interface. So the default should also be some sort of date.

Suggestion regarding the millis as a string coming from the backend: [Here](https://github.com/airyhq/airy/blob/d53c4c0bc13c125a4f84bbd2f8f447f32fa4790e/backend/api/communication/src/main/java/co/airy/core/api/communication/ConversationsController.java#L211) we could store the String already as an ISO string using our [`isoFromMillis`](https://github.com/airyhq/airy/blob/0f3ab3554fbc41b0fe1f44d04fb32c858a1c84b3/lib/java/date/src/main/java/co/airy/date/format/DateFormat.java#L12) function. What do you think?

chrismatix (dismissed):

lib/typescript/model/Conversation.ts
lucile-airy: Ah I see what you mean, I will set it to `null` to prevent rendering.

Concerning the `isoFromMillis`, the MetadataRepository and Metadata builder [do not accept](https://github.com/airyhq/airy/blob/d53c4c0bc13c125a4f84bbd2f8f447f32fa4790e/backend/model/metadata/src/main/java/co/airy/model/metadata/MetadataRepository.java#L14) anything other than a string so this would lead to a lot of refactoring. It could be done but is maybe not necessary?

chrismatix: > Ah I see what you mean, I will set it to null to prevent rendering.
Good solution 👍 
>Concerning the isoFromMillis, the MetadataRepository and Metadata builder do not accept anything other than a string so this would lead to a lot of refactoring. It could be done but is maybe not necessary?
Yes and `isoFromMillis` returns an iso-compliant date string.

lucile-airy: > > Ah I see what you mean, I will set it to null to prevent rendering.
> > Good solution 👍
> > Concerning the isoFromMillis, the MetadataRepository and Metadata builder do not accept anything other than a string so this would lead to a lot of refactoring. It could be done but is maybe not necessary?
> > Yes and `isoFromMillis` returns an iso-compliant date string.

Oh oops I didn't realize. Fixed it now!

frontend/ui/src/pages/Inbox/Messenger/ConversationMetadata/index.tsx
```diff
@@ -261,11 +352,52 @@ const ConversationMetadata = (props: ConnectedProps<typeof connector>) => {
...
+            <div className={styles.notesHeader}>
+              <h3 className={styles.notesHeaderTitle}>Notes</h3>
+              <LinkButton onClick={showAddNotes} type="button" dataCy={cyShowNotesDialog}>
+                {showNotesDialog ? 'Close' : '+ Add Note'}{' '}
+              </LinkButton>
+            </div>
+
+            {showNotesDialog && renderNotesDialog()}
+
+            {conversationNotes(conversation).sort(notesSorter).length > 1 && (
```
chrismatix, line 370: The call to `.sort()` doesn't seem necessary here to check the length

juan-sebastian:
This is not needed anymore

Issue #2682, opened 2021-12-27
Enable AKHQ for multiple namespaces


ljupcovangelski:
Since we are using AKHQ for monitoring Kafka, as we might have multiple `Airy Core` instances with their own brokers in different namespaces, we should update the `helm` chart to be able to deploy a single AKHQ instance, per Kubernetes cluster, to monitor all the Kafka and SchemaRegistry instances in the cluster.

PR #2683, opened 2021-12-27
[#2682] AKHQ for multiple namespaced Kafkas

ljupcovangelski:


infrastructure/helm-chart/charts/tools/charts/akhq/templates/configmap.yaml
```diff
@@ -17,11 +22,13 @@ data:
           properties:
             isolation.level: read_committed
       connections:
-        core:
+      {{- range .Values.kafka.namespaces }}
+        {{ . }}:
           properties:
-            bootstrap.servers: ${KAFKA_BROKERS}
+            bootstrap.servers: {{ $bootstrapServers }}.{{ . }}:{{ $kafkaPort }}
```
ljupcovangelski, line 28: Even if multiple brokers exist in the namespace, this is the URL of the headless service which selects all of the brokers, so it will not be a problem.

infrastructure/helm-chart/charts/tools/charts/akhq/values.yaml
```diff
@@ -1 +1,12 @@
 enabled: false
+workerType:
+hostname:
+namespace: "default"
+kafka:
+  bootstrapServers: "kafka-headless"
```
ljupcovangelski, line 6: The headless service needs to be the same in all namespaces.

infrastructure/helm-chart/charts/tools/charts/akhq/values.yaml
```diff
@@ -1 +1,12 @@
...
+  port: 9092
+  schemaRegistryUrl: "schema-registry"
+  schemaRegistryPort: 8081
+  namespaces:
+    - default
```
ljupcovangelski, line 11: List of namespaces for which to look for `Kafka` and the `SchemaRegistry`. If the `SchemaRegistry` is not namespaced, then the chart needs to be changed.

pascal-airy (approved):
lgtm

PR #2705, opened 2022-01-07
[#2693] contacts import endpoint

### LINKED ISSUE #2693: Contacts service endpoints
Implement the following HTTP API Endpoints:

* POST `/contacts.import`: Imports an Array of contact objects, e.g. names and phone numbers
* ~~POST `/contacts.refetch`: source-specific, for a list of contact ids, refetch the contact information from the original source, e.g. Facebook~~

* POST `/contacts.merge`: target contact_id and array of contact ids to be "merged" into target contact_id, by adding the relevant conversation in the conversations property to target and hiding the formerly separate contact_ids; when a formerly separate contact_id is queried, e.g. from a conversation, the information

* POST `/contacts.delete`: contacts without a conversation should be deletable, otherwise only if the relevant conversation was archived or "removed" first

This is the continuation of the following discussion #2293

juan-sebastian:
Part of #2693

### DIFF
--- a/backend/api/contacts/src/main/java/co/airy/core/contacts/ContactsController.java
+++ b/backend/api/contacts/src/main/java/co/airy/core/contacts/ContactsController.java
@@ -1,5 +1,6 @@
 package co.airy.core.contacts;
 
+import co.airy.avro.communication.Metadata;
 import co.airy.core.contacts.dto.Contact;
 import co.airy.core.contacts.payload.ContactInfoRequestPayload;
 import co.airy.core.contacts.payload.ContactResponsePayload;
@@ -19,11 +20,12 @@ import org.springframework.web.bind.annotation.PostMapping;
 import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RestController;
 
-import javax.validation.Valid;
 import java.time.Instant;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
+import javax.validation.Valid;
 
 import static java.util.stream.Collectors.toList;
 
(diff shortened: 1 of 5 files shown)


backend/api/contacts/src/main/java/co/airy/core/contacts/ContactsController.java
```diff
@@ -63,9 +65,37 @@ public ContactsController(Stores stores) {
...
-        // TODO import an array of contacts
-        return ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).build();
+    public ResponseEntity<?> importContacts(@RequestBody @Valid List<CreateContactPayload> payload) {
+        List<Metadata> contactsMetadata = new ArrayList<Metadata>();
+        List<ContactResponsePayload> createdContacts = new ArrayList<ContactResponsePayload>();
+
+        payload.stream().forEach((p) -> {
+            final Contact newContact = Contact.builder()
+                    .id(UUID.randomUUID().toString())
+                    .createdAt(Instant.now().toEpochMilli())
```
chrismatix, line 75: Wondering if we should allow users to set the creation timestamp? 🤔  Part of me thinks that it's more valuable to know for sure when airy first saw this contact.

backend/api/contacts/src/test/java/co/airy/core/contacts/ImportContactsTest.java
```diff
@@ -0,0 +1,151 @@
...
+                }, "The cache is not sync yet");
+            } catch (Exception e) {
+                assertThat(String.format("exception after retryOnException %s", e.getMessage()), false);
+            }
+        });
+    }
+
+    private List<CreateContactPayload> mockContactsListPayload() {
+        return Arrays.asList(
+                new CreateContactPayload(
```
chrismatix, line 102: This can be more nicely expressed with the `CreateContactPayload.builder()` api.

backend/api/contacts/src/test/java/co/airy/core/contacts/ImportContactsTest.java
```diff
@@ -0,0 +1,151 @@
...
+                objectMapper.writeValueAsString(payload))
+            .andExpect(status().isCreated()).andReturn().getResponse().getContentAsString();
+
+        List<ContactResponsePayload> contactsResp = objectMapper.readValue(
+                content,
+                new TypeReference<List<ContactResponsePayload>>() {});
+        contactsResp.stream().forEach((contactResponse) -> {
+            try {
+                retryOnException(() -> {
+                    final Contact c = stores.getContact(contactResponse.getId());
```
chrismatix, line 86: In the spirit of testing the real thing I think we should test this using either the `/contacts.list` endpoint or by requesting each contact with `/contacts.info`

chrismatix (commented):
Looks pretty good already! 🙌

backend/api/contacts/src/main/java/co/airy/core/contacts/ContactsController.java
juan-sebastian: Not sure I don't have a strong opinion on this. There are some cases it can be useful to set the creation time. But I agree in most cases it will be set automatically

juan-sebastian:
I think it is better to create a PR per end point instead of having a monster PR

backend/api/contacts/src/main/java/co/airy/core/contacts/ContactsController.java
chrismatix: Agreed, then how about: Since we don't allow setting creation time for any other object I'd opt for disallowing it here, too. Until we get a concrete requirement to change this.

chrismatix (approved):

PR #2733, opened 2022-01-17
[#2693] contacts service delete endpoint

juan-sebastian:
Part of #2693


backend/api/contacts/src/main/java/co/airy/core/contacts/dto/Contact.java
```diff
@@ -148,6 +148,32 @@ public Address merge(Address address) {
         }
     }
 
+    public List<Metadata> deleteAllMetadata() {
```
chrismatix, line 151: Good that this is a separate method. One small suggestion for simplification: Call `toMetadata()` and then `.stream().map()` over all records to set their value to `""`. This way we don't have to maintain all keys in this method 🙏

chrismatix (commented):
Looks great. Small nitpick 👍

chrismatix (approved):

PR #2746, opened 2022-01-18
[#2647] Add sender information to messages

### LINKED ISSUE #2647: Extend Message Object by internal sender information

As an Airy developer 
I want to be able to see which user of my Airy instance sent an outbound message via `POST /messages.send`,
so that I can provide more context to multiple support agents,
and so that I can fulfill compliance obligations being able to log who sent a message on behalf of an organization. 

Extending the Message Object in `POST /messages.list` as well as the `last_message` property in `Conversation` as follows:

```JSON5
{
  "id": "{UUID}",
  "content": "{\"text\":\"Hello\"}",
  "state": "pending|failed|delivered",
  "from_contact": false,
  // See glossary
  "sent_at": "{string}",
  //'yyyy-MM-dd'T'HH:mm:ss.SSSZ' date in UTC form, to be localized by clients
  "source": "{String}",
  // one of the possible sources
  "metadata": {
    "sentFrom": "iPhone"
  }
  // metadata object of the message
  "sender": {
      "user_profile": {  
      "nickname": "SOME",
      "name": "SOME NAME",
      "picture": "https://s.gravatar.com/avatar/SOMEID",
      "updated_at": "2021-12-08T16:48:35.891Z",
      "email": "some@domain.com",
      "email_verified": true,
      "sub": "auth0|61a4992ec0baff006ab8af8d",
      "org_id": "org_YTgbi5r9iIqz3huR"
     }
   }
   // sender property containing the oidc user_profile
}
```
It feels most natural to have the sender information as a new primary field on the Message Object and either fill it with contact information or user_profile information. I would prefer this versus nesting it very deeply in the metadata which would be an alternative, unless there are reasonable technical reasons for this.

We could also introduce a separate route to fetch all users that authenticated vis-à-vis an Airy instance (https://github.com/airyhq/airy/issues/2646), so the clients would be expected to query that route and keep it in memory.

Using the oidc security setting from the `airy.yaml` file, we could determine the primary user profile field from the oidc payload and only expose this as the sender of a message. To get all other fields from the user_profile one could look up the corresponding user_profile value form the `POST /user.list` map. 

## Additional context

This backend ticket is a requirement for the corresponding frontend ticket: https://github.com/airyhq/airy/issues/2641

chrismatix:
resolves #2647


lib/java/spring/web/src/main/java/co/airy/spring/web/events/KafkaListener.java
```diff
@@ -5,13 +5,15 @@
...
 import org.apache.kafka.clients.producer.ProducerRecord;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression;
 import org.springframework.context.ApplicationListener;
 import org.springframework.stereotype.Component;
 
 import java.util.Optional;
 import java.util.UUID;
 
 @Component
+@ConditionalOnExpression("${logs.enabled:true}")
```
chrismatix, line 16: Allows us to disable logs for tests

juan-sebastian (approved):
Looks good to me. Just one thing you could change `kafka.commit-interval-ms=0` to commit as soon as possible for tests

PR #2747, opened 2022-01-18
[#2745] conversations.refetch endpoint

### LINKED ISSUE #2745: Refetch endpoint for conversations
There are some sources that send metadata (contact information) as in a push form (with every message). But there are others that don't. The goal if this endpoint `converstaions.refetch` is to give the possibility to the client to update the meta data for a specific conversation.

see related issue #2693

juan-sebastian:
closes #2745

### DIFF
--- a/backend/api/communication/src/main/java/co/airy/core/api/communication/ConversationsController.java
+++ b/backend/api/communication/src/main/java/co/airy/core/api/communication/ConversationsController.java
@@ -291,7 +291,7 @@ public class ConversationsController {
         final Metadata metadata = newConversationMetadata(conversationId, MetadataKeys.ConversationKeys.Contact.FETCH_STATE, "");
 
         try {
-            stores.storeMetadata(metadata);
+            stores.deleteMetadata(metadata);
         } catch (Exception e) {
             return ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).body(new RequestErrorResponsePayload(e.getMessage()));
         }
--- a/backend/api/communication/src/main/java/co/airy/core/api/communication/ConversationsController.java
+++ b/backend/api/communication/src/main/java/co/airy/core/api/communication/ConversationsController.java
@@ -276,4 +276,26 @@ public class ConversationsController {
         return ResponseEntity.noContent().build();
     }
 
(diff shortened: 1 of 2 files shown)


backend/api/communication/src/main/java/co/airy/core/api/communication/ConversationsController.java
```diff
@@ -276,4 +276,26 @@
...
+    ResponseEntity<?> conversationMetadataRefetch(@RequestBody @Valid ConversationByIdRequestPayload payload) {
+        final String conversationId = payload.getConversationId().toString();
+        final ReadOnlyKeyValueStore<String, Conversation> store = stores.getConversationsStore();
+        final Conversation conversation = store.get(conversationId);
+
+        if (conversation == null) {
+            return ResponseEntity.notFound().build();
+        }
+
+        // By removing the FETCH_STATE from the metadata that will trigger the stream to refetch the metadata for the specific conversation
```
chrismatix, line 289:
```suggestion
        // Removing the FETCH_STATE from the metadata that will trigger the source to refetch the metadata for the specific conversation (if supported)
```

backend/api/communication/src/main/java/co/airy/core/api/communication/Stores.java
```diff
@@ -217,7 +217,9 @@ public void storeReadReceipt(ReadReceipt readReceipt) throws ExecutionException,
     }
 
     public void storeMetadata(Metadata metadata) throws ExecutionException, InterruptedException {
```
chrismatix, line 219: Nitpick: This method is also used in other places. I'm not sure if we want to introduce metadata value "" == delete as a **global** convention. Could you here maybe add a `deleteMetadata`? :)

juan-sebastian: Not sure if saving empty strings `""` on the store is useful anyway. I am missing something?

chrismatix: True. Perhaps we can keep this until someone complains about their empty strings not being saved 👍

chrismatix (approved):

chrismatix (approved):

PR #2767, opened 2022-01-24
[#2764] Bug null pointer exception on api contacts service

### LINKED ISSUE #2764: NullPointerException on api-contacts service
In some clusters `api-contacts` service crashes with a `NullPointerException`

juan-sebastian:
closes #2764 

There where two issues
* When the stream is not ready
* Some messages are empty (null)


chrismatix (approved):
Thanks, looks good.

PR #2772, opened 2022-01-25
[#2769] Added custom host url

### LINKED ISSUE #2769: Make hostname optional for the frontend-chatplugin
**Current state**
Currently the frontend-chatplugin app uses the `API_HOST` environmental variable to determine the location of the backend. This env variable must be set, otherwise the app will not work.

**Desired state**
We want to make that env variable optional and if it is not set, the code snippet for the chatplugin should contain as default the same `{protocol}://{hostname}` as the customer is using at the moment to load the frontend-ui.

We can also create a field in the configuration of the chatplugin where this value can be overwritten.

Related to  #1761

thorstenairy:
closes #2769

### DIFF
--- a/frontend/ui/src/pages/Channels/Providers/Airy/ChatPlugin/sections/CustomiseSection.tsx
+++ b/frontend/ui/src/pages/Channels/Providers/Airy/ChatPlugin/sections/CustomiseSection.tsx
@@ -1,5 +1,5 @@
 import React, {createRef, useEffect} from 'react';
-import {Button, Dropdown, Input, ListenOutsideClick, Toggle} from 'components';
+import {Button, Dropdown, ErrorNotice, Input, ListenOutsideClick, Toggle} from 'components';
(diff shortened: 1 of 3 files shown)


frontend/ui/src/pages/Channels/Providers/Airy/ChatPlugin/sections/CustomiseSection.tsx
```diff
@@ -19,11 +19,13 @@ enum BubbleState {
 
 interface CustomiseSectionProps {
   channelId: string;
-  host: string;
+  host?: string;
 }
 
 export const CustomiseSection = ({channelId, host}: CustomiseSectionProps) => {
   const useLocalState = getUseLocalState(channelId);
+  const defaultHost = 'http://airy.core';
```
chrismatix, line 27: The default host should probably be the one the web app is running on currently, since "airy.core" is only for development

AudreyKj (approved):

PR #2785, opened 2022-01-31
[#2742] list objects in s3 bucket

### LINKED ISSUE #2742: Media Resolver health check requires ListAllBuckets permission
The [health check](https://github.com/airyhq/airy/blob/12d2ff5d2a37424ae8a3265fe8792e15f8e544fd/backend/media/src/main/java/co/airy/core/media/services/MediaUpload.java#L81) uses the ListAllBuckets method to check if the S3 credentials are correct.

This requires access to all buckets of the account. In order to follow the Principle of Least Privilege, the health check should try to put a file in the exact `s3bucket` and in the `s3path` as specified in the `media-storage` config map.

juan-sebastian:
closes #2742


juan-sebastian:
@pascal-airy could you please test this with a real S3 config? 🙏

chrismatix (approved):
Looks good to me.

pascal-airy (approved):
Screen Shot 2022-02-01 at 17 04 09
Tested successfully with those permissions

PR #2786, opened 2022-02-01
[#2641] Internally display message sender 

### LINKED ISSUE #2641: Internally display message sender
Currently, members can log into their instance and send messages but there is no record of who sends which messages.

Ideally, we should display (internally) which member sends which message. I know that we don't currently have a user list so we have to figure out where this information comes from.

It might also make sense to show a name linked to the member's email address - for example: we would know that liz@airy.co is Liz Hutter (or I would have a unique ID like LH2619)

We could render the member's name or unique ID either below the message/above the timestamp (on all messages sent by the organization member), or to the right of the timestamp (happy to discuss this and get feedback).

We should also give the option to hide this feature if the organization does not want to use it.

AudreyKj:
closes #2641 

**screenshot** 
Screenshot 2022-02-01 at 09 34 28

### DIFF
--- a/frontend/chat-plugin/lib/src/components/chat/index.tsx
+++ b/frontend/chat-plugin/lib/src/components/chat/index.tsx
@@ -39,6 +39,7 @@ const defaultWelcomeMessage: Message = {
   deliveryState: DeliveryState.delivered,
   fromContact: false,
   sentAt: new Date(),
+  sender: {id: '12345'},
 };
 
 type Props = AiryChatPluginConfiguration;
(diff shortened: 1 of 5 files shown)


thorstenairy (approved):

M-Shorouk:
@AudreyKj @thorstenairy nice one, where i can see/test it? 
Also, it wasn't in the last release, although it was merged on the same day of last release, right?

thorstenairy:
@M-Shorouk we gonna push the lastest develop image into staging so you can test on staging.airy.co, will ping you when it's done

PR #2792, opened 2022-02-01
[#2791] Contacts service should be off by default

### LINKED ISSUE #2791: Contacts service should be off by default
## Describe the bug

Currently the `api-contacts` service starts by default with a new installation. This should not be the case.

## To Reproduce

1. Install and start a fresh Airy instance

## Expected behavior

To run contacts you have to configure it in the airy yaml first.

chrismatix:
resolves #2791

### DIFF
--- a/backend/api/contacts/src/main/java/co/airy/core/contacts/dto/Contact.java
+++ b/backend/api/contacts/src/main/java/co/airy/core/contacts/dto/Contact.java
@@ -99,12 +99,12 @@ public class Contact implements Serializable {
             }
 
             return this.toBuilder()
-                    .addressLine1(Optional.ofNullable(this.getAddressLine1()).orElse(address.getAddressLine1()))
-                    .addressLine2(Optional.ofNullable(this.getAddressLine2()).orElse(address.getAddressLine2()))
-                    .city(Optional.ofNullable(this.getCity()).orElse(address.getCity()))
-                    .country(Optional.ofNullable(this.getCountry()).orElse(address.getCountry()))
-                    .postalCode(Optional.ofNullable(this.getPostalCode()).orElse(address.getPostalCode()))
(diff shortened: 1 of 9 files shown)


backend/api/contacts/src/main/java/co/airy/core/contacts/dto/Contact.java
```diff
@@ -99,12 +99,12 @@ public Address merge(Address address) {
             }
 
             return this.toBuilder()
-                    .addressLine1(Optional.ofNullable(this.getAddressLine1()).orElse(address.getAddressLine1()))
```
chrismatix, line 102: This was the wrong way around. Third time's a charm 😄

backend/api/contacts/src/test/java/co/airy/core/contacts/ImportContactsTest.java
```diff
@@ -72,77 +60,57 @@ void beforeEach() throws Exception {
...
+                .andExpect(status().isCreated());
 
-        List<ContactResponsePayload> contactsResp = objectMapper.readValue(
-                importContent,
-                new TypeReference<List<ContactResponsePayload>>() {});
-
-        ContactsList cl = new ContactsList();
-        retryOnException(cl, "Not able to get contacts list");
-
-        final ListContactsResponsePayload contactsList = objectMapper.readValue(cl.getListContent(), ListContactsResponsePayload.class);
```
chrismatix, line 106: Small refactor. Generally we try to avoid using our own payload POJOs to test APIs, because that way we don't check what clients actually receive.

backend/api/contacts/src/main/java/co/airy/core/contacts/dto/Contact.java
juan-sebastian: 🙈

juan-sebastian (approved):
Looks good to me

PR #2847, opened 2022-02-15
[#2722] Webhook subscription editable and resubscribable

### LINKED ISSUE #2722: Webhook API: Make webhook subscription editable and resubscribable

Currently, the only way to update a webhook subscription as an Airy developer
e.g. in order to change the name or to modify the subscribed events or the subscribed URL is to 
(1.) unsubscribe an existing webhook via `POST /webhooks.unsubscribe`, and then
(2.) create a new webhook subscription via `POST /webhooks.subscribe`,
eventually leading to a long polluted result list when calling `POST /webhooks.list` 
containing a lot of old webhook subscriptions all having the state `unsubscribed`

(1.) I would suggest to introduce a new route `POST /webhooks.update`

Request:
```JSON5
{
   "id": "SOME_ID", // required, uuid
   "name": "SOME_NAME", // optional, string
   "url": "SOME_URL" // optional, string
   "headers": {} // optional, object
   "events": [] // optional, array.
   "signature_key": "SOME_KEY" // optional, string
}
```
Response: 

`200, returning updated Webhook payload`

AND (2): Resubscribe existing webhooks

Make `POST /webhooks.subscribe` also accepts a body with `{"id": "SOME_ID"}`  to resubscribe an existing webhook subscription (instead of only accepting a full webhook payload, resulting in creating an additional webhook subscription).

Response:  `200, returning updated Webhook payload`

- (1): We could add `POST /webhooks.delete` with body `{ "id": "SOME_ID" }`  in order to clean up `POST /webhooks.list`; or
- (2): We could remove unsubscribed webhook subscriptions per default from the result of `POST /webhooks.list` so it doesn't contain unsubscribed webhooks anymore

## Additional context

-

abhi-airy:
Fixes #2722

### DIFF
--- a/backend/api/admin/src/main/java/co/airy/core/api/admin/payload/WebhookListResponsePayload.java
+++ b/backend/api/admin/src/main/java/co/airy/core/api/admin/payload/WebhookListResponsePayload.java
@@ -10,5 +10,5 @@ import java.util.List;
 @NoArgsConstructor
 @AllArgsConstructor
 public class WebhookListResponsePayload {
-    private List<WebhookResponsePayload> data;
+    private List<WebhookListPayload> data;
 }
--- a/docs/docs/api/webhook.md
+++ b/docs/docs/api/webhook.md
@@ -100,6 +100,26 @@ Subscribes the webhook for the first time or update its parameters.
 
 Update the webhook parameters.
 
+**Sample request**
+
+```json5
+{
+  "id": "3e639566-29fa-450d-a59f-ae3c25d7260f", // required
+  "url": "https://endpoint.com/webhook", // optional
+  "events": ["message.created", "message.updated", "conversation.updated", "channel.updated"], // optional
+  "headers": {
+    "X-Custom-Header": "e.g. authentication token"
+  }, // optional
+  "signature_key": "secret-key-for-hmac-header" // optional
+}
+```
+
+- `id` (required) provide for updates
+- `url` (optional) Endpoint to be called when sending events.
+- `headers` (optional) HTTP headers to set on each request (useful for authentication)
+- `signature_key` (optional) when set, the webhook will also send a header `X-Airy-Content-Signature` that contains the SHA256 HMAC of the specified key and the content.
+- `events` (optional) List of event types to receive. [See below](#events) for a detailed list. Omit to receive all event types.
+
 **Sample response**
 
 ```json5
--- a/docs/docs/api/webhook.md
+++ b/docs/docs/api/webhook.md
@@ -94,15 +94,45 @@ Subscribes the webhook for the first time or update its parameters.
 }
 ```
 
+## Update
+
+`POST /webhooks.update`
+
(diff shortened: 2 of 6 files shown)


backend/api/admin/src/main/java/co/airy/core/api/admin/payload/WebhookListResponsePayload.java
```diff
@@ -10,5 +10,5 @@
 @NoArgsConstructor
 @AllArgsConstructor
 public class WebhookListResponsePayload {
-    private List<WebhookResponsePayload> data;
+    private List<WebhookListPayload> data;
```
chrismatix, line 13: Why can we not use `WebhookResponsePayload` here?

docs/docs/api/webhook.md
```diff
@@ -94,23 +94,52 @@ Subscribes the webhook for the first time or update its parameters.
 }
 ```
 
+## Update
+
+`POST /webhooks.update`
+
+Update the webhook parameters.
+
+**Sample response**
```
chrismatix, line 103: If this is the response payload then we still lack a sample request payload

chrismatix (commented):
Looks good sofar. Minor comments.

docs/docs/api/webhook.md
abhi-airy: Sure I will add sample request too.

backend/api/admin/src/main/java/co/airy/core/api/admin/payload/WebhookListResponsePayload.java
chrismatix: That's true and I think in this PR we can leave it like that. 

But for a more general discussion: Currently when you don't set a status on `WebhookResponsePayload` the status is returned as:
```json
{
"status": null
}
```

@AitorAlgorta As far as I can recall we had introduced this default in the old system to make it easier for mobile clients to parse payloads. Perhaps it's worth revisiting this discussion because in my opinion, (and it's just taste I guess) `null` values on JSON payloads add more clutter than information. It would also allow us to recycle payloads as in the example above.

cc @abhi-airy @AitorAlgorta @juan-sebastian what's your opinion on this default? Should we consider changing it? If so we can open a new ticket for discussion and implementation :)

chrismatix (dismissed):

backend/api/admin/src/main/java/co/airy/core/api/admin/payload/WebhookListResponsePayload.java
juan-sebastian: For me the response should not include `null` values (see [ignore null values](https://www.baeldung.com/jackson-ignore-null-fields)), the fort-end can assume that a missing value is a `null` value. This way the response body is not **clutter** as you said.

juan-sebastian (approved):
Looks good to me. Nice first PR 💪

PR #2879, opened 2022-02-22
Point custom instance from UI local

### LINKED ISSUE #2774: Ability to point to remote instance from local UI
We want to have the ability to not just point to the local instance but also be able to point to some remote instances. 

Instances to be pointed to: 
- staging.airy.co
- demo.airy.co

AitorAlgorta:
closes #2774 

Right now the websocket doesn't work if you point to an instance

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,4 @@
+# Copy and paste this file to .env and edit these values
+# Useful for connecting to remote instances
+API_HOST=http://airy.core
+SYSTEM_TOKEN=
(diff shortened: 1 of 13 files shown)


AudreyKj (approved):

chrismatix (changes requested):
Not ready yet

chrismatix (commented):
lgtm

chrismatix (approved):

PR #2896, opened 2022-03-03
[#2627] prevent 404 responses for the messages send endpoint

### LINKED ISSUE #2627: Prevent 404 responses for the messages.send endpoint

Often we want to immediately react to a message of a new conversation, but the send message controller may not yet have that conversation. Then the client needs to retry with a 404.

Always accept the message with a `202` and if the conversation ended up not existing relay the error async.

juan-sebastian:
closes #2627


backend/api/communication/src/main/java/co/airy/core/api/communication/SendMessageController.java
```diff
@@ -78,22 +78,33 @@
...
+                // If channelId is not set we should not set the conversationId either
+                .setConversationId(channelId != "" ? conversationId : "")
                 .setHeaders(Map.of())
                 .setDeliveryState(DeliveryState.PENDING)
-                .setSource(channel.getSource())
+                .setSource(Optional.ofNullable(channel).map((c) -> c.getSource()).orElse(""))
                 .setSenderId(userId)
                 .setSentAt(Instant.now().toEpochMilli())
                 .setIsFromContact(false)
                 .build();
```
juan-sebastian, line 97: @chrismatix I am not sure if the system can tolerate empty values on the `ConversationId` `ChannelId` and `Source`

docs/docs/api/endpoints/messages.md
```diff
@@ -81,6 +81,16 @@ to see learn how to send text, media, and many more message types.
...
 
+Because of the nature of being an asynchronous system, sometimes the conversation is not ready at the time of sending
+the message.
+
+`HTTP 200 OK`
+In this case the converstion was found and the message is attached to it
+
+`HTTP 202 ACCEPTED`
+In this case the converstion is not ready yet. the message will be attached to the converstion
+asynchronously, in the case of an error. it will be delivered asynchronously. See error handling section.
```
chrismatix, line 92:
```suggestion
In this case, the conversation is not ready yet. the message will be attached to the conversation
asynchronously, in the case of an error. it will be delivered asynchronously. See error handling section.
```

backend/api/communication/src/main/java/co/airy/core/api/communication/SendMessageController.java
chrismatix: It can't actually. So my suggestion would be, having thought about this again, that we don't even need to store the placeholder message. We can only store the error metadata. 

This way the sender gets informed async without creating broken conversations.

docs/docs/api/endpoints/messages.md
chrismatix: So the original ticket says to always respond with a `202`. Is there a reason for doing it differently? :)

repositories.bzl
```diff
@@ -7,6 +7,7 @@ airy_jvm_deps = [
     "com.fasterxml.jackson.module:jackson-module-afterburner:2.10.0",
     "com.github.everit-org.json-schema:org.everit.json.schema:1.12.2",
     "com.google.auth:google-auth-library-oauth2-http:0.20.0",
+    "com.google.guava:guava:29.0-jre",
```
chrismatix, line 10: This doesn't seem used so let's get rid of it

repositories.bzl
```diff
@@ -20,6 +21,7 @@ airy_jvm_deps = [
     "io.jsonwebtoken:jjwt-jackson:0.10.5",
     "io.micrometer:micrometer-registry-prometheus:1.6.5",
     "javax.activation:javax.activation-api:1.2.0",
+    "javax.annotation:javax.annotation-api:1.3.2",
```
chrismatix, line 24: Duplicate dependency? (let's just upgrade)

backend/api/communication/src/main/java/co/airy/core/api/communication/AsyncSendMessagesHandler.java
```diff
@@ -0,0 +1,159 @@
...
+                    Message msg = pendingMsg.getMsg();
+
+                    if (conversation != null) {
+                        Channel channel = conversation.getChannel();
+                        if (channel.getConnectionState().equals(ChannelConnectionState.DISCONNECTED)) {
+                            msg = setMessageSateToFailed(msg, "Channel not connected");
+                        }
+                        updateMessage(conversation, msg);
+
+                    } else if (messageExpired) {
```
chrismatix, line 105: `messageExpired` is always true because of the if check in line 92

backend/api/communication/src/main/java/co/airy/core/api/communication/AsyncSendMessagesHandler.java
```diff
@@ -0,0 +1,159 @@
...
+        this.thread = new Thread(this);
+    }
+
+    @Override
+    public void run() {
+        final List<PendingMessage> pending = new LinkedList<>();
+
+        while (keepAlive) {
+            try {
+                Thread.sleep(this.periodMillis);
```
chrismatix, line 75: I think an easier way to do this would be to use [`@Scheduled`](https://github.com/airyhq/airy/blob/0f3ab3554fbc41b0fe1f44d04fb32c858a1c84b3/backend/api/admin/src/main/java/co/airy/core/api/config/ServiceDiscovery.java#L53)

But in general I wonder if there is a way that lets us avoid maintaining this linked list as it's prone to fail. How about having something like
```java
@Async 
public void sendMessage(Message message) {
}

@Scheduled() // with some config
private void trySendMessage(Message message) {
// ...
}
```
that relies on the async executor service we use for PNS and the [`@Retry`](https://www.baeldung.com/spring-retry) annotation.

More info: https://docs.spring.io/spring-batch/docs/current/reference/html/retry.html

What do you think?

backend/api/communication/src/main/java/co/airy/core/api/communication/SendMessageController.java
```diff
@@ -78,22 +77,32 @@
...
 
-        producer.send(new ProducerRecord<>(applicationCommunicationMessages.name(), message.getId(), message)).get();
-        return ResponseEntity.ok(MessageResponsePayload.fromMessageContainer(new MessageContainer(message, new MetadataMap())));
+        // If there is no channelId it implies that the converstaion was not found
+        // instally and it will be handled asynchronously
+        HttpStatus s = HttpStatus.OK;
```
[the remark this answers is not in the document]
juan-sebastian: I think it is ok to have variables with this names as long as the scope is very short. In this case the variable `s` it is only used at the return and that is it.

docs/docs/api/endpoints/messages.md
```diff
@@ -81,6 +81,16 @@ to see learn how to send text, media, and many more message types.
 
 **Sample response**
 
+Because of the nature of being an asynchronous system, sometimes the conversation is not ready at the time of sending
+the message.
+
+`HTTP 200 OK`
+In this case the conversation was found and the message is attached to it
+
+`HTTP 202 ACCEPTED`
```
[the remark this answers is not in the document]
juan-sebastian: If the conversation is ready there is no need to do this asynchronously. basically from the caller prospective `200` and `202` are basically the same call. the only thing that the caller need to be aware is that some other calls like `message.list` or `message.suggest-replies` won't reply with the expected data if the message is not ingested yet.

But for the propuse of the issue that is to be able to react instantly to a message it can be treated like a regular `200`

backend/api/communication/src/main/java/co/airy/core/api/communication/AsyncSendMessagesHandler.java
juan-sebastian: It is inside a `else if` because it can be a case where the message expired but the conversation is there. so in that case we just want to ingest the message.

backend/api/communication/src/main/java/co/airy/core/api/communication/SendMessageController.java
```diff
@@ -78,22 +77,32 @@
...
+                .setSource(Optional.ofNullable(channel).map((c) -> c.getSource()).orElse(""))
                 .setSenderId(userId)
                 .setSentAt(Instant.now().toEpochMilli())
                 .setIsFromContact(false)
                 .build();
 
-        producer.send(new ProducerRecord<>(applicationCommunicationMessages.name(), message.getId(), message)).get();
-        return ResponseEntity.ok(MessageResponsePayload.fromMessageContainer(new MessageContainer(message, new MetadataMap())));
+        // If there is no channelId it implies that the converstaion was not found
+        // instally and it will be handled asynchronously
```
chrismatix, line 97:
```suggestion
        // If there is no channelId it implies that the conversation was not found
        // instantly and it will be handled asynchronously
```

docs/docs/api/endpoints/messages.md
```diff
@@ -81,6 +81,16 @@ to see learn how to send text, media, and many more message types.
...
+In this case the conversation is not ready yet. the message will be attached to the conversation
+asynchronously, in the case of an error. it will be delivered asynchronously. See error handling section.
+
```
chrismatix, line 93: Small rewrite to keep it inline with our "Sample response (status code)" structural convention:

```suggestion
**Sample response (202)**

Because of the nature of being an asynchronous system, sometimes the conversation is not ready at the time of sending the message. The status code `202` indicates that this is the case and also does not include a response body. In case the conversation was eventually not found error metadata will be delivered asynchronously using the [webhook](api/webhook.md) and [websocket]((api/websocket.md)).

**Sample response (200)**

The conversation was found and the message is pending to be delivered.

```

backend/api/communication/src/main/java/co/airy/core/api/communication/AsyncSendMessagesHandler.java
chrismatix: That's true, but what I'm saying is that if the conversation is not null, and the else gets triggered the message has to be expired because of the check. The `else if` is redundant.

PR #2919, opened 2022-03-10
[#2918] Resend messages

### LINKED ISSUE #2918: Resend message

It would be great to be able to resend messages that are in a `Failed` state

An endpoint `/messages.resend` that only resends messages that were failed

chrismatix:
resolves #2918

### DIFF
--- a/backend/api/communication/src/main/java/co/airy/core/api/communication/SendMessageController.java
+++ b/backend/api/communication/src/main/java/co/airy/core/api/communication/SendMessageController.java
@@ -4,12 +4,15 @@ import co.airy.avro.communication.Channel;
 import co.airy.avro.communication.ChannelConnectionState;
 import co.airy.avro.communication.DeliveryState;
 import co.airy.avro.communication.Message;
+import co.airy.core.api.communication.payload.ResendMessageRequestPayload;
 import co.airy.core.api.communication.payload.SendMessageRequestPayload;
 import co.airy.model.conversation.Conversation;
+import co.airy.model.message.MessageRepository;
 import co.airy.model.message.dto.MessageContainer;
 import co.airy.model.message.dto.MessageResponsePayload;
 import co.airy.model.metadata.dto.MetadataMap;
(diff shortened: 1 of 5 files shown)


backend/api/communication/src/main/java/co/airy/core/api/communication/SendMessageController.java
```diff
@@ -105,4 +108,26 @@
...
         return ResponseEntity.status(s).body(MessageResponsePayload.fromMessageContainer(new MessageContainer(message, new MetadataMap())));
     }
+
+    @PostMapping("/messages.resend")
+    public ResponseEntity<?> sendMessage(@RequestBody @Valid ResendMessageRequestPayload payload) {
+        final String messageId = payload.getMessageId().toString();
+
+        final MessageContainer messageContainer = stores.getMessageContainer(messageId);
+        if (messageContainer == null) {
+            return ResponseEntity.status(HttpStatus.NOT_FOUND).build();
```
juan-sebastian, line 118: Just to be aware the in the case that the conversation is never ready. The message won't be ingested so this will always return `404`

backend/api/communication/src/main/java/co/airy/core/api/communication/SendMessageController.java
```diff
@@ -105,4 +108,26 @@
...
+        }
+
+        final Message message = MessageRepository.markMessageForResend(messageContainer.getMessage());
+        try {
+            stores.storeMessage(message);
+        } catch (InterruptedException | ExecutionException e) {
+            return ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).body(e.getMessage());
+        }
+
+        return ResponseEntity.status(HttpStatus.ACCEPTED).build();
```
juan-sebastian, line 131: I think it will be better to return `200`. I know that `stores.storeMessage(message)` is async. but to keep consistency with other endpoint in this app and the rest

juan-sebastian (approved):

PR #2922, opened 2022-03-11
[#1279] Helm template on local chart with bazel

### LINKED ISSUE #1279: Introduce a helm_test rule
There is currently no validation for the Airy Core helm charts in `infrastructure/helm-chart`. Often they get bundled in the container and the helm templating syntax, which is prone to errors, crashes at runtime

**Proposal**

Add a Bazel test rule `helm_test` that runs [helm template](https://helm.sh/docs/helm/helm_template/) on the directory to see if there are any syntax errors _given the default configuration

```python
load("//tools/test:helm_test.bzl", "helm_test")

helm_test(
    values = "values.yaml",
    srcs = ["Chart.yaml"] + glob(["charts/**/*", "templates/**/*"])
)
```

### LINKED ISSUE #2256: Split stable and develop helm chart repositories
Introduce separate repositories for stable and unstable charts of `Airy Core`, the prerequisites and the tools.
Instead of having them behind an S3 url (https://airy-core-helm-charts.s3.amazonaws.com/stable), put them behind a more user-friendly hostname (example: https://helm.airy.co and https://helm.airy.co/develop).

ljupcovangelski:
Resolves  #2256 
Resolves #1279


tools/build/helm.bzl
```diff
@@ -0,0 +1,17 @@
+load("@com_github_airyhq_bazel_tools//helm:helm.bzl", "helm_push_")
```
chrismatix, line 1: Forgot to mention this, but you can also rename imports like so:
```
load("@com_github_airyhq_bazel_tools//helm:helm.bzl", lib_helm_push = "helm_push")
```

For when you want to avoid name collisions :)

ljupcovangelski: Ah cool, I will rename the function in the `bazel-tools` repo - only `helm_push` is more appropriate for it. :+1:

infrastructure/helm-chart/charts/prerequisites/charts/beanstalkd/BUILD
```diff
@@ -0,0 +1,29 @@
...
+)
+
+helm_template_test(
+    name = "template",
+    chart = ":package",
+)
+
+helm_push_version(
+    chart = ":package",
+    version = "1.0"
```
chrismatix, line 28: Just a suggestion, but you could also wrap all of these rules up in a macro that only depends on the version. I.e.: 
```
helm_package(version = "1.0")
```

ljupcovangelski: Thanks, good suggestion :+1: . As we have two rules - `helm_push` and `helm_push_version`, I assume we will make some more improvements here, so I'd leave this to some next iteration.

juan-sebastian (commented):
Looks good to me

PR #2923, opened 2022-03-11
[#2892] Failed message indicator

### LINKED ISSUE #2892: Display an indication for a failed message on Airy UI
### Is your feature request related to a problem? Please describe?

On Airy UI, when the user agent sends a message and it fails for any reason, this message does not go through our Airy instance to pass it to the relative channel. However, the user agent doesn't know that this message has failed to be sent. 

The problem here is in the payload sent to Airy instance with the message ```
```
"content": {
  "text": "x"}
``` 
already contains the "failed content message". Allowing the user agent to experience a feedback for this failed message, maybe by a red cross next to the text", would send another payload to Airy instance with "new"
 ```
"content": {
  "text": "x"}
```
++ Adding a screenshot from the payload with failed state 

```
{
"payload": {
"conversation_id": "f98969b1-7cb7-50d4-a23c-540a4e31aed6",
"channel_id": "9cf5f1d7-eaa9-53ac-9d9e-0f11a46273c5",
 "message": {
 "id": "4a11de87-c2c5-4929-a239-34fd92e3ecbe",
 "content": {
  "text": "testing outbound",
  "representative": {
  "representativeType": "HUMAN"
  }
  },
 "sent_at": "2022-02-02T09:21:24.284Z",
 "delivery_state": "failed",
 "source": "google",
 "metadata": {
  "error": "Invalid PKCS#8 data."
 },
  "from_contact": false
  }
 },
 "timestamp": 1643793685082,
  "type": "message.updated"
  }

```

### Describe the solution you'd like:

Case scenario: 

1. User agent send a message. 
2. Payload sent to Websocket results via HTTP API "delivery_state" : "failed" .
3. Websocket send back error message.
4. Show  on message UI feedback that this message has failed to be sent. (Could be a red cross next to the text)
4. Allow user agent to "re-send" this message. (LOOP until HTTP API is updated to "status" :"Delivered ").
5. Remove the feedback icon (if it is a red cross).
6.Rend conversation without failed messages where "hide"? messages that contains "delivery_state" : "failed".

### PR DESCRIPTION
**Pull request by thorstenairy**
closes #2892

### CODE REVIEW

File: lib/typescript/websocketclient/payload.ts [hunk 1]
```diff
@@ -18,6 +18,20 @@ export interface MessageCreatedPayload extends Event {
...
+    channel_id: string;
+    message: {
+      id: string;
+      content: string;
+      delivery_state: DeliveryState;
+      from_contact: boolean;
+      sent_at: Date;
+    };
+  };
+}
```
AudreyKj, line 34: some types are missing in the payload 🤔: https://airy.co/docs/core/api/websocket#messageupdated

lib/typescript/components/message/MessageInfoWrapper/index.tsx
```diff
@@ -12,23 +13,67 @@ type MessageInfoWrapperProps = {
...
+    deliveryState,
+    messageId,
+    resendFailedMessage,
+  } = props;
 
   const isContact = isChatPlugin ? !fromContact : fromContact;
   const senderIdentity = sentAt ? ` - sent by ${senderName}` : `sent by ${senderName}`;
 
+  const handleResendFailedMessage = () => {
+    resendFailedMessage(true, messageId);
```
chrismatix, line 40: The name of this prop is confusing: The function doesn't actually resend the message but marks it as to be resend (?)

Also the retry is in the wrong place. I'd suggest that the class only exposes a `onRetryMessage` that gets passed to the `onClick` handler

chrismatix (commented):
LGTM, minor comments

frontend/ui/src/actions/messages/index.ts
```diff
@@ -55,6 +61,19 @@ export function sendMessages(messagePayload: SendMessagesRequestPayload) {
   };
 }
 
+export function resendMessages(messagePayload: ResendMessagesRequestPayload) {
```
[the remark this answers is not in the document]
thorstenairy: just noticed that we use plural for sending a message (send.messages) and the endpoint is plural (messages.resend)
we might change all to singular then? @AitorAlgorta @chrismatix

AudreyKj (approved):

PR #2924, opened 2022-03-11
[#2774] Fix: point staging from UI local

### LINKED ISSUE #2774: Ability to point to remote instance from local UI

chrismatix:
resolves #2774


AitorAlgorta (approved):
🚀

PR #2934, opened 2022-03-15
[#2933] fix inbound audio messages bug 

### LINKED ISSUE #2933: Fix inbound audio messages bug
Some inbound audio messages are crashing because they come from a source who stores its media in a CORS restricted bucket. Thus, calling fetch on these resources does not work and triggers a CORS issue (which means that the waveform cannot be drawn on the component). 

**to do:** 
- fix the audioClip component so that it would display a random audio frequencies if it's not possible to fetch its audio data
- handle error in case of broken audio files (for ex. when there is not supported source found)

AudreyKj:
closes #2933


lib/typescript/components/general/AudioClip/index.tsx
```diff
@@ -108,8 +107,19 @@ export const AudioClip = ({audioUrl}: AudioRenderProps) => {
...
+    if (playPromise) {
+      playPromise
+        .then(() => {
+          audioElement.current.play();
+          setIsPlaying(true);
+        })
+        .catch(() => {
+          setError(true);
+        });
+    }
```
AudreyKj, line 122: this handle errors in case the audio file is broken (for ex. no supported source found): in this case, the component displays 'could not load audio'

lib/typescript/components/general/AudioClip/services/generateRandomFrequencies.ts
```diff
@@ -0,0 +1,7 @@
+const generateRandomNumber = () => {
+  return Math.floor(Math.random() * 18);
+};
+
+export const generateRandomFrequencies = (): number[] => {
+  return Array(20).fill(0).map(generateRandomNumber);
+};
```
AudreyKj, line 7: this generates random audio frequencies

AitorAlgorta (approved):

PR #3002, opened 2022-04-07
[#2900] rendering bug in messagelist

### LINKED ISSUE #2900: Rendering bug in MessageList
## Describe the bug

Messages re-render in `MessageList` when: 
- a new message is sent
- contacts is edited
- a tag is added/deleted 
- state is changed

note: this is not a new bug, we've noticed this since we support media messages. We also have the same bug in the chat plugin #3003 

## To Reproduce

1 - navigate to `ui/inbox` and create a conversation 
2 - send several videos & audios attachments 
3 - send a text message and see how the videos & audios message reload
4- edit contacts and see how the videos & audios message reload
5- add/delete tag in conv Metadata and see how the videos & audios message reload
6- change state and see how the videos & audios message reload

## Expected behavior

Media attachments messages should not reload when: 
- a new message is sent
- contacts is edited
- a tag is added/deleted 
- state is changed

## Environment

- OS: iOS
- Browser: Chrome, Safari, Firefox

### LINKED ISSUE #3003: rendering bug in the messenger of the chat plugin
On the chat plugin's messenger, all the media attachments re-render when the user types in the message input.

related to #2900 (fixed in the same PR as both bugs have the same cause)

https://user-images.githubusercontent.com/38159391/162139023-f8070812-86ef-4900-b64b-429547c673a9.mov

AudreyKj:
closes #2900 
closes #3003 

re-rendering fix 
---> reorganized the structure of the message wrapper 
---> removed `react.memo `from everywhere in the codebase (for now, we don't need it if the code is well structured)


lib/typescript/components/message/MessageWrapper/MessageContainer/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+import {SourceMessage, CommandUnion, ContentType} from 'render';
+import {DeliveryState, Source, Message} from 'model';
+import {ReactComponent as ErrorMessageIcon} from 'assets/images/icons/errorMessage.svg';
+import {Reaction} from '../../Reaction';
+import styles from './index.module.scss';
+
+interface MessageContainerProps {
+  message: Message;
+  source: Source;
+  isChatPlugin: boolean;
```
chrismatix, line 11: nitpick: It'd be great to get rid of this prop. Especially since it seems like the chatplugin itself can simply pass down the `invertSides` and `commandCallback` props.

frontend/ui/src/services/types/sourceTypes.ts
```diff
@@ -0,0 +1,10 @@
+import {Source} from 'model';
+
+export const getSourceType = (source: string): Source => {
```
chrismatix, line 3: I think you can simply access the values of an enum like so `Source[source]`, but I'm wondering what this even gets us? is it so that we get the `Source` enum type? In that case, this is the same as just casting, right?

chrismatix (commented):
Excellent improvements to the code! Small nitpicks.

frontend/ui/src/services/types/sourceTypes.ts
AudreyKj: good catch! yes, `Source[source] `gives the right value 👍

AitorAlgorta (approved):

PR #3055, opened 2022-04-21
Change path to /inbox

### LINKED ISSUE #3052: Change path from /ui to /inbox

AitorAlgorta:
closes #3052


lib/java/spring/auth/src/main/java/co/airy/spring/auth/AuthConfig.java
```diff
@@ -49,8 +49,8 @@ public AuthConfig(@Value("${systemToken:#{null}}") String systemToken,
                       @Value("${jwtSecret:#{null}}") String jwtSecret,
                       List<IgnoreAuthPattern> ignorePatternBeans,
                       ConfigProvider configProvider,
-                      @Value("${oidc.logoutSuccessUrl:/ui/}") String logoutSuccessUrl,
-                      @Value("${oidc.loginSuccessUrl:/ui/}") String loginSuccessUrl
+                      @Value("${oidc.logoutSuccessUrl:/control-center/}") String logoutSuccessUrl,
```
chrismatix, line 52: This one should be changed to `/start`

AitorAlgorta: we are discussing a bit because it's not really clear where the start should go and how it should behave

AitorAlgorta: we will change it to just `/` since it will redirect to the correct start page @chrismatix

ljupcovangelski (dismissed):
Checked and tested the helm chart and it works good :+1:

ljupcovangelski (approved):
Looks good :+1:

PR #3056, opened 2022-04-21
[#3031] component status

### LINKED ISSUE #3031: Component status
Here we are adding to the status view the dependent "services" on the component. 

_Successful User Story:_

1. User signs-in Control Center.
2. User view "Components" view; a list of segmented "Components"
     2.1.Component segmented health status (NB: The service status is only healthy/green when all dependent components are healthy.)
     2.2. Toggle button to turn ON/OFF "collection of deployment artefacts/Service".
3. User clicks on "Component" name
     3.1. user views expanded "Services" status shown with separate health status.

**0.2 Status list components**
DESIGN URL:  

https://www.figma.com/file/64dGaYVQCuV9aLAt0uF7mM/Control-Center?node-id=2%3A250

AudreyKj:
closes #3031


frontend/control-center/src/pages/Status/ComponentListItem/ItemInfo.tsx
```diff
@@ -0,0 +1,115 @@
...
+};
+
+export const ItemInfo = (props: ComponentInfoProps) => {
+  const {healthy, itemName, isComponent, isExpanded, setIsExpanded, enabled} = props;
+  const [channelSource, setChannnelSource] = useState('');
+  const [componentName, setComponentName] = useState('');
+  const [componentEnabled, setComponentEnabled] = useState(enabled);
+
+  useEffect(() => {
+    if (itemName.includes('twilio')) {
```
AitorAlgorta, line 28: We need to find a better solution for this in the future, we cannot keep adding here if we have 200 components

AitorAlgorta (approved):
Left one comment

frontend/control-center/src/pages/Status/ComponentListItem/ItemInfo.tsx
AudreyKj: Yeah we can actually use a function that takes in the name of the component as input and returns the second word capitalized for the title: eg. `sources-facebook` --> `Facebook`

I will implement it!

PR #3094, opened 2022-04-27
[#3092] Adding name to webhooks

### LINKED ISSUE #3092: webhooks.list should return all previous connected webhooks with status in the response

Currently webhooks.list just returns all "Subscribed" webhooks without the status in the response.

webhooks.list should return all webhooks which had been connected not only the "Subscribed" ones and should return it with status ("Subscribed" || "Unsubscribed").

### LINKED ISSUE #3093: webhooks.subscribe & webhooks.update do not accept name in request, but returns it in response

Right now we are not able to set a name when subscribing to a new webhook.

We need to extend the webhooks.subscribe and webhooks.update payload by name.

AitorAlgorta:
closes #3092 
closes #3093


juan-sebastian (approved):
Looks good to me

PR #3147, opened 2022-05-10
[#2755] Authenticate controller

### LINKED ISSUE #2755: Authentication to the Airy controller
**Background:** [See milestone](https://github.com/airyhq/airy/milestone/36)

**Motivation:** Since we need to expose controller APIs to clients we need to be able to authenticate against it.

- [x] Support Airy authentication (system token and jwt) on the controller service
- [x] Enable internal requests from `api-admin` to still work
- [ ] Add an `airy login` command so that the CLI can obtain system token authentication 
 

Perhaps we can use this routing lib: https://github.com/gorilla/mux

chrismatix:
required for #2755 

The `airy login` command will follow in a separate PR


infrastructure/controller/pkg/endpoints/server.go
```diff
@@ -0,0 +1,42 @@
...
+
+	"k8s.io/client-go/kubernetes"
+)
+
+func Serve(clientSet *kubernetes.Clientset, namespace string) {
+	r := mux.NewRouter()
+
+	// Load authentication middleware only if auth env is present
+	authEnabled := false
+	systemToken := os.Getenv("systemToken")
```
ljupcovangelski, line 18: Should we put this ENV also in the deployment.yaml file, from the security configMap?

ljupcovangelski (commented):
Looks great :+1:

infrastructure/controller/pkg/endpoints/server.go
chrismatix: Good catch

ljupcovangelski (dismissed):

juan-sebastian (commented):
Good overall. Left some minor comments

juan-sebastian (approved):

ljupcovangelski (approved):

PR #3149, opened 2022-05-10
Feature/3004 add service to every connector

### LINKED ISSUE #3004: Status and health micro service
Currently we don't have a proper way to get and monitor our apps in airy ecosystem. The Idea is to create (or see if there is a ready to use solution) a micro service that interacts directly with the kubernetes API.

juan-sebastian:
closes #3004


chrismatix (approved):

ljupcovangelski (approved):
Looks good :+1:

Issue #2756, opened 2022-01-21
Applying the configuration through the Airy controller


ljupcovangelski:
We need to be able to do `airy config apply` or use some other flag to apply the configuration through the `airy-controller` and not directly to Kubernetes.

If the controller exposes an endpoint to create/read/update/delete configuration, then we can also apply configuration for the sources and other components through the UI.

Depends on #2755 .

ljupcovangelski:
After internal discussions, there are two ways how we can approach this issue:

1. Build a full API service as part of the `airy-controller` that accepts create, update, delete and list of components. Having that, we would be able to manage every component independently.
2. Build a simple endpoint that accepts the whole airy.yaml, it un-marsals it and creates the resources in order to manage the components.

As we see the first option as a more long-term option, for now we will do the second option in order to remove the dependence of the Airy CLI on Kubernetes. The controller will have no state, the airy.yaml will be saved in a configMap, from where it will also be retrieved.

We would need to be able to send the following request, to send the `airy.yaml` file to the controller

```
curl -X POST -H "Content-Type: application/json" -d @airy.yaml http://airy.core/controller/config
```

This endpoint should be made functional once the dependent authentication issue is resolved.

We will need an option to also retrieve the current `airy.yaml` configuration, as well as manage concurrency requests to update the configuration. We will create separate issues for these items.

chrismatix:
To get the API proposal in the format of our docs I would propose something like this:

```POST /cluster.applyState```

```json5
{
    "data": "{stringified airy yaml}"
}
```

ljupcovangelski:
Currently the `airy.yaml` file holds both the values for helm and the configuration of the Airy components. The configuration of the Airy components is under:

```
security:
  ...
components:
  component-type:
    component-name:
      ...
```

On the long run, I see the `airy.yaml` file holding only the helm values, as the configuration will be applied mostly from the UI, or by posting JSON from the CLI. I advise not to write the configurations of the components back to the `airy.yaml` file, as those are mostly security credentials.

As for the values in the `airy.yaml` file, if it relates only to helm, they can be easily be retrieved if one has access to the Kubernetes cluster, using:

```
helm get values airy-release
```

In that sense was thinking to make it possible to now apply the configuration using a binary transfer of a YAML file for now and at the same time through JSON. The YAML should work only temporary while we maintain the legacy `config apply` in the CLI.

Eventually, configuration should be applied only through JSON this will be handled 

For example we would need to post this:

```
{
[...]

PR #3176, opened 2022-05-18
 [#2756] Apply config through the airy controller 

ljupcovangelski:
addresses #2756 

A draft PR for proof of concept and also to get some feedback :pray: .

The CLI returns this:

```
$ airy config apply
Configured component  Enabled
security              true
sources-facebook      true
sources-google        true
integration-webhook   true

```
The endpoint accepts `json` 

```
{
  "security": {
    "allowedOrigins": "*",
    "jwtSecret": "secret",
    "systemToken": "test"
  },
  "components": {
    "sources": {
      "facebook": {
        "token1": "token",
        "token2": "token"
      }
    }
  }
}
```

and also the legacy `yaml` format of the `airy.yaml` file.


lib/go/payloads/BUILD
```diff
@@ -1,15 +1,18 @@
 load("@com_github_airyhq_bazel_tools//lint:buildifier.bzl", "check_pkg")
 load("@io_bazel_rules_go//go:def.bzl", "go_library")
-# gazelle:prefix github.com/airyhq/airy/lib/go/httpclient/payloads
+# gazelle:prefix github.com/airyhq/airy/lib/go/payloads
```
ljupcovangelski, line 3: Moved the `payloads` outside of the `httpclient` package, because they are also used by the controller.

lib/go/config/airy_yaml.go
```diff
@@ -32,3 +32,8 @@ type AiryConf struct {
 	Security   SecurityConf              `yaml:"security"`
 	Components map[string]ComponentsConf `yaml:"components,omitempty"`
 }
+
+type ConfigComponent struct {
+	Removed   bool   `json:"removed"`
+	Component string `json:"component"`
+}
```
ljupcovangelski, line 39: This is currently only used by the ConfigApplyResponsePayload, so not sure if it should be here or in `payloads`. Also `Removed` is just to be in line with the current concept (when a configuration is applied without a component - that component is removed).

I'd be happy to remove this and add a separate endpoint for removing a component.

infrastructure/controller/pkg/endpoints/cluster_apply.go
```diff
@@ -0,0 +1,97 @@
...
+	body, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+	}
+
+	var conf config.AiryConf
+	namespace := "default"
+	responseConfigComponents := payloads.ConfigApplyResponsePayload{}
+
+	err = json.Unmarshal(body, &conf)
```
chrismatix, line 34: As mentioned I would suggest that we nest the config inside a `data` field. This will make the API more maintainable as we can add more options to the request body later when needed.

infrastructure/controller/pkg/endpoints/cluster_apply.go
```diff
@@ -0,0 +1,97 @@
...
+			http.Error(w, "Error parsing the configuration.\nError:\n"+err.Error(), http.StatusBadRequest)
+			return
+		}
+	}
+
+	secData := config.GetSecurityData(conf.Security)
+	if len(secData) != 0 {
+		applyErr := k8s.ApplyConfigMap("security", namespace, secData, map[string]string{}, s.clientSet)
+		if applyErr != nil {
+			klog.Error("Unable to apply configuration for \"security\"\nError:\n" + err.Error())
```
chrismatix, line 49: In this case, the request should fail and terminate early (same below)

infrastructure/controller/pkg/endpoints/cluster_apply.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	configmapList, _ := s.clientSet.CoreV1().ConfigMaps(namespace).List(context.TODO(), v1.ListOptions{LabelSelector: "core.airy.co/component"})
+	for _, configmap := range configmapList.Items {
+		if !configuredComponents[configmap.ObjectMeta.Name] {
+			deleteErr := k8s.DeleteConfigMap(configmap.ObjectMeta.Name, namespace, s.clientSet)
+			if deleteErr != nil {
+				klog.Error("Unable to remove configuration for component:" + configmap.ObjectMeta.Name + "\nError:\n" + err.Error())
+			} else {
+				responseConfigComponents = append(responseConfigComponents, config.ConfigComponent{
```
chrismatix, line 85: Could you please add some docs on what the response is supposed to look like?

infrastructure/controller/pkg/endpoints/cluster_apply.go
ljupcovangelski: > In this case, the request should fail and terminate early (same below)

I am not sure about this. The components are independent from each-other, so an error in deploying one component/configMap shouldn not impact the ones that follow after. 

We can return perhaps a list of failed components

chrismatix: It's true that we could do that but for clients, a completely failed state is easier to recover from than a state that is half failed. 

What do you return for a request that only did 50% of what you asked it to? `200` aka "success"? In that case, clients may not be aware that anything failed at all

ljupcovangelski: Yes, I get it :+1: We can return the `failed-components` in a separate list, but I think as it is very unlikely that only one configMap fails - I will set it to return a `500` - `InternalServerError`.
From the UI I assume it will mostly be used for applying a single component at a time anyway.

docs/docs/api/endpoints/components.md
```diff
@@ -0,0 +1,143 @@
...
+
+Update the configuration of a list of components.
+
+`POST /components.update`
+
+**Sample request**
+
+```json
+{
+  "Security": {
```
chrismatix, line 51: Payload needs to follow our `snake_case` convention

ljupcovangelski: > Payload needs to follow our `snake_case` convention

Thanks :+1:. I changed it for `security` and `data`. For the values which were in the `airy.yaml` (`allowedOrigins`, `systemToken`, `jwtSecret`, ...) I suggest we do it in a subsequent PR, as it has lots of changes across almost all components.

docs/docs/api/endpoints/components.md
```diff
@@ -0,0 +1,143 @@
...
+`POST /components.get`
+
+The list of the currently configured components is returned.
+
+**Sample response**
+
+```json5
+{
+  "security": {
+    "systemToken": "token",
```
chrismatix, line 21: We want to make this API the new source of truth for interacting with Airy. Therefore I'd argue that as with all other endpoints we also convert these keys into `snake_case` so that the interface is uniform.

docs/docs/api/endpoints/components.md
```diff
@@ -0,0 +1,143 @@
...
+  ]
+}
+```
+
+**Sample response**
+
+The list of the deleted components is returned.
+
+```json5
+{
```
chrismatix, line 137: Delete outer `{}` since this isn't valid json

infrastructure/controller/pkg/endpoints/components_update.go
```diff
@@ -0,0 +1,74 @@
...
+
+func (s *ComponentsUpdate) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+
+	body, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+	}
+
+	var conf payloads.ComponentsUpdateRequestPayload
+	namespace := os.Getenv("NAMESPACE")
```
chrismatix, line 29: I think we can use the variable that's already stored in `s.namespace` here

docs/docs/api/endpoints/components.md
```diff
@@ -0,0 +1,143 @@
...
+**Sample response**
+
+The list of configured components is returned.
+
+```json5
+[
+  {
+    "name": "security",
+    "enabled": true,
+    "data": null
```
chrismatix, line 95: Can data ever be `non-null`? If no, then I'd remove it.

Also: security isn't a component

lib/go/httpclient/components.go
```diff
@@ -7,6 +7,17 @@ import (
 	"github.com/airyhq/airy/lib/go/payloads"
 )
 
+//NOTE: As stated on httpclient.go for now we can use a general interface. because we are only getting
+//      and printing the data
+func (c *Client) ComponentsGet() (interface{}, error) {
+	components, err := c.get("components.get")
+	if err != nil {
```
chrismatix, line 14: Could we not simply write `return c.get("components.get")` here? 🤔

lib/go/httpclient/components.go
```diff
@@ -0,0 +1,43 @@
...
+
+	components.Components = config.GetComponents(conf)
+
+	payload, err := json.Marshal(components)
+	if err != nil {
+		return nil, err
+	}
+
+	res := payloads.ComponentsUpdateResponsePayload{}
+	e := c.post("components.update", payload, &res)
```
chrismatix, line 37: same here

juan-sebastian (commented):
Look good. Juste some small things to adresse and it will be good to merge

lib/go/httpclient/components.go
juan-sebastian: I will fix this in my PR https://github.com/airyhq/airy/pull/3221

juan-sebastian (approved):
I did some minor changes it looks good now to me

docs/docs/api/endpoints/cluster.md
```diff
@@ -0,0 +1,39 @@
+---
+title: Cluster
+sidebar_label: Cluster
+---
+
+The `cluster.*` API is about controlling the general configuration of the `Airy Core` Cluster, which is independent of the individual components. Currently the only supported argument is
```
chrismatix, line 6:
```suggestion
The `component.*` API controls the system-wide configuration of the `Airy` Cluster, which is independent of the individual components.
```

docs/docs/api/endpoints/cluster.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Update
+
+Update the configuration of a list of components.
+
+`POST /components.update`
```
chrismatix, line 12:
```suggestion
`POST /cluster.update`
```

infrastructure/controller/pkg/endpoints/components_get.go
```diff
@@ -17,7 +17,7 @@ type ClusterGet struct {
 }
 
 func (s *ClusterGet) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	components, err := configmaps.GetComponentsConfigMaps(r.Context(), s.namespace, s.clientSet, maskSecrets)
+	components, err := k8s.GetComponentsConfigMaps(r.Context(), s.namespace, s.clientSet, maskSecrets)
```
chrismatix, line 20: Does this actually work yet? 🤔  When I run the image in my local cluster this endpoint returns:

```json5
{"components":{}}
```

Or perhaps there is something more than add the image that I have to do :)

ljupcovangelski: > Does this actually work yet? thinking When I run the image in my local cluster this endpoint returns:
> 
> ```js
> {"components":{}}
> ```
> 
> Or perhaps there is something more than add the image that I have to do :)

You would also need to add the ingress rules.

chrismatix: This request is directly from within the cluster. If it was a networking issue the response would be a `404`. Will do more testing later

juan-sebastian (approved):
Perfect. great work!! Thanks for adding the label everywhere

PR #3177, opened 2022-05-18
[#3130] Feature/endpoint for retrieving the current configuration

### LINKED ISSUE #3130: Endpoint for retrieving the current configuration from the airy-controller
We need an endpoint on the `airy-controller` from where we can retrieve the current configuration from the `Airy Core` instance.

The following should be made possible on the `airy-controller` side:

```
curl -X GET -H "Content-Type: application/json" http://airy.core/controller/config
```

Then we need to implement this in the `Airy CLI`. When running `airy config` we should write the configuration to an `airy.yaml` file in the workspace.

Related to #2756

juan-sebastian:
closes #3130

### DIFF
--- a/infrastructure/controller/pkg/endpoints/cluster_get.go
+++ b/infrastructure/controller/pkg/endpoints/cluster_get.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"log"
 	"net/http"
+	"regexp"
 	"strings"
 
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -41,7 +42,7 @@ func (s *ClusterGet) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			components[componentsGroup] = componentsGroupContent
 		}
 
-		componentsGroupContent[componentName] = configmap.Data
+		componentsGroupContent[componentName] = maskSecrets(configmap.Data)
 	}
 
 	blob, err := json.Marshal(map[string]interface{}{"components": components})
@@ -63,4 +64,33 @@ func getComponentFromLabel(l string) (string, string, bool) {
 	}
 
 	return c[0], c[1], true
+}
+
+var secretMatcher = regexp.MustCompile(`(?i)secret|key|token`)
+
+func maskSecrets(data map[string]string) map[string]string {
+	mask := func(s string) string {
+		if len(s) < 2 {
+			return "..."
+		}
+
+		if len(s) > 8 {
+			return s[:4] + "..."
+		}
+
+		return s[:1] + "..."
+	}
+	out := make(map[string]string, len(data))
+
+	for k, v := range data {
+		if k == "saFile" {
+			out[k] = "<service account keys>"
+		} else if secretMatcher.MatchString(k) {
+			out[k] = mask(v)
+		} else {
+			out[k] = v
+		}
+	}
+
+	return out
 }
--- a/infrastructure/controller/pkg/endpoints/cluster_get.go
+++ b/infrastructure/controller/pkg/endpoints/cluster_get.go
@@ -1,0 +1,66 @@
+package endpoints
+
+import (
+	"encoding/json"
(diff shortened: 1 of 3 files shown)


ljupcovangelski (commented):
Maybe you can also add the CLI part that when we do

```
airy config or airy config get
```

that the configuration is printed on the CLI.

Default by yaml, maybe if we pass an option `airy config -o json` - it prints it in json.

infrastructure/controller/pkg/endpoints/cluster_get.go
```diff
@@ -0,0 +1,96 @@
...
+type ClusterGet struct {
+	clientSet *kubernetes.Clientset
+	namespace string
+}
+
+func (s *ClusterGet) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	configmapList, err := s.clientSet.CoreV1().ConfigMaps(s.namespace).List(r.Context(), v1.ListOptions{LabelSelector: "core.airy.co/component"})
+	if err != nil {
+		log.Printf("Unable to list config maps. Error: %s\n", err)
+		w.WriteHeader(http.StatusInternalServerError)
```
chrismatix, line 23: Do you think it'd be useful to return the error in a json response payload like we do with other endpoints?

infrastructure/controller/pkg/endpoints/cluster_get.go
```diff
@@ -0,0 +1,96 @@
...
+			componentsGroupContent = make(map[string]interface{})
+			components[componentsGroup] = componentsGroupContent
+		}
+
+		componentsGroupContent[componentName] = maskSecrets(configmap.Data)
+	}
+
+	blob, err := json.Marshal(map[string]interface{}{"components": components})
+	if err != nil {
+		log.Printf("Unable to marchal config Error: %s\n", err)
```
chrismatix, line 50:
```suggestion
		log.Printf("Unable to marshal config Error: %s\n", err)
```

infrastructure/controller/pkg/endpoints/server.go
```diff
@@ -38,5 +39,8 @@ func Serve(clientSet *kubernetes.Clientset, namespace string) {
 	s := &Services{clientSet: clientSet, namespace: namespace}
 	r.Handle("/services", s)
 
+	cg := &ClusterGet{clientSet: clientSet, namespace: namespace}
+	r.Handle("/cluster.get", cg).Methods("GET")
```
chrismatix, line 43: Like all our other endpoints this should be a `POST`

infrastructure/controller/pkg/endpoints/cluster_get.go
```diff
@@ -0,0 +1,96 @@
...
+	if len(c) != 2 {
+		return "", "", false
+	}
+
+	return c[0], c[1], true
+}
+
+var secretMatcher = regexp.MustCompile(`(?i)secret|key|token`)
+
+func maskSecrets(data map[string]string) map[string]string {
```
chrismatix, line 71: Very good thinking! Long term I think the services need to add some metadata so that we know which fields to mask without hardcoding it. What do you think?

chrismatix (commented):
First look. Most importantly docs are still missing

infrastructure/controller/pkg/endpoints/cluster_get.go
juan-sebastian: When it comes to internal server errors like kubectl errors. they should not be forwarded to the client

infrastructure/controller/pkg/endpoints/cluster_get.go
juan-sebastian: Yes this is what we were talking yesterday. we can do this in a second iteration

juan-sebastian:
The docs are going to made in https://github.com/airyhq/airy/issues/3130

ljupcovangelski (approved):
Awesome :+1:

PR #3203, opened 2022-05-24
[#3192] update contacts update

### LINKED ISSUE #3192: Display name update bug
**Current state:** 
at the moment, we have 2 display names for contacts: there is one `displayName` in `conversation.metadata` and one `displayName` in `contacts`

The problem is that these 2 infos can get out of sync: for example, when the user updates the display name on the UI, the display name in `conversation.metadata` gets updated but not the display name in `contacts`. 

We can fix this by updating both informations when the user updates the display name on the UI.

AudreyKj:
closes #3192

### DIFF
--- a/frontend/inbox/src/pages/Inbox/Messenger/ConversationMetadata/index.tsx
+++ b/frontend/inbox/src/pages/Inbox/Messenger/ConversationMetadata/index.tsx
@@ -2,7 +2,7 @@ import React, {FormEvent, useEffect, useState} from 'react';
 import {connect, ConnectedProps} from 'react-redux';
 import {Tag as TagModel, TagColor} from 'model';
 
-import {createTag, listTags} from '../../../../actions';
+import {createTag, listTags, updateContactDetails} from '../../../../actions';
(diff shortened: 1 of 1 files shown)


frontend/inbox/src/pages/Inbox/Messenger/ConversationMetadata/index.tsx
```diff
@@ -142,6 +144,7 @@ const ConversationMetadata = (props: ConnectedProps<typeof connector>) => {
 
   const saveEditDisplayName = () => {
     updateConversationContactInfo(conversation.id, displayName);
+    updateContactDetails(conversation.id, {id: contacts[conversation.id].id, displayName: displayName});
```
chrismatix, line 147: What if the contact is not in the store? 🤔

AudreyKj: the `ContactDetails ` component fetches the contact for each conversation and puts it in the store when it is rendered, so the contact will be in the store by the time the user updates the display name.

PR #3219, opened 2022-05-26
[#3171] Admin api unhealthy on minikube

### LINKED ISSUE #3171: Api-Admin unhealthy in local instance
## Describe the bug

Api-Admin appears as unhealthy (both in the frontend and in the `client.config` response) in the **local** (on staging it appears as healthy) instance for the release candidate 0.44.0. 

## To Reproduce

1.create an airy instance 
2. check the response of `client.config` 

```
kubectl exec -it kafka-0 -c kafka-broker -- curl api-admin/actuator/health
{"status":"UP","groups":["liveness","readiness"]}% 

```
Screenshot 2022-05-17 at 11 29 02

Screenshot 2022-05-17 at 11 28 36

chrismatix:
resolves #3171 

This is caused by a longstanding minikube [issue](https://github.com/kubernetes/minikube/issues/1568) and fixing it by adding `--cni=true` to the startup options would apparently incur a `~30%` performance cost.

### DIFF
--- a/infrastructure/helm-chart/charts/components/templates/api/admin/deployment.yaml
+++ b/infrastructure/helm-chart/charts/components/templates/api/admin/deployment.yaml
@@ -38,6 +38,10 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
+        - name: KUBERNETES_APP
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.labels['app']
         - name: KAFKA_BROKERS
           valueFrom:
             configMapKeyRef:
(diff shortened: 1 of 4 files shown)


ljupcovangelski (approved):
Nicely done. We probably need an issue to update our docs and introduce the term `app` as a building block of a component. A `component` can be comprised of multiple `apps`.

infrastructure/helm-chart/charts/components/templates/api/admin/deployment.yaml
```diff
@@ -38,6 +38,10 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
+        - name: KUBERNETES_APP
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.labels['app']
```
ljupcovangelski, line 44: Cool :+1:

PR #3221, opened 2022-05-27
[#3178] Feature/ enable disable componets from control center

### LINKED ISSUE #3178: Enabling and disabling - Components on Control Center
We need to allow System user to enable/disable components/features such as Contacts on Control Center.

Success story:

System user is on Control Center, Catalog view.
System user wants to enable a component on their Airy instance. e.g: Contacts, Inbox
System user clicks on "Enable".
The frontend should send such config. to CLI endpoint.
E.g.: "Contacts" is enabled on Airy core.
E.g.: Contacts appear on the left nav bar on control center to allow config.
E.g.: Contacts is Enabled on Inbox by "Contacts" view.
Control Center - DISABILING options:

Disable a "Service" from Status view.
Disable Connector from Connectors view.
Disable Component after installing and config from left nav. bar.
Control Center - ENABLING options:
1- Enable Components/connector from "Catalog" view,

juan-sebastian:
closes #3178


juan-sebastian:
I will do a separate PR for the front-end part

infrastructure/controller/pkg/endpoints/enable_disable_components.go
```diff
@@ -0,0 +1,109 @@
...
+			return
+		}
+
+		if c.Enable {
+			*deployment.Spec.Replicas = 1
+		} else {
+			*deployment.Spec.Replicas = 0
+		}
+		//TODO: Use Patch insdead
+		_, err := s.clientSet.AppsV1().Deployments(s.namespace).Update(
```
ljupcovangelski, line 62: Why are we updating the deployment from the endpoint? Maybe it will be a better approach if the endpoint only changes the configMap/component and then the controller, when he detects this change, reacts and applies the change.

infrastructure/controller/pkg/endpoints/server.go
```diff
@@ -51,5 +51,8 @@ func Serve(clientSet *kubernetes.Clientset, namespace string) {
 	clusterUpdate := &ClusterUpdate{clientSet: clientSet, namespace: namespace}
 	r.Handle("/cluster.update", clusterUpdate)
 
+	enableDisableComponents := &EnableDisableComponents{clientSet: clientSet, namespace: namespace}
+	r.Handle("/components.enable", enableDisableComponents).Methods("POST")
+
```
ljupcovangelski, line 56: As the `config.Component` struct already has the `Enabled` parameter, perhaps we can also use the `component.update` endpoint? And then annotate the appropriate configMap. In that case, we will not need a new endpoint which perhaps makes the system simpler.

ljupcovangelski (commented):
The upgrade of Go to 1.18 is awesome :tada: 
Left couple of questions on the feature.

infrastructure/controller/pkg/endpoints/enable_disable_components.go
```diff
@@ -0,0 +1,109 @@
...
+			w.WriteHeader(http.StatusInternalServerError)
+		}
+	}
+
+	configmaps, err := s.clientSet.CoreV1().ConfigMaps(s.namespace).List(
+		r.Context(),
+		v1.ListOptions{LabelSelector: fmt.Sprintf("core.airy.co/component=%s", c.Name)},
+	)
+	if err != nil {
+		klog.Errorf("not able to get config-map for component %s, %s", c.Name, err)
```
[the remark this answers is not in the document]
chrismatix: But these aren't internal errors and the endpoints are authenticated, i.e. the client is trusted to know this information. Without this clients cannot determine what went wrong and would always have to go and check the logs.

PR #3222, opened 2022-05-30
[#3187] upgrade to react 18

### LINKED ISSUE #3187: Upgrade to React 18
We are currently using react v17; to do: upgrade to v18. 

**changes & breaking changes:** https://github.com/facebook/react/releases/tag/v18.0.0

AudreyKj:
closes #3187


frontend/control-center/src/pages/Connectors/ConnectedChannelsList/DisableModal/index.tsx
```diff
@@ -51,8 +51,10 @@ export const DisableModal = (props: DisableModalProps) => {
             }}
             type="submit"
           >
-            {isLoading && <RefreshIcon height={24} width={24} />}
-            {isLoading ? t('unsubscribing') : error ? t('tryAgain') : t('confirm')}
+            <>
```
chrismatix, line 54: Why is this necessary? 🤔

AudreyKj: we need the fragment to wrap several children

chrismatix: Locally I can build this without the fragment. Do you get an error?

frontend/control-center/src/pages/Connectors/Providers/Facebook/Messenger/FacebookConnect.tsx
```diff
@@ -28,7 +28,7 @@ const FacebookConnect = (props: ConnectedProps<typeof connector>) => {
   const [token, setToken] = useState(channel?.metadata?.pageToken || '');
   const [name, setName] = useState(channel?.metadata?.name || '');
   const [image, setImage] = useState(channel?.metadata?.imageUrl || '');
-  const [buttonTitle, setButtonTitle] = useState(t('connectPage'));
+  const [buttonTitle, setButtonTitle] = useState(t<string>('connectPage'));
```
chrismatix, line 31: So these type annotations are only necessary when we do
```typescript
import {t} from 'i18next';
```

rather than
```typescript
import {useTranslation} from 'i18next';

///...
const {t} = useTranslation();
```

Not sure why and how the first way works, but the second one looks more correct according to the docs so let's use that one

frontend/control-center/src/pages/Connectors/ConnectedChannelsList/DisableModal/index.tsx
AudreyKj: You're right! I thought I had an error about this, but I actually tested it now, and it works without the fragment. Will fix now 👍

AitorAlgorta (approved):
LGTM 🚀

PR #3251, opened 2022-06-03
[#3247] fixed catalog view card

### LINKED ISSUE #3247: Installed - Catalog view card
On Control Center, when user selects catalog view and open an installed component, the top left back button contains "Channels" as context which is wrong. 

FIX

This back button should contain contain wither "Catalog" as the back action brings the user to Catalog main view, or contains no text at all.  

EXAMPLE of how it looks currently on Control center: 

Steps to reproduce: 

1. Select catalog view. 
2. Click on any installed component.
3. Opens the channels view related to this component with "Channels" back button at the top left corner.

AudreyKj:
closes #3247 

- fixed Catalog view card 
- added missing translations


thorstenairy (approved):
🚀

PR #3322, opened 2022-06-21
[#3279] Minikube installation docs clarification

### LINKED ISSUE #3279: Improvements to make installation docs. clearer and easier to follow
Notes/comments on ambiguities found when installing airy using the different options.

- [x] Ngrok URL might be confused with the API/UI URL. Add a sentence to make clear that the API URL is also the UI one.
- [x] Change Minikube install link from German one to https://kubernetes.io/docs/tasks/tools/.
- [x] Insert note `:::note` syntax to remind users if it fails, to check the permission on Minikube.

armanjindal:
Resolves #3279


ljupcovangelski (approved):
Nice :+1:

PR #3323, opened 2022-06-21
[#3096] Allow connecting to a remote Kafka cluster

### LINKED ISSUE #3096: Authentication to a remote Kafka cluster
Currently Airy Core runs on a local Kafka cluster, deployed in the same Kubernetes cluster. However someone might want to use Airy Core with a remote Kafka cluster and access it via the Internet using an authentication method.

We need to change our backend apps that they support connecting to a remote Kafka cluster using authentication and update the docs.

chrismatix:
resolves #3096 

- Cleans up Kafka variables in the `application.properties`
- Adds jaas-based Kafka authentication
- Allows for applying Kafka configuration using `airy config apply`
- Helm chart upgrade tested on minikube env ✅

### DIFF
--- a/infrastructure/controller/pkg/endpoints/cluster_update.go
+++ b/infrastructure/controller/pkg/endpoints/cluster_update.go
@@ -40,7 +40,7 @@ func (s *ClusterUpdate) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		err := k8s.ApplyConfigMap("security", s.namespace, secData, map[string]string{}, map[string]string{}, s.clientSet, r.Context())
 		if err != nil {
 			klog.Error("unable to apply configuration for \"security\"\nError:\n" + err.Error())
-			http.Error(w, err.Error(), http.StatusInternalServerError)
+			w.WriteHeader(http.StatusInternalServerError)
 			return
 		}
 		response.ClusterConfig["security"] = true
@@ -51,7 +51,7 @@ func (s *ClusterUpdate) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		applyErr := k8s.ApplyConfigMap("kafka-config", s.namespace, kafkaData, map[string]string{}, map[string]string{}, s.clientSet, r.Context())
 		if applyErr != nil {
 			klog.Error("unable to apply configuration for \"kafka\"\nError:\n" + err.Error())
-			http.Error(w, err.Error(), http.StatusInternalServerError)
+			w.WriteHeader(http.StatusInternalServerError)
 			return
 		}
 		response.ClusterConfig["kafka"] = true


docs/docs/guides/remote-kafka-cluster.md
```diff
@@ -0,0 +1,34 @@
...
+This should leave you with a single line JAAS configuration string that you also have to add to your `airy.yaml` file:
+
+```yaml
+kafka:
+  bootstrapServers: kafka-broker-1.example.com:9092,kafka-broker-2.example.com:9092
+  authJaas: "org.apache.kafka.common.security.plain.PlainLoginModule required username=\"Grace\" password=\"trustno1\";"
+```
+
+Now you can apply this configuration by running the `airy config apply` command. Once you have verified that everything is
+working you can scale down the default Kafka cluster by running the `kubectl scale sts kafka --replicas 0` command.
```
ljupcovangelski, line 34: Perhaps will be good to add a reference on how to run Airy without the pre-installed Kafka, using Helm - https://airy.co/docs/core/getting-started/installation/helm#kafka

ljupcovangelski (approved):
Nice :tada:

lib/java/spring/kafka/streams/src/main/resources/kafka-streams.properties
```diff
@@ -0,0 +1,6 @@
+kafka.brokers=${KAFKA_BROKERS}
+kafka.schema-registry-url=${KAFKA_SCHEMA_REGISTRY_URL}
+kafka.commit-interval-ms=${KAFKA_COMMIT_INTERVAL_MS:1000}
+
+# Only used for testing:
+kafka.cleanup=${KAFKA_CLEANUP:false}
```
juan-sebastian, line 6: How does this works for testing? There are other properties like.
```
kafka.commit-interval-ms=0
kafka.cache.max.bytes=0
kafka.delete.retention.ms=0
```
There are still taken into account?

backend/api/admin/src/main/resources/application.properties
```diff
@@ -1,6 +1,2 @@
-kafka.brokers=${KAFKA_BROKERS}
```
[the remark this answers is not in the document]
juan-sebastian: How this will work for `cloud` apps?

infrastructure/controller/pkg/endpoints/cluster_update.go
```diff
@@ -37,15 +37,26 @@ func (s *ClusterUpdate) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 		}
 		response.ClusterConfig["security"] = true
 	}
 
+	kafkaData := config.GetKafkaData(conf.ClusterConfig.Kafka)
+	if len(kafkaData) != 0 {
+		applyErr := k8s.ApplyConfigMap("kafka-config", s.namespace, kafkaData, map[string]string{}, map[string]string{}, s.clientSet)
+		if applyErr != nil {
+			klog.Error("unable to apply configuration for \"kafka\"\nError:\n" + err.Error())
+			http.Error(w, err.Error(), http.StatusInternalServerError)
```
juan-sebastian, line 54: We should not return the reason of a `500` error to the client

lib/go/k8s/configmaps.go
```diff
@@ -29,28 +29,28 @@ func ApplyConfigMap(
...
+	for k, v := range labels {
+		cm.Labels[k] = v
+	}
+	if cm.Annotations == nil {
+		cm.Annotations = make(map[string]string)
+	}
+	for k, v := range annotations {
+		cm.Annotations[k] = v
+	}
+	_, err := clientset.CoreV1().ConfigMaps(namespace).Update(context.TODO(), cm, v1.UpdateOptions{})
```
juan-sebastian, line 52: Here you should use the correct `context`, Maybe the one coming from the http request or the one for k8s

backend/api/admin/src/main/resources/application.properties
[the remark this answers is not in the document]
chrismatix: Nothing has to be changed but in the future these configs can also be removed for them.

lib/java/spring/kafka/streams/src/main/resources/kafka-streams.properties
chrismatix: Yes they are still taken into account. The only thing that changed is that these values are now mapped in a single place. Test properties are still a hard override.

juan-sebastian (approved):

PR #3335, opened 2022-06-22
[#3332] Authenticate AWS using profile

### LINKED ISSUE #3332: Duplicate authentication credentials for AWS in Terraform
The AWS access key and AWS secret key are passed into the `main.tf` file to authenticate, even though they are already set in the `aws` CLI.

armanjindal:
Resolves #3332 by authenticating AWS using the profile instead of credentials in the `main.tf` file.


ljupcovangelski (changes requested):
I think it will be better to unify the variables and name `region` also `aws_region`.

ljupcovangelski (approved):
Looks good now

PR #3351, opened 2022-06-27
[#3334] Terraform install in two environments: AWS-EKS

### LINKED ISSUE #3334: Two Terraform environments for a single Airy Core Install
To install `Airy Core` there are two separate Terraform environments `kubernetes` and `airy-core`. 

This makes the installation more complicated as `terraform init` and `terraform apply` have to be called first in the `kuberenetes` directory and then the in `airy-core` directory. Destroy is also complicated as the reverse order has to be followed. 

When installing Airy Core with the CLI there can only be one deployment per workspace, so to have two TF environments to manage a single deployment unnecessarily complicates the management of the Airy deployment.

armanjindal:
Resolves #3334 

How to test:
```
cd infrastructure/terraform/install

cat <<EOF >test.flags
PROVIDER=aws-eks
WORKSPACE="../"
EOF

./install.sh
kubectl --kubeconfig ./kube.conf get pods
./uninstall.sh
```


infrastructure/terraform/install/airy-core/variables.tf
```diff
@@ -0,0 +1,17 @@
...
+  default = ""
+}
+
+variable "hosted_zone" {
+  default = "airy.co"
+}
+
+variable "airy_workspace" {
+  type    = string
+  default = "../../"
```
ljupcovangelski, line 16:
```suggestion
  default = "../.."
```

ljupcovangelski (commented):
Looks good. It is good that the install.sh/uninstall.sh scripts are really simple, I'd just add:
- checks if the exit code from apply/destroy terraform is `0` before continuing to the other in the sequence.
- check if the AWS_PROFILE and AWS_REGION are already set and work on that
- add an argument with default value for setting the provider.

ljupcovangelski (approved):
Let's go :tada:

PR #3384, opened 2022-07-05
[#3359] Add method to check environment pre-install

### LINKED ISSUE #3359: No check for prerequisites for Terraform AWS-EKS install  in CLI
As we modify the CLI command `airy create --provider=aws` to use the AWS-EKS Terraform install, instead of the current hard-coded version, we must first check the necessary binaries are installed on the installer's computer.

- [x] `aws` CLI tool  is installed  
- [x] `terraform` is installed

armanjindal:
Resolves #3359 and also adds a new method to the `provider` interface called `IsEnvironmentValid()`. This function allows us to check that the installer's local environment has the necessary binaries and configurations we expect to make the install run. In this case we are checking for `aws` and `terraform` (but not checking their version)

### DIFF
--- a/cli/pkg/cmd/create/create.go
+++ b/cli/pkg/cmd/create/create.go
@@ -65,9 +65,15 @@ func create(cmd *cobra.Command, args []string) {
 	overrides.Version = version
 	overrides.Namespace = namespace
 	overrides.TrackingDisabled = disableTracking
+	isValidEnv, err := provider.IsEnvironmentValid()
+
+	if !isValidEnv {
+		console.Exit("Please check if the install requirements are set up", err)
+	}
+
 	dir, err := workspace.Create(workspacePath, overrides)
 	if err != nil {
-		console.Exit("could not initialize Airy workspace directory", err)
+		console.Exit("Could not initialize Airy workspace directory", err)
 	}
 	fmt.Println("📁 Initialized Airy workspace directory at", dir.GetPath("."))
 	if initOnly {
--- a/cli/pkg/cmd/create/create.go
+++ b/cli/pkg/cmd/create/create.go
@@ -68,12 +68,12 @@ func create(cmd *cobra.Command, args []string) {
 	isValidEnv, err := provider.IsEnvironmentValid()
 
 	if !isValidEnv {
-		console.Exit("Please check if the install requirements are set up", err)
+		console.Exit("please check if the install requirements are set up", err)
 	}
 
 	dir, err := workspace.Create(workspacePath, overrides)
 	if err != nil {
-		console.Exit("Could not initialize Airy workspace directory", err)
+		console.Exit("could not initialize Airy workspace directory", err)
 	}
 	fmt.Println("📁 Initialized Airy workspace directory at", dir.GetPath("."))
 	if initOnly {
--- a/cli/pkg/providers/aws/aws.go
+++ b/cli/pkg/providers/aws/aws.go
@@ -52,16 +52,15 @@ func (p *provider) GetOverrides() tmpl.Variables {
 	}
 }
 
-func (p *provider) CheckEnvironment() (bool, string) {
+func (p *provider) IsEnvironmentValid() (bool, error) {
 	necessaryBinaries := [2]string{"aws", "terraform"}
 	for _, binary := range necessaryBinaries {
 		_, err := exec.LookPath(binary)
 		if err != nil {
-			errMsg := fmt.Sprintf("Please check %s\nError - %s", binary, err)
-			return false, errMsg
+			return false, err
 		}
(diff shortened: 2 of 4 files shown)


cli/pkg/providers/aws/aws.go
```diff
@@ -51,6 +52,18 @@ func (p *provider) GetOverrides() tmpl.Variables {
 	}
 }
 
+func (p *provider) CheckEnvironment() (bool, string) {
```
chrismatix, line 55: nitpicks:
- It's more idiomatic in go to return an `err` than a string. Also may as well return the error directly and let the function caller add the "Please check..." string.
- `IsEnvironmentValid` seems a bit better since this method returns a boolean

cli/pkg/providers/provider.go
```diff
@@ -21,6 +21,7 @@ const (
 type Provider interface {
 	Provision(providerConfig map[string]string, dir workspace.ConfigDir) (kube.KubeCtx, error)
 	GetOverrides() template.Variables
+	IsEnvironmentValid() (bool, error)
```
ljupcovangelski, line 24: Perhaps it is better to return just `error`. As if `err := nil`, means that the check failed.

cli/pkg/cmd/create/create.go
```diff
@@ -65,9 +65,15 @@ func create(cmd *cobra.Command, args []string) {
...
+	isValidEnv, err := provider.IsEnvironmentValid()
+
+	if !isValidEnv {
+		console.Exit("Please check if the install requirements are set up", err)
+	}
+
 	dir, err := workspace.Create(workspacePath, overrides)
 	if err != nil {
-		console.Exit("could not initialize Airy workspace directory", err)
+		console.Exit("Could not initialize Airy workspace directory", err)
```
ljupcovangelski, line 76: These messages are deliberately outputted starting with small letter, I recommend to return it as it was. Also for the previous `console.Exit`.

cli/pkg/providers/aws/aws.go
```diff
@@ -51,6 +52,17 @@ func (p *provider) GetOverrides() tmpl.Variables {
 	}
 }
 
+func (p *provider) IsEnvironmentValid() (bool, error) {
+	necessaryBinaries := [2]string{"aws", "terraform"}
```
ljupcovangelski, line 56: Maybe it is better to define these outside of the function in a const. That way the function becomes re-usable.

ljupcovangelski (commented):
Merging this PR will break the AWS provider, as `airy create --provider aws` will exit if terraform is not installed (and it is not needed at the moment).
The concept is good, but I recommend adding the switch to Terraform in the same PR.

cli/pkg/providers/aws/aws.go
armanjindal: I see. So we do not make it provider specific. I was thinking if ever need to check specific things such as environment variables etc. this could be helpful. But then this function would be focused on only checking binaries (which is highly reusable)

cli/pkg/providers/aws/aws.go
```diff
@@ -51,6 +52,17 @@ func (p *provider) GetOverrides() tmpl.Variables {
...
+	for _, binary := range necessaryBinaries {
```
juan-sebastian, line 57: You could just do
```go
for _, binary := range []string{"aws", "terraform"} {
 ...
}
```

cli/pkg/providers/aws/aws.go
```diff
@@ -51,6 +52,17 @@ func (p *provider) GetOverrides() tmpl.Variables {
 	}
 }
 
+func (p *provider) IsEnvironmentValid() (bool, error) {
```
juan-sebastian, line 55: You can just return `error`.

```go
if err != nil { ... }
```

If error is not nil this means that the environment is not valid, the `bool` value is redundant

cli/pkg/providers/minikube/minikube.go
```diff
@@ -45,6 +45,10 @@ func (p *provider) GetOverrides() template.Variables {
 	}
 }
 
+func (p *provider) IsEnvironmentValid() (bool, error) {
+	return true, nil
+}
+
```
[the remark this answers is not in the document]
armanjindal: Since I added this function to the `provider interface` it needs to be implemented for every provider. This makes sense as each install with this new TF approach will require that we first check binaries are on the local machine as well as certain environment variables. I am closing this PR and will open up a new one on the umbrella issue with these changes from your comments. Thanks!

armanjindal:
This issue will be fixed on the larger feature branch `feat/3397-terraform-install-cli`, where this issue is a dependency for #3397

PR #3389, opened 2022-07-06
Feature/3209 install uninstall components

### LINKED ISSUE #3209: Create endpoints to install components
Create an endpoint that receives a request to install a component and then the controller would deploy the appropriate helm chart for it

juan-sebastian:
closes #3209

### DIFF
--- a/infrastructure/controller/pkg/endpoints/components_install_uninstall.go
+++ b/infrastructure/controller/pkg/endpoints/components_install_uninstall.go
@@ -1,6 +1,7 @@
 package endpoints
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
@@ -11,16 +12,24 @@ import (
 	"github.com/airyhq/airy/lib/go/payloads"
 	helmCli "github.com/mittwald/go-helm-client"
 	"helm.sh/helm/v3/pkg/repo"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/klog"
 )
 
 type ComponentsInstallUninstall struct {
 	cli       helmCli.Client
+	clientSet *kubernetes.Clientset
 	namespace string
 }
 
-func MustNewComponentsInstallUninstall(namespace string, kubeConfig *rest.Config, reposFilePath string) ComponentsInstallUninstall {
+func MustNewComponentsInstallUninstall(
+	namespace string,
+	kubeConfig *rest.Config,
+	clientSet *kubernetes.Clientset,
+	reposFilePath string,
+) ComponentsInstallUninstall {
 	cli, err := helmCli.NewClientFromRestConf(&helmCli.RestConfClientOptions{
 		Options: &helmCli.Options{
 			Namespace: namespace,
@@ -47,7 +56,7 @@ func MustNewComponentsInstallUninstall(namespace string, kubeConfig *rest.Config
 		klog.Info("Added ", r.Name, " repo")
 	}
 
-	return ComponentsInstallUninstall{namespace: namespace, cli: cli}
+	return ComponentsInstallUninstall{namespace: namespace, cli: cli, clientSet: clientSet}
 }
 
 func (s *ComponentsInstallUninstall) ServeHTTP(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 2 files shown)


infrastructure/controller/pkg/endpoints/components_install_uninstall.go
```diff
@@ -0,0 +1,130 @@
...
+	//NOTE: For now we are setting the repos here. But in the future we could get them from an external configuration
+	chartRepoCore := repo.Entry{
+		Name: "airy-core",
+		URL:  "https://helm.airy.co",
+	}
+	if err := cli.AddOrUpdateChartRepo(chartRepoCore); err != nil {
+		log.Fatal(err)
+	}
+
+	//NOTE: We Look for user and password for airy-enterprise. If found we add it to our helm client.
```
chrismatix, line 44: nitpick: We shouldn't add code for things that we haven't decided yet on. Until we do this code is dead weight.

docs/docs/api/endpoints/components.md
```diff
@@ -156,19 +156,11 @@ Install a new component.
 
 ```json
 {
-  "name": "sources-chat-plugin"
+  "name": "airy-core/sources-chatplugin"
```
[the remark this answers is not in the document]
juan-sebastian: There is no need to hardcoded. Actually it was quite easy to integrate this so why I did it

infrastructure/controller/pkg/endpoints/components_install_uninstall.go
```diff
@@ -0,0 +1,130 @@
...
+
+	chartSpec := &helmCli.ChartSpec{
+		ReleaseName: releaseName,
+		ChartName:   chartName,
+		Namespace:   s.namespace,
+		UpgradeCRDs: true,
+		Replace:     true,
+	}
+
+	if r.URL.Path == "/components.install" {
```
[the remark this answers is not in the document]
juan-sebastian: There will be a lot of receptive code. All the code will be the same except for one function. This is why I decided to go for this route

infrastructure/controller/pkg/endpoints/components_install_uninstall.go
juan-sebastian: Yes I moved this to a config-map

docs/docs/api/endpoints/components.md
```diff
@@ -180,16 +172,12 @@ Install an existing component.
 
 ```json
 {
-  "name": "sources-chat-plugin"
+  "name": "airy-enterprise/dialogflow-connector"
```
ljupcovangelski, line 175:
```suggestion
  "name": "airy-core/sources-chat-plugin"
```

docs/docs/api/endpoints/components.md
```diff
@@ -156,19 +156,11 @@ Install a new component.
...
-The response includes the name of the component that was installed with an indication of the process was completed successfully (true) or not (false).
-
-```json5
-{
-  "sources-chat-plugin": true
-}
-```
+**(201) Success Response Payload**
 
 ## Uninstall
```
ljupcovangelski, line 166: *Uninstall an existing component*.

infrastructure/controller/pkg/endpoints/components_install_uninstall.go
```diff
@@ -0,0 +1,133 @@
...
+		_, err := s.cli.InstallOrUpgradeChart(
```
ljupcovangelski, line 78: `install+replace` will not work if the chart is already there (or at least it is not stable). 
It results with:
```
Error: cannot re-use a name that is still in use
```

I suggest having `install` only now at the moment, and in the future we can introduce a `components.upgrade` endpoint that does `helm upgrade`.

infrastructure/controller/pkg/endpoints/components_install_uninstall.go
```diff
@@ -0,0 +1,133 @@
...
+		fmt.Fprintf(w, `{"error": "invalid component name"}`)
+		return
+	}
+
+	chartSpec := &helmCli.ChartSpec{
+		ReleaseName: releaseName,
+		ChartName:   chartName,
+		Namespace:   s.namespace,
+		UpgradeCRDs: true,
+		Replace:     true,
```
ljupcovangelski, line 74: I guess this is only needed because the job is `async` and we should be able to install the same chart multiple times. Otherwise, I think it is better to error if the chart is already installed.

infrastructure/helm-chart/templates/controller/configmap.yaml
```diff
@@ -9,7 +9,7 @@ data:
       "repositories": [
         {
           "name": "airy-core",
-          "url": "https://github.com/airyhq/airy/tree/develop/infrastructure/helm-chart/charts/components/charts"
+          "url": "https://helm.airy.co"
```
ljupcovangelski, line 12: What if the repository requires username/password? Should we also document that and is it considered in the controller code?

ljupcovangelski (commented):
When installing/upgrading the helm chart, also the `global` values need to be considered, as they are influencing mostly the ingress.

The global values can be found in
```
kubectl get configmap core-config -o jsonpath='{.data.global\.yaml}'
```

So the real concept should be 
```
helm install component airy-core/component --values global.yaml
```

docs/docs/api/endpoints/components.md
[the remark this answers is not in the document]
ljupcovangelski: In my opinion this is better as the client must know from which repository it is installing the component. The flow is: 1) What are the repositories, 2) What are the components in these repositories, 3) Install component from repository.

PR #3481, opened 2022-07-25
[#3465] status view enable button error when component is not configured

### LINKED ISSUE #3459: Salesforce !config hide "Enable" button
On Control Center, Connectors view to configure a new "Salesforce" installed connector. 

While the user is firstly configuring the connector, they shouldn't be able to see "Enable" button. Also this button works without checking for configuration. If you click on "Enable" button, it will automatically enable the component without configuration. 

HOW IT LOOKS CURRENTLY WHEN A USER IS CONFIGURING THE CONNECTOR FOR THE FIRST TIME 

1.

### LINKED ISSUE #3460: Dialogflow configuration disappears after update
On Control Center, Connectors view > "Dialogflow" configuration 

- When updating one of the fields and confirm restart, all variables are deleted. 
- Also the component status is enabled not "To configure" in case if the configuration has been deleted. 

1.

### LINKED ISSUE #3465: Status view - "Enable" button error when component is not configured
On Control Center, on Status view, as shown below: 

1. "Enterprise Push Notification" component is not configured BUT the enable button is on
2.  "Enterprise Push Notification" component is supposed to be "Not configured"BUT it has a service added which IS RUNNING. 
3. Viber component is not configured BUT the enable button is on. 
4. Viber component is supposed to be "Not configured"BUT it has a service added which has an issue. 

HOW CURRENTLY IT LOOKS LIKE ON STAGING: 

FIX: 

1. If the component has "Not configured" status, it means 
    1.1. No services/bots have been added on this component. 
    1.2. Enabled button should be hidden.

AudreyKj:
closes #3465 
closes #3459
closes #3460

### DIFF
--- a/frontend/control-center/src/actions/catalog/index.ts
+++ b/frontend/control-center/src/actions/catalog/index.ts
@@ -32,16 +32,16 @@ export const listComponents = () => (dispatch: Dispatch<any>) => {
 
 export const installComponent =
   (installComponentRequestPayload: InstallUninstallComponentRequestPayload) => (dispatch: Dispatch<any>) => {
-    dispatch(installComponentAction(installComponentRequestPayload));
     return HttpClientInstance.installComponent(installComponentRequestPayload).then(() => {
+      dispatch(installComponentAction(installComponentRequestPayload));
       return Promise.resolve(true);
     });
   };
 
 export const uninstallComponent =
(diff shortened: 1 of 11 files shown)


thorstenairy (approved):
🚀

PR #3482, opened 2022-07-26
Feature/3353 components.list endpoint version 2

### LINKED ISSUE #3353: /components.list endpoint to retrieve component details
For each component on Control Center, we need to retrieve specific detailed data about the component details such as: 

- Component Name.
- Component Description.
- Status of component "NOT INSTALLED"/Installed, as a tag.
- Available for : "Open-source" / "Enterprise" / "Cloud"
- Categories : "Communication Source" / "API" / "UI" / "Machine learning & Conversational AI" / "Analytics" / "Storage" / 9. "Customer Service" / "Marketing".
- Price : Free

Docs for content: https://docs.google.com/presentation/d/1ecTHtSDABuoKZMDhSw5GWycXFKt7VRMnu5jlzOtfnQQ/edit#slide=id.p 

I WILL ALLOW ACCESS AFTER REVIEWING THE CONTENT ;)

juan-sebastian:
closes #3353


infrastructure/controller/pkg/db/dynamodb.go
```diff
@@ -0,0 +1,75 @@
...
+
+//NOTE: we can change this later for an interface to accept multiples
+//      providers
+
+type DB struct {
+	svc *dynamodb.DynamoDB
+}
+
+func MustNewDB() DB {
+	accessKey, keyOk := os.LookupEnv("DYNAMODB_ACCESS_KEY_ID")
```
chrismatix, line 21: So this implies that we need to ship the dynamodb credentials with every open source installation, right? This must be very well documented as those values will always be in the `airy.yaml`.

A question that remains is also how we can stop people from abusing these access keys?

One solution would be to build a small public endpoint on cloud that we call from here.

infrastructure/controller/pkg/cache/deployed_charts.go
```diff
@@ -0,0 +1,66 @@
...
+	"time"
+
+	helmCli "github.com/mittwald/go-helm-client"
+	"k8s.io/klog"
+)
+
+type DeployedCharts struct {
+	deployedCharts map[string]bool
+	cli            helmCli.Client
+	mu             sync.RWMutex
```
chrismatix, line 14: nice 👏

infrastructure/controller/pkg/db/dynamodb.go
juan-sebastian: Yes this credentials must be shipped with every instance in orders to have access to the DB.

This credentials only have access to DynamoDB in read-only. Not sure that they can be abused.

PR #3531, opened 2022-08-02
[#2755] Add an "airy login" command to the CLI

### LINKED ISSUE #2755: Authentication to the Airy controller

chrismatix:
resolves #2755

### DIFF
--- a/cli/integration/golden/cli.no-args.golden
+++ b/cli/integration/golden/cli.no-args.golden
@@ -9,6 +9,7 @@ Available Commands:
   config      Manages an Airy Core instance via airy.yaml
   create      Creates an instance of Airy Core
   help        Help about any command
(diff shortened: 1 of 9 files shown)


juan-sebastian (dismissed):

juan-sebastian (commented):
Looks good but the build still fails

ljupcovangelski (approved):

Issue #2757, opened 2022-01-21
Remove helm jobs from the CLI


ljupcovangelski:
By decoupling the Airy CLI from Kubernetes, we cannot directly run `helm install` or `helm upgrade`.

We should run these jobs from the `airy-controller` or use the `helm` binary directly.

Depends on #2754

ljupcovangelski:
Decoupling the Airy CLI from Kubernetes means that the process of installing Airy are separated from applying the configuration and interacting with the API.
For installing the Airy helm chart, access to Kubernetes is mandatory (otherwise the Kubernetes resources cannot be deployed). Therefore, I propose that we also decouple this process from the CLI altogether. I propose that `airy create` only initializes the workspace and `prints out next steps which the user should execute on his own`, in order to install Airy Core on a new or on an existing Kubernetes cluster.

`airy create --provider minikube`:
- Checks if Minikube is installed
- Checks if helm is installed
- Checks if port 80 is open for the ingress congtroller (we can make it optional in the future for the user to selct his port)
- Creates the airy.yaml file with the recommended settings for Minikube.
- Prints out the command for Creating the Minikube cluster [^1].
- Prints the command for deploying the helm chart [^2].

`airy create --provider aws`:
- Checks if terraform is installed
- Creates the Terraform files in the workspace directory
- Creates the airy.yaml file with the recommended settings for AWS.
- Prints a note to add the credentials or profile to `airy.tfvars`
- Prints out the command for Creating an AWS EKS cluster with Airy[^3].

`airy create --provider generic` (or generic/kubernetes/existing, this should be the default option):
- Check if there is an existing access to Kubernetes and prints the cluster where it will be installed
- Checks if helm is installed
- Creates the airy.yaml file with the recommended generic settings for cloud.
- Prints the command for deploying the helm chart [^2].


[^1]: Command for creating a Minikue cluster:
minikube --profile=airy-core start --extra-config=apiserver.service-node-port-range=1-65535 --ports=80:80 --driver=docker  --container-runtime=containerd --cpus=4 --memory=7168

[^2]: Commands for deploying Airy Core with Helm:
helm repo add airy https://helm.airy.co
helm install airy airy/airy --namespace default --values ./airy.yaml

[^3]: Commands to create Airy with Terraform:
terraform init
terraform apply

AitorAlgorta:
I think this is a good solution but I would like to hear what the product team has to say. 
This looks like a question for them.

chrismatix:
[...]

PR #3555, opened 2022-08-08
[#2757] Use Terraform for Minikube provider

### LINKED ISSUE #2754: Run initial helm job from the CLI
As we want to decouple the Airy CLI from kubernetes, we need to have a way to run an initial helm job or deploy at least the `airy-controller`, so that we can interact later with it to apply the configuration and deploy additional helm jobs.

One way is to Deploy/Install Airy only with Helm and then after we install it, we use the CLI to interact with the cluster, only through the API.

The value that is in `airy create` is for quickly creating a Kubernetes cluster and installing Airy for quickstart and testing purposes. So we can still decide to maintain the `airy create` sub-command and migrate the code to Terraform (for now AWS or we can add multiple providers in the future).

ljupcovangelski:
Use Terraform for Minikube as well:
- Minikube module
- Write the `install.flags` from the CLi
- Run the `install.sh` script when running `airy create --provider minikube`

Resolves #2754.

### DIFF
--- a/cli/pkg/cmd/create/create.go
+++ b/cli/pkg/cmd/create/create.go
@@ -2,18 +2,14 @@ package create
 
 import (
 	"cli/pkg/console"
-	"cli/pkg/helm"
 	"cli/pkg/providers"
 	"cli/pkg/workspace"
 	"fmt"
 	"os"
 	"runtime"
 
-	"github.com/airyhq/airy/lib/go/k8s"
-
 	"github.com/TwinProduction/go-color"
 	"github.com/spf13/cobra"
-	"github.com/spf13/viper"
 	"gopkg.in/segmentio/analytics-go.v3"
 )
 
@@ -88,69 +84,9 @@ func create(cmd *cobra.Command, args []string) {
 	fmt.Fprintln(w)
 	fmt.Fprintln(w, providerName, "provider output:")
 	fmt.Fprintln(w)
-	context, err := provider.Provision(providerConfig, workspace.ConfigDir{Path: installDir})
+	err = provider.Provision(providerConfig, workspace.ConfigDir{Path: installDir})
 	fmt.Fprintln(w)
 	if err != nil {
 		console.Exit("could not install Airy: ", err)
-	}
-	if providerName == "minikube" { // TEMP fix to keep minikube working
-		clientset, err := context.GetClientSet()
-		if err != nil {
-			console.Exit("could not get clientset: ", err)
-		}
-
-		if err = context.Store(); err != nil {
-			console.Exit("could not store the kube context: ", err)
-		}
-
-		helm := helm.New(clientset, version, namespace, dir.GetAiryYaml())
-		if err := helm.Setup(); err != nil {
-			console.Exit("setting up Helm failed with err: ", err)
-		}
-
-		fmt.Println("🚀 Starting core with default components")
-
-		if err := helm.InstallCharts(); err != nil {
-			console.Exit("installing Helm charts failed with err: ", err)
-		}
-
-		if err = provider.PostInstallation(providerConfig, namespace, dir); err != nil {
-			console.Exit("failed to run post installation hook: ", err)
-		}
-
-		fmt.Println("🎉 Your Airy Core is ready")
-
-		coreConfig, err := k8s.GetCmData("core-config", namespace, clientset)
-		if err != nil {
-			console.Exit("failed to get hosts from installation")
-		}
-
-		fmt.Println("\t 👩‍🍳 Available hosts:")
-		for hostName, host := range coreConfig {
-			switch hostName {
-			case "HOST":
-				fmt.Printf("\t\t %s:\t %s", "Host", host)
-				fmt.Println()
-			case "API_HOST":
-				fmt.Printf("\t\t %s:\t %s", "API", host)
-				fmt.Println()
-			case "NGROK":
-				fmt.Printf("\t\t %s:\t %s", "NGROK", host)
-				fmt.Println()
-			}
-		}
-
-		fmt.Println("✅ Airy Installed")
-		fmt.Println()
-
-		viper.Set("provider", provider)
-		viper.Set("namespace", namespace)
-		viper.WriteConfig()
-
-		airyAnalytics.Track(analytics.Track{
-			UserId: coreConfig["CORE_ID"],
-			Event:  "installation_succesful"})
-		fmt.Printf("📚 For more information about the %s provider visit https://airy.co/docs/core/getting-started/installation/%s", providerName, providerName)
-		fmt.Println()
 	}
 }
--- a/cli/pkg/providers/minikube/minikube.go
+++ b/cli/pkg/providers/minikube/minikube.go
@@ -5,17 +5,15 @@ import (
 	"cli/pkg/kube"
 	"cli/pkg/workspace"
 	"cli/pkg/workspace/template"
-	"context"
-	"fmt"
 	"io"
+	"os"
 	"os/exec"
-	"path/filepath"
-	"runtime"
 	"strings"
 
+	"github.com/airyhq/airy/lib/go/tools"
+
+	"github.com/hashicorp/go-getter"
 	"gopkg.in/segmentio/analytics-go.v3"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/util/homedir"
 )
 
 const (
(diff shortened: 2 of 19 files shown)


infrastructure/terraform/install/minikube/main.tf
```diff
@@ -0,0 +1,6 @@
+module "minikube" {
+  #source = "github.com/airyhq/airy.git/infrastructure/terraform/modules/minikube"
+  source = "../../modules/minikube"
```
ljupcovangelski, line 3: Temporary, for testing only.

cli/pkg/cmd/create/create.go
```diff
@@ -88,69 +84,9 @@ func create(cmd *cobra.Command, args []string) {
...
-				fmt.Println()
-			case "API_HOST":
-				fmt.Printf("\t\t %s:\t %s", "API", host)
-				fmt.Println()
-			case "NGROK":
-				fmt.Printf("\t\t %s:\t %s", "NGROK", host)
-				fmt.Println()
-			}
-		}
-
```
chrismatix, line 142: I guess we still want to keep everything below this line, right? (printing docs, storing the provider in the config, tracking)

ljupcovangelski: All the output is printed currently by Terraform. When Terraform finishes, it prints out the outputs which have the API/UI url and the link for the docs. An alternative would be to mute the Terraform output and once Terraform finishes successfully, to get the outputs and print them from the CLI.

infrastructure/terraform/install/airy-core/main.tf
```diff
@@ -17,7 +17,8 @@ provider "kubernetes" {
 }
 
 module "airy_core" {
-  source         = "github.com/airyhq/airy.git/infrastructure/terraform/modules/core"
+  #source         = "github.com/airyhq/airy.git/infrastructure/terraform/modules/core"
+  source         = "../../modules/core"
```
ljupcovangelski, line 21: Temporary for testing purposes

infrastructure/terraform/modules/core/outputs.tf
```diff
@@ -3,6 +3,21 @@ output "core_id" {
 }
 
 output "loadbalancer" {
-    description = "The URL for the load balancer of the cluster. Used to access the UI via the browser"
-    value       = data.kubernetes_service.ingress.status.0.load_balancer.0.ingress.0.hostname
+  description = "The URL for the load balancer of the cluster. Used to access the UI via the browser"
+  value       = data.kubernetes_service.ingress.status.0.load_balancer.0.ingress.0.hostname
+}
+
```
ljupcovangelski, line 9: The api_host output holds also the protocol (http or https)

cli/pkg/cmd/create/create.go
chrismatix: Understood, but what about the tracking and storing the provider in the `cli.yaml`?

cli/pkg/cmd/create/create.go
```diff
@@ -88,69 +84,9 @@ func create(cmd *cobra.Command, args []string) {
...
-		viper.Set("namespace", namespace)
-		viper.WriteConfig()
-
-		airyAnalytics.Track(analytics.Track{
-			UserId: coreConfig["CORE_ID"],
-			Event:  "installation_succesful"})
-		fmt.Printf("📚 For more information about the %s provider visit https://airy.co/docs/core/getting-started/installation/%s", providerName, providerName)
-		fmt.Println()
-	}
 }
```
armanjindal, line 92: What is the desired behavior if we do not have a set provider? When I run `airy create -h` under provider string we say:

`--provider string                  One of the supported providers (aws|minikube). (default "minikube")`

But if I just run `airy create` 

`Error: required flag(s) "provider" not set`

Do we want to default to "minikube"? I think it is better to have people install Airy using cloud providers but not to set a default. I think we can just remove the default note in the CLI.

cli/pkg/providers/minikube/minikube.go
```diff
@@ -45,114 +43,49 @@ func (p *provider) GetOverrides() template.Variables {
...
-
-func (p *provider) Provision(providerConfig map[string]string, dir workspace.ConfigDir) (kube.KubeCtx, error) {
-	if err := p.startCluster(providerConfig); err != nil {
-		return kube.KubeCtx{}, err
+func (p *provider) PreInstallation(workspacePath string) (string, error) {
+	remoteUrl := "github.com/airyhq/airy/infrastructure/terraform/install"
+	installDir := workspacePath + "/terraform"
+	installFlags := strings.Join([]string{"PROVIDER=minikube", "WORKSPACE=" + workspacePath}, "\n")
+
+	var gitGetter = &getter.Client{
```
armanjindal, line 53: Now that we have two providers, and are soon to support more, i.e `aws-ec2` and GCP, the way we have it set up, it will download all files in the `install` directory including the other provider directories. Do we want this? 

Perhaps a fix would be to do two separate calls? One where we append "/core" to `remoteUrl` in line 54, and then do something like `gitGetter.Src = remote URL + "provider_name". I'm sure we could also do them using go routines too.

cli/pkg/providers/minikube/minikube.go
```diff
@@ -45,114 +43,49 @@ func (p *provider) GetOverrides() template.Variables {
 }
 
 func (p *provider) CheckEnvironment() error {
-	return workspace.CheckBinaries([]string{"minikube"})
-}
-func (p *provider) PreInstallation(workspace string) (string, error) {
-	return workspace, nil
+	return workspace.CheckBinaries([]string{"minikube", "terraform"})
```
armanjindal, line 46: I would also add a check for the docker binary as it is needed in the `local exec` resource in minikube's `main.tf`.

armanjindal: As a side note - maybe to include in the docs. It is better if users install docker via the official website rather than brew. Otherwise the you get an error where the `docker deamon is not installed`.

infrastructure/terraform/modules/core/outputs.tf
armanjindal: I wonder if we can also print out the description of the output.. Right now it is just the title on the CLI.

armanjindal (commented):
Looks great! I love how much code is being deleted. I ran into an error when testing re topi provisioning job and topics not getting created in Kafka. Also just dropped some ideas.

cli/pkg/providers/minikube/minikube.go
armanjindal: But I guess this depends on which driver they choose to use. i.e, they can still use VirtualBox.. Then the solution might be to capture the provider-specific variables from the CLI and then check those. This is an issue for later that we discussed.

cli/pkg/cmd/create/create.go
ljupcovangelski: The tracking was moved to the Provision stage. But I guess it makes sense that is also in the  PostInstallation, so that we record an event when the installation finished.

Good idea to store the provider in the `cli.yaml`. I was thinking perhaps in the `install.flags`. Will have a look :+1:.

cli/pkg/cmd/create/create.go
ljupcovangelski: No, I think it is good that we don't have a default and that we output an error if the provider is not set.
So we should remove the default message and also improve the error :smile:  .

cli/pkg/providers/minikube/minikube.go
ljupcovangelski: Good point. But I think this falls under the `minikube installation`. Perhaps we can add a note that a driver is required to run Minikube successfully, but I wouldn't check for any other binaries and drivers.

cli/pkg/providers/minikube/minikube.go
ljupcovangelski: Good point :+1:. I'd open an issue for this for now.

infrastructure/terraform/modules/core/outputs.tf
ljupcovangelski: Yeah that would be good. Terraform doesn't have an option for this unfortunately. So this we can do if we control the output from the CLI.

cli/pkg/providers/minikube/minikube.go
armanjindal: Ok 👍 ! But if they do not have the `docker` installed correctly, for example, the terraform will fail mid-way

bilgeAiry (approved):
not tested after merge will do!

bilgeAiry (approved):
👍🏼

Issue #3558, opened 2022-08-09
Docs for "Contributing Components"


armanjindal:
## Is your feature request related to a problem? Please describe.

As we continue to develop components and work toward supporting external contributions for components, there is discussion as well as docs, which need to be had and created. 

## Describe the solution you'd like

An addition to our Contributing guide docs called "Contributing Components," where we will have an "in development" warning on top of the page. This will allow us to adopt a *docs first* development approach to our internal and external creation of Components. The PR associated with this issue will serve as a place for us to continue our discussion and add to the docs. as we progress.

## Additional context

The need for this has come out of three related points of discussion. 

1. How do we store and update (non-deployment) information relating to components for both "Enterprise" and "Open Source"?
2. How will we update and version the components as the Airy APIs continue to change? 
3. How will we, eventually, support third-party contributions for components which are not part of `Airy Core`?

chrismatix:
This was done

PR #3592, opened 2022-08-15
[#3589] Retrieve installed components from cm

### LINKED ISSUE #3589: components.list returns installed=true for uninstalled component
## Describe the bug

When installing and then uninstalling a component, the `components.list` endpoint, still returns `"installed": true` for that component.

## To Reproduce

- inspect `components.list`  (installed = false)
- install a component (for example the chatplugin `{ "name": "airy-core/sources-chatplugin" }`)
- inspect `components.list` (installed = true)
- uninstall the component
- inspect `components.list` (still installed = true)

## Expected behavior

`installed` should be `false`.

## Screenshots

/

## Environment

All

## Additional context

/

### LINKED ISSUE #3599: components.install installs the latest chart in the repo
## Describe the bug

When installing a new component, the backend installs the latest helm chart in the repo, instead of instaling the version of the currently running `Airy Core` installation.

## To Reproduce

1. Create an Airy Core instance with version other then the latest develop
2. Install a new component
3. See with `helm list` how the versions of Airy Core and the component differ

## Expected behavior

As we agreed for now that the components will track the version of Airy Core, the new component (helm chart) should be installed with the same version.

## Screenshots

/

## Environment

- All

## Additional context

/

ljupcovangelski:
Instead of getting the installed helm charts and using the cache, I did a change where we fetch the installed components by retrieving the ConfigMaps which have the label `core.airy.co/component`.

With this change, I think that also the cache is not needed any more.

Resolves #3589.
Resolves #3599.

### DIFF
--- a/docs/docs/getting-started/components.md
+++ b/docs/docs/getting-started/components.md
@@ -79,6 +79,30 @@ Airy Core contains the following components:
 
 ## Installation
 
-For installation purposes Airy Core is packaged in a Helm chart which creates all the necessary Kubernetes resources. However, every component is an independent entity and can be installed or uninstalled separately. Every component is packaged in its own independent Helm chart and belongs to a `repository`. By default, the `airy-core` repository is added with the components that can be found under `infrastructure/helm-chart/charts/components/charts`.
+For installation purposes Airy Core is packaged in a Helm chart which creates all the necessary Kubernetes resources. However, every component is an independent entity and can be installed or uninstalled separately. Every component is packaged in its own independent Helm chart and belongs to a `repository`. By default, the `airy-core` repository is added with the components that can be found under https://helm.airy.co.
 
-Some components are installed by default, while others can be added or removed later, depending on the particular use case. Components can be managed through the `Control center` or through the [components](/api/endpoints/components) and [repositories](/api/endpoints/repositories) API endpoints.
+You can see all the available components either from the `Catalog` in the `Control center`. From the `Control center` you can also install, uninstall, configure, enable or disable the components.
+
+The following components are part of `Airy Core` and they cannot be uninstalled:
+
+- airy-controller
+- api-admin
+- communication
+- api-websocket
+- frontend-inbox
+- frontend-control-center
+
+Here is a list of the open source components which can be added to `Airy Core`:
+
+- api-contacts
+- integration-source-api
+- integration-webhook
+- media-resolver
+- sources-chatplugin
+- sources-facebook
+- sources-google
+- sources-twilio
+- sources-viber
+- sources-whatsapp
+
+More information about the components API can be found [here](/api/endpoints/components).
--- a/infrastructure/controller/pkg/endpoints/components_install_uninstall.go
+++ b/infrastructure/controller/pkg/endpoints/components_install_uninstall.go
@@ -46,6 +46,13 @@ func (s *ComponentsInstallUninstall) ServeHTTP(w http.ResponseWriter, r *http.Re
 		return
 	}
 
+	chart_version, err := s.getVersion(r.Context())
+	if err != nil {
+		klog.Error(err.Error())
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
 	chartSpec := &helmCli.ChartSpec{
 		ReleaseName: releaseName,
 		ChartName:   chartName,
@@ -53,6 +60,7 @@ func (s *ComponentsInstallUninstall) ServeHTTP(w http.ResponseWriter, r *http.Re
 		UpgradeCRDs: true,
 		Replace:     true,
 		Force:       true,
+		Version:     chart_version,
 		ValuesYaml:  globals,
 	}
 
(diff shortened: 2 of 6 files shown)


infrastructure/controller/pkg/endpoints/components_update.go
```diff
@@ -40,10 +40,8 @@ func (s *ComponentsUpdate) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	for _, component := range requestComponents.Components {
 		if !s.isComponentInstalled(component.Name) {
-			w.Header().Set("Content-Type", "application/json")
-			w.WriteHeader(http.StatusBadRequest)
-			w.Write([]byte(fmt.Sprintf(`{"error": "component %s is not installed"}`, component.Name)))
-			return
+			klog.Error("Trying to apply configuration for a component that is not installed: " + component.Name)
+			continue
```
ljupcovangelski, line 44: If there is a configuration in the `airy.yaml` file for at least one component that is not installed, the whole `airy config apply` command errors with a `Bad Request`. Therefore, I think that it is better if we just skip that component and at the end we return a list of components which have been updated.

infrastructure/controller/pkg/endpoints/components_install_uninstall.go
```diff
@@ -46,13 +46,21 @@ func (s *ComponentsInstallUninstall) ServeHTTP(w http.ResponseWriter, r *http.Re
...
+	}
+
 	chartSpec := &helmCli.ChartSpec{
 		ReleaseName: releaseName,
 		ChartName:   chartName,
 		Namespace:   s.Namespace,
 		UpgradeCRDs: true,
 		Replace:     true,
 		Force:       true,
+		Version:     chart_version,
```
ljupcovangelski, line 63: Install the same version of the component as the version of `Airy Core`.

docs/docs/getting-started/components.md
```diff
@@ -73,6 +73,39 @@ Airy Core contains the following core components:
```
armanjindal, line 73: Regarding the structure of the components page, I suggest we add the updated definition to the top of the page underneath the image. As you can see in PR #3558 I created a re-usable .mdx snippet for the definition. I would suggest adding this before we expose the cards.

docs/docs/getting-started/components.md
```diff
@@ -73,6 +73,39 @@ Airy Core contains the following core components:
...
+- integration-webhook
+- media-resolver
+- sources-chatplugin
+- sources-facebook
+- sources-google
+- sources-twilio
+- sources-viber
+- sources-whatsapp
+
+However, you can also manage the components as helm charts. For example, to install the `sources-chatplugin` component, run this in your workspace:
```
armanjindal, line 101: Do we not want users to download components via the Control Center? Or is this still under development? Do we prefer that they use the endpoints over the UI?

docs/docs/getting-started/components.md
ljupcovangelski: Sounds good :+1:. I will re-use the snippet from that PR :pray:

docs/docs/getting-started/components.md
ljupcovangelski: The `Control Center` is fully functional. I just thought of sharing additional info here, so that it is more clear how we install/uninstall them and as a failback if the UI fails.

ljupcovangelski: Ah, I now see what you mean. The `Control center` should come as a first suggestion :). I will add this :+1:.

infrastructure/controller/pkg/endpoints/components_install_uninstall.go
```diff
@@ -46,13 +46,21 @@ func (s *ComponentsInstallUninstall) ServeHTTP(w http.ResponseWriter, r *http.Re
 		return
 	}
 
+	chart_version, err := s.getVersion(r.Context())
```
chrismatix, line 49: nitpick: This should be in camelCase

infrastructure/controller/pkg/endpoints/components_update.go
chrismatix: I agree. Small suggestion: Could you add a single sentence about this to the HTTP API docs?

infrastructure/controller/pkg/endpoints/components_update.go
```diff
@@ -71,7 +69,10 @@ func (s *ComponentsUpdate) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 //NOTE: Prevent the upload of a configmap if the component is not present
 func (s *ComponentsUpdate) isComponentInstalled(configName string) bool {
-	deployedCharts := s.DeployedCharts.GetDeployedCharts()
-
+	deployedCharts, err := k8s.GetInstalledComponents(context.TODO(), s.namespace, s.clientSet)
+	if err != nil {
+		klog.Error("Unable to get installed components:\n" + err.Error())
```
chrismatix, line 74: In the case of an error here the `/components.update` call should fail though, right? It means that there is something not right with the cluster. 

In any case this function should return the error so that we can differentiate between both cases 🤔

ljupcovangelski: Ok, in that case I will make the function return `bool,err` :+1:

chrismatix (approved):

PR #3610, opened 2022-08-17
[#3605] Fix connectors connected channels

### LINKED ISSUE #3605: Fix connectors connected page if there is no channel connected
Right now, if a connector is installed & do not have any channel connected, the page displays a "Results not found"content, which is not accurate. 

It would be better to display content that reflects that the connector is installed but does not have any connected channels.

Screenshot 2022-08-17 at 10 53 24

AudreyKj:
closes #3605


thorstenairy (approved):
❤️

thorstenairy (approved):

PR #3611, opened 2022-08-18
[#3403] Rasa Connector 

### LINKED ISSUE #3403: Rasa Connector
[In progress]

armanjindal:
To Do:
- [x] Forward messages from Airy to Rasa 
- [x] Reply with a text response
- [x] Support multiple replies
- [x] Renders images as URLs
- [x] Packed as a Helm Chart

Resolves #3403


backend/connectors/rasa-connector/src/main/java/co/airy/core/rasa_connector/RasaConnectorService.java
```diff
@@ -0,0 +1,67 @@
...
+    public void send(Message userMessage) {
+        try {
+            List<MessageSendResponse> rasaResponseList = this.rasaClient.sendMessage(MessageSend.builder()
+                    .message(getTextFromContent(userMessage.getContent()))
+                    .sender(userMessage.getId())
+                    .build());
+            // Unpack multiple response(s)
+            for (MessageSendResponse rasaResponse: rasaResponseList) {
+                try {
+                    messageHandler.writeReplyToKafka(userMessage, rasaResponse);
```
chrismatix, line 41: @armanjindal this is good, but doing it [this wasy](https://github.com/airyhq/airy/pull/3614/files#diff-3f2f43c5b7b114c93a58727cc16cddb743f7e498250bc5aef09dc4b616fc7aabR117) is better because it gives us better guarantees: Processing will be retried properly when writing to Kafka fails.

backend/connectors/rasa-connector/BUILD
```diff
@@ -0,0 +1,51 @@
...
+    "//:lombok",
+    "//backend/model/message",
+    "//:feign",
+    "//lib/java/log",
+    "//lib/java/spring/kafka/core:spring-kafka-core",
+    "//lib/java/spring/core:spring-core",
+    "//lib/java/spring/kafka/streams:spring-kafka-streams",
+    "//lib/java/spring/async:spring-async",
+    "//backend/model/metadata",
+    "@maven//:commons_codec_commons_codec",
```
chrismatix, line 22: How is this dependency used?

backend/connectors/rasa-connector/src/main/java/co/airy/core/rasa_connector/RasaConnectorService.java
armanjindal: Thanks for the tip @chrismatix! The link points me to line 117 of the `EventsRouter.java` file and the `.to` method there. Is this what you meant to link to? Are you perhaps referencing `MessageMapper.java` file? Since it seems like you do add the responses to a list instead of sending one at a time. I'll keep digging

chrismatix: @armanjindal Yes, so how this works is that this `send` method returns a `List<KeyValue<String, Record>>` where the pairs are what you would send to Kafka manually. In the `to` method you then send records to topics depending on type.

backend/connectors/rasa-connector/src/main/java/co/airy/core/rasa_connector/Stores.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    @Override
+    public void onApplicationEvent(ApplicationStartedEvent event){
+        final StreamsBuilder builder = new StreamsBuilder();
+        builder.<String, Message>stream(
+                new ApplicationCommunicationMessages().name(),
+                Consumed.with(Topology.AutoOffsetReset.LATEST)
+        ).filter((messageId, message) -> message != null && isNewMessage(message) && message.getIsFromContact())
+                .peek((messageId, message) -> rasaConnectorService.send(message));
```
juan-sebastian, line 38: You can add a comment here to say that `rasaConnectorService.send` is an async call

backend/connectors/rasa-connector/src/main/java/co/airy/core/rasa_connector/RasaClientConfig.java
```diff
@@ -0,0 +1,24 @@
...
+public class RasaClientConfig {
+    @Bean
+    public RasaClient rasaClient(@Value("${rasa.rest-webhook-url}") String rasaRestUrl) {
+        return Feign.builder()
+                .client(new OkHttpClient())
+                .encoder(new JacksonEncoder())
+                .decoder(new JacksonDecoder())
+                .logger(new feign.Logger.ErrorLogger())
+                .logLevel(feign.Logger.Level.FULL)
+                .target(RasaClient.class, rasaRestUrl);
```
juan-sebastian, line 22: There is no authentication token? How do you authenticate witht he rasa api?

juan-sebastian (approved):
Looks good to me, I just left 2 small comments

backend/connectors/rasa-connector/src/main/java/co/airy/core/rasa_connector/RasaClientConfig.java
armanjindal: So to just call the REST API endpoint, no authentication is needed. You need authentication when calling other endpoints like /models.parse etc. (i.e for a human-handoff like feature) but not for getting plain responses (without any other data).

chrismatix (approved):

juan-sebastian (approved):

PR #3637, opened 2022-08-23
[#3634] Support whatsApp source in the render library

### LINKED ISSUE #3634: Support whatsApp source in the render library
## Describe the bug

Inbox is broken on the latest `develop ` (v49 alpha) : the error comes from the render library. 

the  getOutboundMapper doesn't recognize 'whatsapp' as a source and triggers an error. 

To do: 
- we should add support for 'whatsapp' as a source
- prevent the render library of breaking if it doesn't recognize the source

## To Reproduce

Navigate to the Inbox UI and look at the JS console for errors

## Screenshots

Screenshot 2022-08-23 at 11 23 39

## Environment

- OS: iOS
- Browser: Chrome

AudreyKj:
closes #3634

### DIFF
--- a/frontend/inbox/src/components/IconChannel/index.tsx
+++ b/frontend/inbox/src/components/IconChannel/index.tsx
@@ -73,7 +73,7 @@ const SOURCE_INFO = {
     avatar: () => <ViberIcon />,
   },
   whatsapp: {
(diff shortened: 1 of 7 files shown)


thorstenairy (approved):
🚀

Issue #3671, opened 2022-09-02
Restructure backend directories


chrismatix:
## Is your feature request related to a problem? Please describe.

The current directory structure is a bit arbitrary and it feels pointless to keep moving directories around whenever we change our understanding of component categories.

## Describe the solution you'd like

We, therefore, want to implement this restructuring:

Current:
```
.
└── backend/
    ├── api/
    │   ├── admin
    │   ├── communication
    │   └── ...
    ├── avro
    ├── model
    ├── connectors/
    │   └── rasa
    └── sources/
        ├── api
        ├── facebook
        └── ...
```
New
```
.
└── backend/
    ├── components/
    │   ├── {whatever the component name is}
    │   ├── api-admin
    │   ├── api-communication
    │   ├── rasa
    │   ├── media-resolver
    │   ├── sources-facebook
    │   └── ...
    ├── avro
    └── model
```

ljupcovangelski:
Great initiative and it is good to go now like this. If we have in the future components that have both frontend and backend part (now it is just the sources-chatplugin I think), we can have a structure similar to this:

```
.
└── components/
    ├── api-admin/
    │   ├── backend/
    ├── chatplugin-connector/
    │   ├── backend/
    │   ├── frontend/
    ├── dialogflow-connector/
    │   ├── backend/
    ├── control-center
    │   ├── frontend/
    ├── avro
    └── model
```

But for now, I agree with the proposed directory structure :+1: .

PR #3675, opened 2022-09-02
[#3671] Update links in contributing docs

ljupcovangelski:


docs/docs/guides/contributing-components.md
```diff
@@ -3,12 +3,6 @@ title: Contributing Components
 sidebar_label: Contributing Components
 ---
 
-:::warning
```
ljupcovangelski, line 6: Why was this removed? The repo is still empty. I think we should put our own components first, so that they serve as an example.

ljupcovangelski (commented):
I think we should keep the comment that this is still WIP.

docs/docs/guides/contributing-components.md
chrismatix: It read a bit confusing and imprecise: Of course, you can already contribute components. It's just that the catalog mechanism is not complete yet. I updated it to reflect that.

ljupcovangelski: Ok, feel free to approve, as I cannot do it :smile:

chrismatix (approved):

PR #3704, opened 2022-09-07
[#3699] Allow customization of repositories

### LINKED ISSUE #3699: Repositories cannot be changed through the Helm chart
We need to add a possibility for customizing/adding repositories through the Helm chart, so that not only the `Airy Core` components can be installed through the Catalog.

ljupcovangelski:
The output on `helm template` renders the ConfigMap like this:

```
# Source: airy/templates/controller/configmap.yaml
kind: ConfigMap
apiVersion: v1
metadata:
  name: repositories
  namespace: 
data:
  repositories.json: |
    {
      "repositories":
            [
        {
          "name": "airy-core",
          "password": null,
          "url": "https://helm.airy.co",
          "username": null
        }
      ]
    }
```

Tested also with a list of multiple repositories.

Resolves #3699 .


juan-sebastian (approved):

bilgeAiry (approved):
Looks good!

Issue #3536, opened 2022-08-03
Whatsapp cloud source: Mark messages as read


chrismatix:
[Parent issue](https://github.com/airyhq/airy/issues/3464)

We should be able to mark messages as read in the source (https://developers.facebook.com/docs/whatsapp/cloud-api/guides/mark-message-as-read)

chrismatix:
completed in #3765

PR #3765, opened 2022-09-23
[#3536] Whatsapp mark messages read

chrismatix:
prerequisite for #3536


backend/components/unread-counter/src/main/java/co/airy/core/unread_counter/Stores.java
```diff
@@ -0,0 +1,112 @@
+package co.airy.core.unread_counter;
```
chrismatix, line 1: This PR:
1. Moves counting messages not read by the business into its own app 
2. Extends the topology to produce more granular "message was read" metadata that can be consumed by components

juan-sebastian (approved):
Looks good to me

bilgeAiry (approved):
👍🏼

PR #3766, opened 2022-09-23
[#3589] Remove the cache from the controller

ljupcovangelski:


chrismatix (approved):
So this is a follow up fix for #3589? :)

ljupcovangelski:
> So this is a follow up fix for #3589? :)

Yes :). The solution of that issue resulted in not using the cache at all, but the code remained.

PR #3799, opened 2022-10-05
[#3520] Fix flaky webhook consumer test

### LINKED ISSUE #3520: webhook.consumer.ConsumerTest is flaky
## Describe the bug

The `co.airy.core.webhook.consumer.ConsumerTest` fails inconsistently both on CI and on development machines. This causes a delay in the process.

## To Reproduce

Run the test a couple of times using the no cache flag.

## Expected behavior

The test should execute successfully every time.

chrismatix:
resolves #3520

### DIFF
--- a/backend/components/webhook/consumer/src/main/java/co/airy/core/webhook/consumer/Consumer.java
+++ b/backend/components/webhook/consumer/src/main/java/co/airy/core/webhook/consumer/Consumer.java
@@ -34,7 +34,7 @@ public class Consumer implements DisposableBean {
 
     @Scheduled(fixedDelay = 50)
     public void process() {
-        final Job job = consumer.reserveJob(10);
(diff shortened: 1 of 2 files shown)


backend/components/webhook/consumer/src/main/java/co/airy/core/webhook/consumer/Consumer.java
```diff
@@ -34,7 +34,7 @@ public class Consumer implements DisposableBean {
 
     @Scheduled(fixedDelay = 50)
     public void process() {
-        final Job job = consumer.reserveJob(10);
+        final Job job = consumer.reserveJob(30);
```
chrismatix, line 37: This is not the actual fix

ljupcovangelski (approved):
Great :raised_hands:

Issue #3764, opened 2022-09-23
Kafka helm chart with SASL and authentication


ljupcovangelski:
We need to improve the Helm chart for Kafka, to include options for encryption and authentication on the serer side, so that it can be deployed in another Kubernetes cluster on the Internet then the `Airy Core` components.

chrismatix:
@ljupcovangelski The Kafka chart is bundled with the airy chart, right? In what scenario would the Kafka chart be installed in a different cluster?

ljupcovangelski:
@chrismatix I agree it is outside of the scope of deploying Airy (which includes the bundles Kafka chart). However, as we do have the chart independent and it is possible to install Kafka separately in another place (namespace) and connect Airy to it, I think it is good that we include this functionality in the chart.

chrismatix:
@ljupcovangelski yes makes sense. Thanks for clarifying

PR #3882, opened 2022-10-26
[#3764] Kafka 3.3.1 docker image and helm chart

ljupcovangelski:
- Decouple the Kafka installation from the main helm chart
- Build and push a Kafka 3.3.1 docker image
- Deploy a Kafka 3.3.1 helm chart


ljupcovangelski:
I'm not sure if this structure is best (that is why the PR is draft). We will probably want to have multiple different Kafka versions in the future (to be deployed in different Airy instances), but whether Kafka should be directly under infrastructure? Perhaps there should be something like prerequisites, but then again Kafka is the main infrastructure component...

I would propose that we have something like this in the future:
```
infrastructure
|--helm         (main chart for deploying Airy)
|--kafka        (images and charts for the Kafka installation)
|--terraform    (terraform modules for installation)
|--tools        (other tools)
```
(the controller and the lib directories will soon be removed anyway I think).

bilge-cakmak:
I think this structure makes a lot sense.. Terraform providers will have their own directory. Maybe. there we will have another directory for confluent. I am not sure about the effects of having these out of terraform workspace as now currently we are using shared files for installation. Anyhow I cannot foresee any obstacle having them outside of terraform directory. For me it is good to go and still open for further discussion.

bilge-cakmak (approved):
👍

PR #3891, opened 2022-10-28
[#3885] Whatsapp connector !config error

### LINKED ISSUE #3885: Whatsapp connector !config error
## Describe the bug

When installing Whatsapp Business Cloud component, on Connectors view, the connector appears to be "Not Configured' but has 1 connected channel. 

## To Reproduce

1. Install Whatsapp Business Cloud component from Catalog view. 
2. Navigate to Connectors view and find Whatsapp Business Cloud Connector.
3. See error on connector card that the status is not configured but the connector has one connected channel. 

## Expected behavior

When a user installs a new source component. it can't have any connected channels as the user needs to apply configuration first before being able to add a channel. 

## Screenshots

FIG1. WHATSAPP BUSINESS CLOUD CONNECTOR CARD ON CONNECTOR VIEW

## Environment

- Browser: [Chrome]
- Version: [0.51.0]

AudreyKj:
closes #3885 

We agreed on this flow:
---> if a component is `not configured`, we do not show the connected channels (if the component has any)
--->  if a component is  `disabled` and `unhealthy` status: do we show the connected channels (because it means that the connector is configured and you can add channels)

### DIFF
--- a/frontend/control-center/src/pages/Connectors/ChannelCard/index.module.scss
+++ b/frontend/control-center/src/pages/Connectors/ChannelCard/index.module.scss
@@ -52,7 +52,7 @@
   align-items: center;
   margin-bottom: 14px;
 
-  span {
+  p {
     color: var(--color-airy-blue);
     margin-right: 4px;
     text-decoration: underline;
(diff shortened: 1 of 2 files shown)


thorstenairy (approved):
🚀

PR #3914, opened 2022-11-04
[3901 & 3900] Feature/3901 create kafka topic for components events

### LINKED ISSUE #3900: Concurrent installing new connectors leads to problems
## Describe the bug

When a user tries to install several connectors at the same time, the backend is unable to reset the cache properly which leads to further state issues. 

## To Reproduce

Steps to reproduce the behaviour:

1. Go to Control-Center
2. Go to Catalog
3. Try to connect multiple connectors concurrent

## Expected behaviour

The requested installs getting queued correctly and installed according the queue

### LINKED ISSUE #3901: Create Kafka topic for components events
We need to create a Kafka topic that streams, components events, like `installationStatus`

juan-sebastian:
closes #3901 
closes #3900

### DIFF
--- a/backend/components/installer/src/main/java/co/airy/core/api/components/installer/Stores.java
+++ b/backend/components/installer/src/main/java/co/airy/core/api/components/installer/Stores.java
@@ -1,0 +1,71 @@
+package co.airy.core.api.components.installer;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import org.slf4j.Logger;
+import org.apache.avro.specific.SpecificRecordBase;
+import org.apache.kafka.clients.producer.KafkaProducer;
+import org.apache.kafka.clients.producer.ProducerRecord;
+import org.springframework.boot.actuate.health.Health;
+import org.springframework.boot.actuate.health.HealthIndicator;
+import org.springframework.stereotype.Component;
+
+import co.airy.log.AiryLoggerFactory;
+import co.airy.avro.communication.Metadata;
+import co.airy.kafka.schema.ops.OpsApplicationComponents;
+import lombok.Synchronized;
+
+import static co.airy.model.metadata.MetadataRepository.getId;
+
+@Component
+public class Stores implements HealthIndicator {
+
+    private static final Logger log = AiryLoggerFactory.getLogger(Stores.class);
+
+    private final KafkaProducer<String, SpecificRecordBase> producer;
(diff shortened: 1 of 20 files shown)


backend/components/installer/src/main/java/co/airy/core/api/components/installer/Stores.java
```diff
@@ -0,0 +1,72 @@
...
+        storeComponent(cacheChanges);
+
+        prevoiusCacheMap = cacheStore
+            .entrySet()
+            .stream()
+            .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+    }
+
+    @Override
+    public Health health() {
```
chrismatix, line 69: nitpick: Remove as this health indicator does not do anything

backend/components/installer/src/test/java/co/airy/core/api/components/installer/CatalogHandlerTest.java
```diff
@@ -89,88 +82,57 @@ static void afterAll() throws Exception {
...
-                .filter((c -> c.getName().equals("cognigy-connector")))
-                .findAny()
-                .orElse(null);
-            assertThat(cognigyConnector, is(notNullValue()));
-            assertThat(cognigyConnector.getInstallationStatus(), equalTo(InstallationStatus.uninstalled));
-        }
-
+        doReturn(getInstallationStatues()).when(installationStatusComponentsHandler).getInstallationStatusComponentsCache();
+
+        final List<ComponentDetails> listComponents = catalogHandler.listComponents();
```
chrismatix, line 91: It would be great if we could mock this. Otherwise, this test might fail again if the data in the catalog changes. What do you think?

chrismatix (commented):
Looks good! Minor comments

backend/components/installer/src/main/java/co/airy/core/api/components/installer/Stores.java
juan-sebastian: But no class implements this interface. The health URL will still be instantiated by spring?

backend/components/installer/src/test/java/co/airy/core/api/components/installer/CatalogHandlerTest.java
juan-sebastian: Well this is an integration test. and I think the test should fail. If for example, someone does something wrong on the `catalog` page. Otherwise we have no way to know something went wrong until is in production. I will suggest to mock it when we have CI/Test for the catalog repo.

PR #3924, opened 2022-11-08
[#3923] Use auth-provider in GCP kubeconfig file

### LINKED ISSUE #3923: The kubeconfig file for the GCP provider expires very fast
The kubeconfig file generated by the `infrastructure/terraform/modules/gcp-gke` module expires very fast.
The file should not store a static `token`, but a `auth-provider` method, so that the same file can stay and the token gets refreshed over time.

Probably this will mean replacing the `gke_auth` module, as the used module doesn't support this option.

ljupcovangelski:
Resolves #3923


bilge-cakmak (approved):
Tested 👍

Issue #3919, opened 2022-11-07
Add Component Websocket event updates in the backend 


AudreyKj:
Create a new stream for `Components.update ` that would gives updates on Components' events

PR #3930, opened 2022-11-10
[#3919] component Websocket update backend

AudreyKj:


juan-sebastian (approved):
Great work, Maybe we can add this to the release `0.52.0`

PR #3940, opened 2022-11-15
[#3939] Fix showing airy products in connectors

### LINKED ISSUE #3939: Airy product are displayed in connectors
We need to hide airy products like webhooks, contacts, mobile in Connectors as we do not need to configure them.

thorstenairy:
closes #3939


ljupcovangelski (approved):
:+1:

Issue #3943, opened 2022-11-15
Fix of the gcp-gke terraform output file


bilge-cakmak:
Fix the referenced output file for `infrastructure/terraform/install/gcp-gke/output.tf`

PR #3944, opened 2022-11-15
[#3943] Fix of the gcp-gke terraform modules output file

bilge-cakmak:
Fix of the gcp-gke terraform modules output file


ljupcovangelski (approved):
Thanks :+1:

Issue #1614, opened 2021-04-22
The apps should wait for the kaka topics to be created


ljupcovangelski:
Currently the apps are only waiting for the minimum number of kafkas to be available, before they start.

We should change that so that they wait for the topics to be provisioned:
- The job for provisioning the topics should wait for the minimum kafkas to be available
- The apps should wait for the job for provisioning the topics to finish

PR #3966, opened 2022-11-21
[#1614] The apps should wait for the kafka topics to be created

bilge-cakmak:
I tested with a gke cluster. Somehow helm install on a minikube cluster had a wrong ingress configurstion.. the settings for localhost weren't there.

#### observed behaviour 
airy-controller :`4th Running`
api-admin: `11-12th Running`
api-communication : `10th Running`
api-components-installer : `7th running` still on going work here..  
api-websocket : `11-12th Running`
beanstalkd-0 : `3rd Running` 
frontend-control-center : `2nd Running` --> why this pod is up while kafka and zookeeper initialise 
frontend-inbox : `1st Running` -> this pod/app is up earlier than everything while kafka-0 pending
kafka-0 : `8th Running`
*provisioning-topics* : the init container for this there init `9th Running`
schema-registry: `14th Running` the last running
unread-counter:`13th Running`
wait-for-api-communication : `5th Running`
zookeeper-0:`6th Running`

There [wait-for-api-comminnication](https://github.com/airyhq/airy/blob/develop/infrastructure/helm-chart/templates/provisioning/job-wait.yaml) job has 
  ```
annotations:
    "helm.sh/hook": "post-install,post-upgrade"
    "helm.sh/hook-weight": "5"
    "helm.sh/hook-delete-policy": hook-succeeded
```
Although the job has the `post` prefixed helm hook value it is up earlier than the stateful sets


ljupcovangelski (changes requested):
I did a test, but unfortunately the helm lifecycle management and `helm.sh/hook-weight` doesn't work like I thought it will. So still the provisioning job and the deployments are installed at the same time.

I think we have two options:
1. Add a wait container that waits for a job with a certain label to be `Completed`. We need to add the label to the job, but also check that the job is deleted before all the deployments are created (if not, then an old version of the job will be there).
2. Put Kafka, the other prerequisites and the topic creation job into the `pre-install/pre-upgrade` stage.

We can also create the topics from every app individually. This way, when a component is not installed - the topics are also not there. Would not solve the underlying problem however and perhaps we would still need option-1.

ljupcovangelski:
Postponed for a future release.

Issue #3974, opened 2022-11-29
Update docs on Airy Core


ljupcovangelski:
Update the documentation about Airy Core.

PR #3981, opened 2022-12-05
[#3974] Fix pushing local images to Minikube

### LINKED ISSUE #3979: Publishing and running images locally doesn't work
The publishing and running of docker images locally doesn't work because Minikube is defaulting to the `containerd` instead of `docker`.

This is the error

```
 minikube -p airy-core docker-env

❌  Exiting due to MK_USAGE: The docker-env command is only compatible with the "docker" runtime, but this cluster was configured to use the "containerd" runtime.
```

ljupcovangelski:
Resolves #3979.


thorstenairy (approved):
🚀

